PR #6, opened 2021-01-28
Add prefix to the generated lib files

antonio2368:


gitbuda (approved):
I think that's the only possible way to make the distinction between modules for now.

PR #7, opened 2021-01-28
Initialized modules for set cover, distance and tsp

Josipmrden:


python/graph_analyzer.py
```diff
@@ -0,0 +1,280 @@
+import sys
```
gitbuda, line 1: I've already included this module on the top-level main.

python/nxalg.py
```diff
@@ -0,0 +1,825 @@
+import sys
```
gitbuda, line 1: This is also already merged to the main. Please delete.

python/wcc.py
```diff
@@ -0,0 +1,51 @@
+import sys
```
gitbuda, line 1: Already included, please delete.

python/mgp_networkx.py
```diff
@@ -0,0 +1,289 @@
+import sys
+import mgp
```
gitbuda, line 2: Also included, please delete.

gitbuda (changes requested):
- [ ] Remove the `pyc` file.
- [ ] Move everything to the top level of the `python` folder.

Utils are nice, but I think it's easier to have everything required in the actual module because it will be less error-prone. I think that's better initially because everything will work out of the box. Once we reuse the same function 3+ times, it makes sense to make the actual until.

PR #8, opened 2021-01-28
Added tsp, setcover and distance

Josipmrden:
For now, everything is in its own file and distance is duplicated in the tsp query module. When structure is complete, it will be separated into different files

### DIFF
--- a/python/distance_calculator.py
+++ b/python/distance_calculator.py
@@ -1,0 +1,112 @@
+import mgp
+import math
+
+KM_MULTIPLIER = 0.001
+LATITUDE = 'lat'
+LONGITUDE = 'lng'
+
+
+@mgp.read_proc
+def single(context: mgp.ProcCtx,
+              start: mgp.Nullable[mgp.Any],
+              end: mgp.Nullable[mgp.Any],
+              metrics: str = 'm'
+              ) -> mgp.Record(distance=mgp.Number):
+    '''
+    This distance calculator procedure  for one entry returns 1 field.
+
+      * `distance` is the final result obtained by calculating distance (in metres) between the 2 points who
+        each have its latitude and longitude.
+
+    The procedure can be invoked in openCypher using the following calls:
+      CALL distance_calculator.single((:Point {lat:1, lng:2}), (:Point {lat:2, lng:3.5})) YIELD distance;
+      MATCH (n1:Point), (n2:Point) CALL distance_calculator.procedure(n, 1) YIELD * RETURN *;
+    '''
+
+    if not isinstance(start, (mgp.Vertex)) or not isinstance(end, (mgp.Vertex)):
+        return mgp.Record(distance=None)
+
+    distance = calculate_distance_between_points(
+        dict(start.properties.items()),
+        dict(end.properties.items()),
+        metrics)
+
+    return mgp.Record(distance=distance)
+
+
+@mgp.read_proc
+def multiple(context: mgp.ProcCtx,
+              start_points: mgp.Nullable[mgp.Any],
+              end_points: mgp.Nullable[mgp.Any],
+              metrics: str = 'm'
+              ) -> mgp.Record(distances=list):
+    '''
+    This distance calculator procedure  for multiple entries returns 1 field.
+
+      * `distances` is the final result obtained by calculating distances (in metres) between pairs of
+      points who each have its latitude and longitude
+
+    The procedure can be invoked in openCypher using the following calls:
+      CALL distance_calculator.multiple([(:Point {lat:1, lng:2})]), [(:Point {lat:3,lng:4.5})])) YIELD distances;
+    '''
+
+    if len(start_points) != len(end_points) or len(start_points) == 0:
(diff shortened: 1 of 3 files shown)


python/set_cover.py
```diff
@@ -0,0 +1,275 @@
...
+        '''
+
+        element_values = [x.id for x in element_vertexes]
+        set_values = [x.id for x in set_vertexes]
+        all_elements = set(element_values)
+        all_sets = set(set_values)
+
+        elements_by_sets = dict()
+
+        for idx in range(len(element_values)):
```
antonio2368, line 135: `for element, contained_set in zip(element_values, set_values)`?

python/tsp.py
```diff
@@ -0,0 +1,259 @@
...
+
+def solve_1_5_approx(dm):
+    '''
+    Solves the tsp_module problem with 1.5-approximation (Christofides algorithm).
+    :param distance_matrix: Distance matrix.
+    :return: List of indices - path between them (based on distance matrix indexes)
+    '''
+
+    mst = get_mst(dm)
+    odd_matchings = [x[0] for x in list(filter(lambda x: x[1] % 2 == 1, list(mst.degree)))]
```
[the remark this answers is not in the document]
tonilastre: In general, a good rule for converting iterable to `list` is only if you need `list` functionalities: you will iterate multiple times, you need `len` or you will need to pin point value by index. In any other case, you can leave generators (`[1, 2, 3, ...] vs (1, 2, 3, ...)`

tonilastre (commented):
Overall it looks very good! There are some comments about speeding up some things, memory optimizations, and optional syntax sugars.

python/distance_calculator.py
```diff
@@ -0,0 +1,112 @@
...
+
+KM_MULTIPLIER = 0.001
+LATITUDE = 'lat'
+LONGITUDE = 'lng'
+
+
+@mgp.read_proc
+def single(context: mgp.ProcCtx,
+              start: mgp.Nullable[mgp.Any],
+              end: mgp.Nullable[mgp.Any],
```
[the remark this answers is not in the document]
Josipmrden: I was actually looking at example.py query module, I guess it wasn't that good of a reference, will fix.

python/distance_calculator.py
```diff
@@ -0,0 +1,112 @@
+import mgp
+import math
+import itertools
+
+KM_MULTIPLIER = 0.001
+LATITUDE = 'lat'
+LONGITUDE = 'lng'
+valid_metrics = ['m', 'km']
```
tonilastre, line 8: Just like you did correctly for `KM_MULTI...`, `LATITUDE` and `LONGITUDE` it would be great to have this as `VALID_METRICS` because it is a constant.

tonilastre: It should be fixed in both files using the constant.

python/distance_calculator.py
```diff
@@ -0,0 +1,112 @@
...
+    :return: float distance
+    '''
+
+    if (LATITUDE not in start.keys()
+        or LONGITUDE not in start.keys()
+        or LATITUDE not in end.keys()
+        or LONGITUDE not in end.keys()):
+        return None
+
+    if metrics.lower() not in valid_metrics:
```
tonilastre, line 84:
```suggestion
    if not isinstance(metrics, str) or metrics.lower() not in valid_metrics:
```
Because user can send `metrics=None` and `.lower()` will fail.

python/distance_calculator.py
```diff
@@ -0,0 +1,112 @@
...
+        dict(end.properties.items()),
+        metrics)
+
+    return mgp.Record(distance=distance)
+
+
+@mgp.read_proc
+def multiple(context: mgp.ProcCtx,
+              start_points: list,
+              end_points: list,
```
tonilastre, line 38: For all complex types, it is better to use typing objects from `typing` module. E.g (maybe this could be useful for Field Notion):

```python
from typing import List, Dict, Union, Optional, Set, Tuple, Iterable
from myobject import MyObject

list_of_str: List[str] = ['a', 'b', 'c']
set_of_ints: Set[int] = {1, 2, 3}
tuple_of_str_list_str: Tuple[str, List[str]] = ('Mrma', ['a', 'b'])
iterable_of_strs: Iterable[str] = ('a', 'b', 'c')

dict_of_str_to_set_int: Dict[str, Set[str]] = { 'a': {1, 2, 3}, 'b': {4, 5, 6} }
dict_of_str_to_my_object: Dict[str, MyObject] = { 'a': MyObject(), 'b': MyObject() }

list_of_strs_and_none: List[Optional[str]] = ['a', None, 'b', 'c'] # Same as List[Union[str, None]]
```

tonilastre: So in this case, you could have `List[mgp.Nullable[mgp.Vertex]]` which is more correct typing.

Josipmrden: it seems that List[mgp.Nullable[mgp.Vertex]] does not longer work so I specified List[mgp.Vertex] and it is working like before

PR #9, opened 2021-01-29
Extract mg utils

antonio2368:


cpp/pagerank_module/algorithm/pagerank_algorithm.hpp
```diff
@@ -0,0 +1,166 @@
+/// @file
+
+#pragma once
```
antonio2368, line 3: #ifndef guards and #pragma once are both used we need to decide on one.
I'd go with the pragma

gitbuda: I agree.

gitbuda (approved):

PR #12, opened 2021-02-25
Add Rust query modules infrastructure

gitbuda:
The goal of this PR is to write a Rust wrapper for Memgraph query modules.

* [rsmgp-sys](https://crates.io/crates/rsmgp-sys)
* [YouTube](https://www.youtube.com/playlist?list=PL7Eotag2rRhYdr7Ykd759mfuWn5oxUYiL)

## Work Items

- [x] Decide how the library code will be tested and implement initial tests.
    * Probably the best mock framework is [mockall](https://docs.rs/mockall).
    * The final approach is borrowed from [here](https://github.com/asomers/mockall/issues/240).
- [x] Decide about workspace or implementation of the custom build process.
    * Workspaces and custom build both don't seem feasible ([more info](https://internals.rust-lang.org/t/multiple-libraries-in-a-cargo-project/8259)).
    * The decision is to go with a separated `rsmgp-sys` crate.
    * All query modules will be standalone projects (similar to the CPP structure).
- [x] Add initial CI workflow.
- [x] Implement and test all primitive value functions (null, bool, int, string, double).
- [x] Implement `struct rsmgp_sys::memgraph::Memgraph` containing `args`, `graph`, `result`, `memory` and `module`.
- [x] Implement all properties related.
- [x] Implement all vertex related.
- [x] Implement all edge related.
- [x] Implement wrapper and test set_error_msg.
- [x] Polish regular error handling via `MgpResult`.
- [x] Implement all path related.
- [x] Implement all list related.
- [x] Implement all map related.
- [x] Implement all graph related.
- [x] Setup code coverage on CI.
- [x] Test all properties related.
- [x] Test all vertex related.
- [x] Test all edge related.
- [x] Test all path related.
- [x] Test all list related.
- [x] Test all map related.
- [x] Test all graph related.
- [x] Write macro for `XYZ_procedure_c` (try to minimize boilerplate code).
- [x] Implement and test utils for defining returning structure and inserting results (types enum, macros).
- [x] Add catch_unwind everywhere.
- [x] Add full support for procedure arguments.
- [x] Add support for supported transactional stuff.
- [x] Polish error messages.
- [x] Write documentation.
- [x] Publish `rsmgp-sys` crate on crates.io.
- [x] Link all the live stream videos here!
- [x] Implement top-level Python build capabilities.
- [x] Write a blog post about how to use `rsmgp` in complex environment, use `rsmgclient`.


e2e/rsmgp_example_test/test_simple/input.cyp
```diff
@@ -0,0 +1,2 @@
+MATCH (n) DETACH DELETE n;
```
gitbuda, line 1: Reload the modure prior to execution.

gitbuda:
@antaljanosbenjamin & @jmatak please leve high-level comments first because the code is not polished yet

e2e/rsmgp_example_test/test_simple/test.yml
```diff
@@ -0,0 +1,27 @@
+query: >
+    CALL libmemgraph_query_module_in_rust_example.test_procedure() YIELD *;
```
gitbuda, line 2: @jmatak is it possible to execute more than one query here? It would be good to reload the module before executing the test query.

e2e/rsmgp_example_test/test_simple/test.yml
```diff
@@ -0,0 +1,27 @@
...
+
+output:
+    - first_label: ""
+      has_L3_label: false
+      labels_count: 0
+      name_property: "unknown"
+      properties_string: "age: 29"
```
gitbuda, line 9: @jmatak is it possible to test the whole vertex here?

e2e/rsmgp_example_test/test_simple/test.yml
jmatak: Can you elaborate more on what you mean by executing more than one query, why should you do that and not make another test case?

e2e/rsmgp_example_test/test_simple/test.yml
jmatak: Currently, it is not supported, however, if there is a need for that it can be easily extended with the new pymgclient and gqlalchemy modules.

e2e/rsmgp_example_test/test_simple/test.yml
gitbuda: Prior to the execution of a query, I would like to reload the tested module (or all of them). That's not useful inside CI, but during local development, I would like to rerun an e2e test but before running the test query, the module has to be reloaded because it had to be recompiled (Memgraph instances is running in the background with the old module code). Maybe we just add, e.g. `setup_query` and `test_query` (rename `query`)?

e2e/rsmgp_example_test/test_simple/test.yml
gitbuda: It's not critical, but it would be nice to have :smile:

e2e/rsmgp_example_test/test_simple/test.yml
gitbuda: In addition to result values testing, Memgraph is required to figure out is there any memory leakage related to `mgp_memory`, reporting the leakage status is maybe one more feature of the `e2e` harness.

rust/rsmgp-sys/src/vertex/mod.rs
```diff
@@ -0,0 +1,194 @@
...
+            if data.is_null() {
+                None
+            } else {
+                let copy_ptr = ffi::mgp_vertex_copy(data, self.context.memory());
+                if copy_ptr.is_null() {
+                    panic!("Unable to allocate new vertex during vertex iteration.");
+                }
+                Some(Vertex {
+                    ptr: copy_ptr,
+                    context: self.context.clone(),
```
jmatak, line 65: Why cloning here? If context is global and unique?

rust/rsmgp-sys/src/rsmgp.rs
```diff
@@ -0,0 +1,133 @@
...
+pub fn add_read_procedure(
+    proc_ptr: extern "C" fn(*const mgp_list, *const mgp_graph, *mut mgp_result, *mut mgp_memory),
+    name: &CStr,
+    module: *mut mgp_module,
+) -> *mut mgp_proc {
+    unsafe { ffi::mgp_module_add_read_procedure(module, name.as_ptr(), Some(proc_ptr)) }
+}
+
+#[allow(clippy::not_unsafe_ptr_arg_deref)]
+pub fn add_int_result_type(procedure: *mut mgp_proc, name: &CStr) -> MgpResult<()> {
```
jmatak, line 59: This, and other result types can be inserted by providing a MgpValue instance to the Record, looks better from API perspective, rather than calling different methods of insertion e.g. `add_int_result_type`, `add_bool_result_type`

rust/rsmgp-sys/src/vertex/mod.rs
```diff
@@ -0,0 +1,194 @@
...
+        }
+    }
+
+    pub fn property(&self, name: &CStr) -> MgpResult<Property> {
+        let unable_alloc_msg = c_str!("Unable to allocate vertex property.");
+        unsafe {
+            let mgp_value =
+                ffi::mgp_vertex_get_property(self.ptr, name.as_ptr(), self.context.memory());
+            if mgp_value.is_null() {
+                ffi::mgp_result_set_error_msg(self.context.result(), unable_alloc_msg.as_ptr());
```
jmatak, line 122: I think the Error should be propagated to the function that produced it, and their job should be to catch it and send it to the error_result.

rust/rsmgp-sys/src/value/mod.rs
```diff
@@ -0,0 +1,326 @@
...
+//   be called.
+//   * Hold only *mut mgp_value... as soon as there is *const mgp_value, make a copy (own *mut
+//   mgp_value). The same applies for all other data types, e.g. mgp_edge, mgp_vertex.
+//   mgp_value_make_vertex accepts *mut mgp_vartex (required to return user data), but data from
+//   the graph is all *const T.
+
+/// Useful to hold user created `mgp_value` but also to store values coming from Memgraph, e.g.,
+/// property `mgp_value`s.
+#[derive(Debug)]
+pub struct MgpValue {
```
jmatak, line 36: As commented before, this might be renamed to Value, and that would be the only object named Value in the whole module. Example of C++ wrapper: https://github.com/memgraph/mage/blob/T027-MAGE_cpp-wrapper/cpp/mg_procedure/mgp.hpp#L664

rust/rsmgp-sys/src/value/mod.rs
```diff
@@ -0,0 +1,326 @@
...
+    pub fn is_vertex(&self) -> bool {
+        unsafe { ffi::mgp_value_is_vertex(self.ptr) != 0 }
+    }
+
+    pub fn is_edge(&self) -> bool {
+        unsafe { ffi::mgp_value_is_edge(self.ptr) != 0 }
+    }
+}
+
+pub fn make_null_value(context: &Memgraph) -> MgpResult<MgpValue> {
```
jmatak, line 82: This should happen under the hood when calling `MgpResult {value: some_int_value, context: context}`, or however Rust handles object construction

rust/rsmgp-sys/src/value/mod.rs
```diff
@@ -0,0 +1,326 @@
...
+    }
+    Ok(Path {
+        ptr: mgp_path_copy,
+        context: context.clone(),
+    })
+}
+
+/// # Safety
+/// TODO(gitbuda): Write section about safety.
+pub unsafe fn mgp_raw_value_to_value(
```
jmatak, line 272: Result of this should be MgpValue that stores value which can be interpreted different (value.int_value(), value.double_value() ...)

rust/rsmgp-sys/src/vertex/mod.rs
gitbuda: Memgraph (context) is actually not global, because each procedure call has a different context. `context::Memgraph` contains [std::rc::Rc](https://doc.rust-lang.org/std/rc/struct.Rc.html) to `context::MgpMemgraph`, which means there can be `&context::Memgraph` and copies of `context::Memgraph`. In this case, the copy is used, `&` is not convenient because then the lifetime has to be managed (Vertex with the `&` goes out of scope). The current implementation of `context::Memgraph` is the best I can currently think of. Open for suggestions on how to do this better :smile:

rust/rsmgp-sys/src/rsmgp.rs
gitbuda: This is a bit problematic because at the time the result type has to be specified, there is no `MgpValue` (inside `mgp_init_module`). I'll think about how to introduce some "per module" object to hold all registered procedures and types in one place.

rust/rsmgp-sys/src/vertex/mod.rs
```diff
@@ -0,0 +1,194 @@
...
+            Ok(EdgesIterator {
+                ptr: mgp_iterator,
+                context: self.context.clone(),
+                ..Default::default()
+            })
+        }
+    }
+}
+
+pub fn make_graph_vertices_iterator(context: &Memgraph) -> MgpResult<VerticesIterator> {
```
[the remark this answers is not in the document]
gitbuda: Yep, this function will be moved into `context::Memgraph`. I'm also considering refactoring the `rsmgp-sys::context` to `rsmgp-sys::memgraph`. Rename, `context` to `memgraph`. In that case, the code would look like
```
memgraph.vertices()
```
But, yea, :100:

rust/rsmgp-sys/src/value/mod.rs
gitbuda: Yep, good point, it's 100% possible to have one `Value` object, but having 2 allows a couple of nice things:
1. Pure `MgpValue` containing just a pointer can be used internally to destroy `mgp_value` when needed.
2. Pure Rust `Value` could be used in [pattern matching](https://doc.rust-lang.org/book/ch06-00-enums.html).
```
if let Value::Int(value) = value { // do something with int */ }
else if let Value::String(value) = value { /* do something with string */ }
```
I think having both, will simplify the Rust code, but yea, users might be confused with two similar types.
I'll definitely think about what to do with the "value" type.

rust/rsmgp-sys/src/value/mod.rs
gitbuda: `MgpResult` is not a struct, it's a custom type based on [std::result::Result](https://doc.rust-lang.org/std/result/) returned from each call to handle both, the result and any error (take a look under `rsmgp-sys/src/result/mod.rs`). But, valid point, I think I'll move all these methods into the `context::Memgraph`. The final code should look like:
```
memgraph.make_null_value()
```

rust/rsmgp-sys/src/value/mod.rs
gitbuda: Yea, this is I think the same discussion about `MgpValue` and `Value`, once the decision about values will be in place, this will be updated accordingly :smile:

rust/rsmgp-sys/src/property/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+// Property owns CString and Value bacause the underlying C string or value could be deleted during
+// the lifetime of the property. In other words, Property stores copies of underlying name and
+// value.
+#[derive(Debug)]
+pub struct Property {
+    pub name: CString,
+    pub value: Value,
+}
+
+pub struct PropertiesIterator {
```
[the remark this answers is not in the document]
gitbuda: What do you mean here? Could you please also provide some examples? I'm just not 100% sure what you are referring to :smile:

rust/rsmgp-example/Cargo.lock
```diff
@@ -0,0 +1,468 @@
+# This file is automatically @generated by Cargo.
```
antaljanosbenjamin, line 1: This file should be in .gitignore

rust/rsmgp-sys/src/value/mod.rs
```diff
@@ -0,0 +1,370 @@
...
+            Value::Map(x) => MgpValue::make_map(&x, &memgraph),
+            Value::Vertex(x) => MgpValue::make_vertex(&x, &memgraph),
+            Value::Edge(x) => MgpValue::make_edge(&x, &memgraph),
+            Value::Path(x) => MgpValue::make_path(&x, &memgraph),
+        }
+    }
+}
+
+/// # Safety
+/// TODO(gitbuda): Write section about safety.
```
antaljanosbenjamin, line 325: Maybe also emphasize that this function copies the underlying value.

rust/rsmgp-sys/src/value/mod.rs
```diff
@@ -0,0 +1,370 @@
...
+    pub fn to_value(&self, memgraph: &Memgraph) -> MgpResult<Value> {
+        unsafe { mgp_raw_value_to_value(self.mgp_ptr(), &memgraph) }
+    }
+
+    pub fn make_null(memgraph: &Memgraph) -> MgpResult<MgpValue> {
+        unsafe {
+            let mgp_ptr = ffi::mgp_value_make_null(memgraph.memory());
+            if mgp_ptr.is_null() {
+                return Err(MgpError::UnableToMakeNullValue);
+            }
```
antaljanosbenjamin, line 100: It is funny that we return an error if the result is null when we are making a null value... :smile:

rust/rsmgp-sys/src/value/mod.rs
```diff
@@ -0,0 +1,370 @@
...
+        }
+    }
+
+    pub fn is_string(&self) -> bool {
+        unsafe { ffi::mgp_value_is_string(self.ptr) != 0 }
+    }
+
+    /// Required because List implements Drop which destroys the underlying list object. MgpValue
+    /// takes the ownership of the list which means copy has to be created + if all goes well the
+    /// list object shouldn't be deleted.
```
antaljanosbenjamin, line 167: What do you mean by "if all goes well, the list object shouldn't be deleted"? Is this about the `list` argument? because then why would that get deleted? It this is about the new object, then this comment is kind of unnecessary in my opinion.

rust/rsmgp-sys/src/memgraph/mod.rs
```diff
@@ -0,0 +1,183 @@
...
+    graph: *const mgp_graph,
+    result: *mut mgp_result,
+    memory: *mut mgp_memory,
+    module: *mut mgp_module,
+}
+
+/// Main object to interact with Memgraph instance.
+#[derive(Debug, Clone)]
+pub struct Memgraph {
+    memgraph: Rc<MgpMemgraph>,
```
antaljanosbenjamin, line 39: Why `Rc` is needed a for a struct that holds only pointers? Isn't that possible just to copy `MgpMemgraph`?

rust/rsmgp-example/src/lib.rs
```diff
@@ -0,0 +1,153 @@
...
+use rsmgp_sys::rsmgp::*;
+use rsmgp_sys::value::*;
+use rsmgp_sys::vertex::*;
+use std::ffi::CString;
+use std::os::raw::c_int;
+use std::panic;
+
+// TODO(gitbuda): If double free occures, Memgraph crashes -> prevent/ensure somehow.
+
+// Required because we want to be able to propagate Result by using ? operator.
```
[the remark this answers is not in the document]
gitbuda: Yep, that's the case, I'll explain :smile:

rust/rsmgp-sys/src/memgraph/mod.rs
gitbuda: After the discussion, we've decided to remove the whole `MgpValue` and just put all pointers to the Memgraph struct.

rust/rsmgp-sys/src/value/mod.rs
```diff
@@ -0,0 +1,370 @@
...
+//   mgp_value_make_vertex accepts *mut mgp_vartex (required to return user data), but data from
+//   the graph is all *const T.
+//
+// The decision is to move on with a copy of mgp_value because it's cheap to make the copy.
+
+/// Useful to own `mgp_value` coming from / going into Memgraph as a result.
+///
+/// Underlying pointer object is going to be automatically deleted.
+#[derive(Debug)]
+pub struct MgpValue {
```
antaljanosbenjamin, line 66: If the memory would be stored inside MgpValue, then maybe the [From](https://doc.rust-lang.org/std/convert/trait.From.html) and/or [Into](https://doc.rust-lang.org/std/convert/trait.Into.html) traits could be utilized for providing idiomatic rust way for conversions between `MgpValue` and `Value`.

rust/rsmgp-sys/src/vertex/mod.rs
```diff
@@ -0,0 +1,204 @@
...
+    pub(crate) unsafe fn mgp_copy(
+        mgp_vertex: *const mgp_vertex,
+        memgraph: &Memgraph,
+    ) -> MgpResult<Vertex> {
+        // Test passes null ptr because nothing else is possible.
+        #[cfg(not(test))]
+        assert!(
+            !mgp_vertex.is_null(),
+            "Unable to make vertex copy because vertex pointer is null."
+        );
```
antaljanosbenjamin, line 113: Why the `is_null` check is present here but not in the `new` function? I see this pattern in other structs also (edge,map,path). I think ensuring the pointer is not null would be beneficial, because the non-static functions don't do any null checks and calling c function with null pointer is dangerous.

rust/rsmgp-sys/src/vertex/mod.rs
```diff
@@ -0,0 +1,204 @@
...
+
+    pub fn property(&self, name: &CStr) -> MgpResult<Property> {
+        unsafe {
+            let mgp_value =
+                ffi::mgp_vertex_get_property(self.ptr, name.as_ptr(), self.memgraph.memory());
+            if mgp_value.is_null() {
+                return Err(MgpError::UnableToGetVertexProperty);
+            }
+            // TODO(gitbuda): Figure out how to pass Snafu memgraph here.
+            let value = MgpValue::to_value(&MgpValue::new(mgp_value), &self.memgraph)?;
```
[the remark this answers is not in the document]
antaljanosbenjamin: The same happens in edge/mod.rs:160.

rust/rsmgp-sys/src/result/mod.rs
```diff
@@ -0,0 +1,248 @@
...
+use crate::mgp::ffi;
+use mockall_double::double;
+
+pub struct MgpResultRecord {
+    ptr: *mut mgp_result_record,
+    memgraph: Memgraph,
+}
+
+impl MgpResultRecord {
+    pub fn new(memgraph: &Memgraph) -> MgpResult<MgpResultRecord> {
```
antaljanosbenjamin, line 37: I think this function is kind different from e.g.: `Vertex::new` because this function creates a new object, while `Vertex::new` only wraps the pointer. I would use different names for them, maybe using `create` and `wrap` explicitly, but that is a details. The important in my opinion is to differentiate between the two types of functions.

rust/rsmgp-sys/src/property/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    fn next(&mut self) -> Option<Property> {
+        unsafe {
+            let data: *const mgp_property;
+            if self.is_first {
+                self.is_first = false;
+                data = ffi::mgp_properties_iterator_get(self.ptr);
+            } else {
+                data = ffi::mgp_properties_iterator_next(self.ptr);
+            }
```
antaljanosbenjamin, line 77:
```suggestion
            let data = if self.is_first {
                self.is_first = false;
                ffi::mgp_properties_iterator_get(self.ptr)
            } else {
                 ffi::mgp_properties_iterator_next(self.ptr)
            };
```

rust/rsmgp-sys/src/path/mod.rs
```diff
@@ -0,0 +1,121 @@
...
+            let mgp_path = ffi::mgp_path_make_with_start(vertex.mgp_ptr(), memgraph.memory());
+            if mgp_path.is_null() {
+                return Err(MgpError::UnableToCreatePathWithStartVertex);
+            }
+            Ok(Path::new(mgp_path, &memgraph))
+        }
+    }
+
+    /// Fails if the current last vertex in the path is not part of the given edge or if there is
+    /// not memory to expand the path.
```
antaljanosbenjamin, line 88:
```suggestion
    /// Fails if the current last vertex in the path is not part of the given edge or if there is
    /// no memory to expand the path.
```

rust/rsmgp-sys/src/property/mod.rs
antaljanosbenjamin: The same true for Map/Edges/Vertices iterator.

rust/rsmgp-sys/src/property/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+/// Property is used in the following contexts:
+///   * return Property from PropertiesIterator
+///   * return Property directly from vertex/edge.
+///
+/// Property owns CString and Value bacause the underlying C string or value could be deleted
+/// during the lifetime of the property. In other words, Property stores copies of underlying name
+/// and value.
+#[derive(Debug)]
+pub struct Property {
+    pub name: CString,
```
antaljanosbenjamin, line 36: If we decide to go with the "read without copy" approach, then this should be replaced by `CStr`.

antaljanosbenjamin: The same goes for `MapItem::key` .

antaljanosbenjamin:
Have you thought about implementing the [IntoIterator](https://doc.rust-lang.org/std/iter/trait.IntoIterator.html) trait for the iterable types? It would hide the error handling in some way but would make the `for` loops easier.

gitbuda:
> Have you thought about implementing the [IntoIterator](https://doc.rust-lang.org/std/iter/trait.IntoIterator.html) trait for the iterable types? It would hide the error handling in some way but would make the `for` loops easier.

Nope, I don't quite understand how would it make the for loops easier? Could you give a code example? :smile:

antaljanosbenjamin:
> Nope, I don't quite understand how would it make the for loops easier? Could you give a code example? 😄

Yes, this section of the docs makes it clear, I think: https://doc.rust-lang.org/std/iter/index.html#for-loops-and-intoiterator

But I got confused now, because it seems the `IntoIterator` trait requires to move the object, which is not applicable for us. Tomorrow I will spent 15 minutes to check if that can work for us or not.

.github/workflows/test.yml
```diff
@@ -68,3 +71,26 @@ jobs:
...
+    - name: Archive code coverage results
+      uses: actions/upload-artifact@v1
+      with:
+        name: rsmgp-sys-coverage-report
+        path: rust/rsmgp-sys/tarpaulin-report.html
+    - name: Rust example build
+      run: |
+        cd rust/rsmgp-example
+        cargo fmt -- --check
+        cargo build
```
jmatak, line 96: Can you split this in two separate units - building and testing?

e2e/rsmgp_example_test/test_simple/test.yml
jmatak: We don't necessarily need different kind of testing queries, but rather we can call mg.load_all() before each run. That way we can ensure that we are loading the latest version of query modules.

.github/workflows/test.yml
```diff
@@ -68,3 +71,26 @@ jobs:
...
+          cargo install cargo-tarpaulin
+    - name: Run code coverage
+      run: |
+        cd rust/rsmgp-sys
+        RUST_BACKTRACE=1 cargo tarpaulin --out Html --ignore-tests
+    - name: Archive code coverage results
+      uses: actions/upload-artifact@v1
+      with:
+        name: rsmgp-sys-coverage-report
+        path: rust/rsmgp-sys/tarpaulin-report.html
```
jmatak, line 91: Why would you need report in CI?

rust/rsmgp-sys/src/value/mod.rs
gitbuda: Yes, I've added this comment because we can't just wrap the mgp_list pointer to the `List` because then it will be automatically deleted, but yea, the `if all goes well` part is strange. I'll try to rephrase.

rust/rsmgp-example/src/lib.rs
```diff
@@ -0,0 +1,104 @@
...
+use rsmgp_sys::result::*;
+use rsmgp_sys::rsmgp::*;
+use rsmgp_sys::value::*;
+use rsmgp_sys::{close_module, define_procedure, define_type, init_module};
+use std::ffi::CString;
+use std::os::raw::c_int;
+use std::panic;
+
+init_module!(|memgraph: &Memgraph| -> Result<(), MgpError> {
+    memgraph.add_read_procedure(
```
jmatak, line 15: Where are input parameters defined?

rust/rsmgp-sys/src/property/mod.rs
gitbuda: I like this!

.github/workflows/test.yml
gitbuda: CI should usually generate the coverage report (because that's not a required operation like, e.g., running unit tests). But here, the tarpaulin installation takes a couple of minutes. That's why I'm considering removing all that from the CI for now?

.github/workflows/test.yml
gitbuda: What do you mean when you say "split"? `cargo test` also builds the code + here `rsmgp-sys` and `rsmgp-example` are built. Did you mean to just puts an empty line between `rsmgp-sys` and `rsmgp-example`?

rust/rsmgp-example/src/lib.rs
gitbuda: Yea, this is a good point, but that's the `macro` where any [Expression](https://doc.rust-lang.org/reference/expressions.html) can be passed, but if it's wrong, it's going to fail because of the code "inside" `init_module!`. Basically, `|memgraph: &Memgraph| -> Result<(), MgpError>` signature is required, but I don't know how to enforce that because Rust macros just have these "fragment-specifiers" https://doc.rust-lang.org/reference/macros-by-example.html Probably it's possible somehow, but I just don't think it's worth it at this point... But yea, if there is a simple syntax to do so, I can add :smile:

.github/workflows/test.yml
jmatak: It makes sense to have these coverage reports if you use them, but if there is no automatization usage, then there is no reason to have them in Github workflow, but rather run it locally.

.github/workflows/test.yml
jmatak: No, I meant that building the code should be inside `name: Set up Rust`, while running the test should be under `name: Rust library tests`. This is only for the convenience of someone who is looking in those actions on Github

e2e/rsmgp_example_test/test_simple/test.yml
gitbuda: This is fine because I can reload locally manually, and in the CI it's not required.

rust/rsmgp-sys/src/value/mod.rs
gitbuda: I think I'll keep both of these objects for now. @jmatak let me know if that still doesn't make sense :smile:

rust/rsmgp-sys/src/property/mod.rs
gitbuda: We've abandoned the idea with 0 copies because of an issue with the lifetime of C and Rust objects. Take a look into the `mgp_raw_value_to_value` function comment.

.github/workflows/test.yml
gitbuda: Ok, will remove the uploading of the report file :smile:

.github/workflows/test.yml
gitbuda: Does the current state make sense?

rust/rsmgp-sys/src/value/mod.rs
gitbuda: Only `impl From<MgpValue> for Value` makes sense, but it's also not super nice to use it inside, e.g. vertex property, because `MgpValue` has to be created also there so it's I think more compact to just use `MgpValue`'s `to_value` method. + using this hides the `MgpResult` from getting retuned...

That's all for the library code, but for the user code, the `From` implementation makes sense :smile:

gitbuda:
> Have you thought about implementing the [IntoIterator](https://doc.rust-lang.org/std/iter/trait.IntoIterator.html) trait for the iterable types? It would hide the error handling in some way but would make the `for` loops easier.

I think I'm going to quit this idea as well because e.g., `ListIterator` has `&List` that needs the lifetime specifier. While implementing the `IntoIterator` trait, it's not possible to specify the lifetime...

rust/rsmgp-example/src/lib.rs
```diff
@@ -1,116 +1,131 @@
...
     Ok(())
 });
 
-define_procedure!(basic, |memgraph: &Memgraph| -> Result<(), MgpError> {
+define_procedure!(basic, |memgraph: &Memgraph| -> MgpResult<()> {
     let result = memgraph.result_record()?;
-    result.insert_string(c_str!("input_string"), c_str!("TODO: Add input param."))?;
+    let args = memgraph.args()?;
+    let output_string = args.value_at(0)?;
+    let output_int = args.value_at(1)?;
```
antaljanosbenjamin, line 52: This is a bit confusing: I would rename them to `input_....` (to show that this is how input parameters can be read) and I would add a small comment about this is just forwards the input parameters as outputs.

rust/rsmgp-sys/src/edge/mod.rs
```diff
@@ -116,12 +116,12 @@ impl Edge {
         #[cfg(not(test))]
         assert!(
             !ptr.is_null(),
-            "Unable to make edge copy because edge pointer is null."
+            "Unable to create edge copy because the given pointer is null."
```
antaljanosbenjamin, line 119: If we would like to nitpick, then I would suggest `Unable to copy edge because the given pointer is null.`, but I am also fine with this message, if you like the current one more.

antaljanosbenjamin: I just realized this error message is in sync with the regarding `MgpError` so it is totally fine to leave this as is.

rust/rsmgp-sys/src/result/mod.rs
```diff
@@ -137,44 +134,72 @@ pub enum MgpError {
     #[snafu(display("Unable to access list value by index."))]
     UnableToAccessListValueByIndex,
 
+    // MAP
+    #[snafu(display("Unable to copy map."))]
+    UnableToCopyMap,
```
antaljanosbenjamin, line 139: Why map/list is using `UnableToCopyXXX` while edge/vertex/path uses `UnableToCreateXXXCopy`? Is this intentional?

gitbuda: It's not, it's just inconsistent, I'll update everything to the shortest version `UnableToCopyXYZ` :smile:

antaljanosbenjamin (approved):

build
```diff
@@ -85,3 +86,23 @@ ignore_list = [
...
+#################################################
+#              Load Rust modules                #
+#################################################
+
+# List all build modules in a list explicitly.
+# NOTE: Cargo prefixes `.so` files with the `lib` prefix (not possible to
+#       change).
+# NOTE: There are more files with the suffix `libmodule_name.so` under the Rust
+#       build directory.
+RS_MODULES = [
```
jmatak, line 99: If non relevant libnames.so are under the Rust build directory, it would be more suitable to put them in ignore, as in script above. Also, if each time library starts with "lib", there could be a rule to rename it, rather than giving the name explicitely.

jmatak (commented):
Before approving I have just one detail to fix. For the purpose of convenience, I think there should be a better way of automatically pulling Rust modules, and ignoring additionally created files

jmatak (approved):
MAGE becomes a bit rusty :old_key:

PR #16, opened 2021-03-24
C++ Memgraph Utility

jmatak:
Implementation of a C++ Memgraph utility. A simple use case is visible in `connectivity_module`.


cpp/connectivity_module/connectivity_module.cpp
```diff
@@ -1,136 +1,94 @@
 #include <queue>
 #include <unordered_map>
 
-#include "mg_procedure.h"
-#include "utils/on_scope_exit.hpp"
+#include <mg_exceptions.hpp>
+#include <mg_procedure.h>
+#include <mg_utils.hpp>
```
antonio2368, line 6: Add empty line

cpp/mg_utility/mg_graph.hpp
```diff
@@ -0,0 +1,135 @@
...
+#include <map>
+#include <unordered_map>
+
+namespace mg_graph {
+
+/// Graph representation.
+class Graph : public GraphView {
+public:
+  /// Create object Graph
+  explicit Graph();
```
antonio2368, line 17: if it's empty you can 
```
explicit Graph() = default;
```

cpp/mg_utility/mg_graph.cpp
```diff
@@ -0,0 +1,152 @@
+#include "mg_graph.hpp"
+#include "mg_exceptions.hpp"
+#include <algorithm>
```
antonio2368, line 3: Try separating includes into groups.
E.g. project headers and STL headers.
Just add new line before `#include <algorithm>`

cpp/mg_utility/mg_graph.hpp
```diff
@@ -0,0 +1,135 @@
...
+
+  /// Destroys the object.
+  ~Graph();
```
antonio2368, line 20: Also `= default`;

cpp/mg_utility/mg_utils.cpp
```diff
@@ -0,0 +1,170 @@
+#include <iostream>
+#include <map>
+#include <optional>
+#include <sstream>
+#include <unordered_set>
+
+#include "mg_exceptions.hpp"
+#include "mg_graph.hpp"
+#include "mg_procedure.h"
+#include "mg_utils.hpp"
```
antonio2368, line 10: Be consistent with order of the groups.

antonio2368 (commented):
This looks really clean!

cpp/mg_utility/mg_graph.hpp
```diff
@@ -0,0 +1,264 @@
...
+
+  using TNode = Node<TSize>;
+  using TEdge = Edge<TSize>;
+  using TNeighbour = Neighbour<TSize>;
+
+  static_assert(
+      std::is_same<TSize, uint8_t>::value ||
+          std::is_same<TSize, uint16_t>::value ||
+          std::is_same<TSize, uint32_t>::value ||
+          std::is_same<TSize, uint64_t>::value,
```
antonio2368, line 27: https://en.cppreference.com/w/cpp/types/is_unsigned
```suggestion
      std::is_unsigned_v<TSize>,
```

cpp/mg_utility/data_structures/graph_data.hpp
```diff
@@ -0,0 +1,34 @@
...
+/// This file contains the graph data - nodes, edges, variables.
+
+#pragma once
+
+namespace mg_graph {
+
+/// Node representation.
+///
+/// Node has id and type. Id must be unique in the graph.
+template <typename TSize> struct Node { TSize id; };
```
antonio2368, line 11: I would give them a default value of `uint64_t`. Thoughts?
(`template <typename TSize = uint64_t>`)

cpp/mg_utility/mg_graph.hpp
```diff
@@ -0,0 +1,264 @@
...
+#include "mg_exceptions.hpp"
+
+namespace mg_graph {
+
+/// Graph representation.
+template <typename TSize> class Graph : public GraphView<TSize> {
+
+  using TNode = Node<TSize>;
+  using TEdge = Edge<TSize>;
+  using TNeighbour = Neighbour<TSize>;
```
antonio2368, line 21: I think you can do something like
```
using typename Node<TSize>::TNode;
```
It's equally long but I think it's safer so the aliases are in sync.

cpp/mg_utility/mg_graph.hpp
```diff
@@ -0,0 +1,264 @@
...
+  ///
+  /// @return     Created node id
+  TSize CreateNode(uint64_t memgraph_id) {
+    TSize id = nodes_.size();
+    nodes_.push_back({id});
+    adj_list_.emplace_back();
+    neighbours_.emplace_back();
+
+    inner_to_memgraph_id_.insert(std::pair(id, memgraph_id));
+    memgraph_to_inner_id_.insert(std::pair(memgraph_id, id));
```
antonio2368, line 121:
```suggestion
    inner_to_memgraph_id_.emplace(id, memgraph_id);
    memgraph_to_inner_id_.emplace(memgraph_id, id);
```
?

cpp/mg_utility/mg_graph.hpp
```diff
@@ -0,0 +1,264 @@
...
+    neighbours_.clear();
+
+    memgraph_to_inner_id_.clear();
+    inner_to_memgraph_id_.clear();
+  }
+
+private:
+  // Constant is used for marking deleted edges.
+  // If edge id is equal to constant, edge is deleted.
+  static const TSize kDeletedEdgeId = std::numeric_limits<TSize>::max();
```
antonio2368, line 252: make this `static const inline` or `static constexpr` (prefer this) because of the ODR, (if you add this header into multiple translation units and try to link them you'll get annoying and hard to parse errors)

cpp/mg_utility/data_structures/graph_data.hpp
jmatak: If defaulting to `uint64_t`, would the user be able to write something like: `mg_graph::Graph *graph;` (without explicitly writing the template parameter). If yes, I am definitely into the idea of setting the default value.

cpp/mg_utility/mg_graph.hpp
jmatak: This breaks my code with the following exception:
`error: type ‘mg_graph::Node<long unsigned int>’ is not a base type for type ‘mg_graph::GraphView<long unsigned int>` Are you sure this is the right way?

cpp/mg_utility/data_structures/graph_data.hpp
antonio2368: Yep, that would be the point of it. Makes it easier for the most frequent use case and you allow template arg for specific stuff.

cpp/mg_utility/mg_graph.hpp
antonio2368: I meant the base class:
```
using typename GraphView<TSize>::TNode;
```
Sorry!

cpp/connectivity_module/connectivity_module.cpp
```diff
@@ -1,136 +1,95 @@
...
+      vertex_component[vertex.id] = curr_component;
+      while (!q.empty()) {
+        uint64_t v_id = q.front();
+        q.pop();
 
-      mgp_edges_iterator_destroy(edges_iterator);
-    }
+        // Iterate over inbound edges.
+        for (mg_graph::Neighbour neihgbor : graph->Neighbours(v_id)) {
+          uint64_t next_id = neihgbor.node_id;
```
antonio2368, line 54: prefer `auto` so you don't have to search and replace when you change the underlying type.  :smile:

cpp/connectivity_module/connectivity_module.cpp
```diff
@@ -1,136 +1,95 @@
...
-        }
-      });
+  try {
+    mg_graph::Graph<> *graph =
+        mg_utility::GetGraphView(memgraph_graph, result, memory);
 
-      q.pop();
+    std::unordered_map<uint64_t, uint64_t> vertex_component;
+    int64_t curr_component = 0;
+    for (mg_graph::Node vertex : graph->Nodes()) {
```
antonio2368, line 39: here also, use `auto`
```suggestion
    for (auto vertex : graph->Nodes()) {
```

cpp/mg_utility/mg_utils.hpp
```diff
@@ -0,0 +1,209 @@
...
+#include "mg_graph.hpp"
+#include "mg_procedure.h"
+
+namespace mg_graph {
+
+template <typename TSize>
+void CreateGraphNode(mg_graph::Graph<TSize> *graph, const mgp_vertex *vertex) {
+  // Get Memgraph internal ID property
+  auto id_val = mgp_vertex_get_id(vertex);
+  TSize memgraph_id = id_val.as_int;
```
antonio2368, line 19: You mean auto?
Because this will trim to the TSize and not use the memgraphs uint64_t as it should. Be careful now what is of which type.

jmatak: Great catch. Need to check twice that part of the code again before merging.

cpp/connectivity_module/connectivity_module.cpp
```diff
@@ -1,136 +1,94 @@
...
-        if (v != nullptr) {
-          mgp_vertex_destroy(v);
-        }
-      });
+  try {
+    auto *graph = mg_utility::GetGraphView(memgraph_graph, result, memory);
 
-      q.pop();
+    std::unordered_map<uint64_t, uint64_t> vertex_component;
+    uint64_t curr_component = 0;
```
antonio2368, line 37: I forgot about this little detail, do we go with the `std::uint64_t` (and other similar types also)?

cpp/mg_utility/data_structures/graph_data.hpp
```diff
@@ -0,0 +1,34 @@
...
+};
+
+/// Neighbour representation.
+///
+/// Helper structure for storing node and edge id.
+template <typename TSize = uint64_t> struct Neighbour {
+  TSize node_id;
+  TSize edge_id;
+  Neighbour(TSize n_id, TSize e_id) : node_id(n_id), edge_id(e_id) {}
+};
```
antonio2368, line 33: static_assert in all of the structs for is_unsigned_v

cpp/mg_utility/data_structures/graph_view.hpp
```diff
@@ -0,0 +1,72 @@
...
+#include "graph_data.hpp"
+
+namespace mg_graph {
+
+/// Graph view interface.
+///
+/// Interface provides methods for fetching graph data.
+/// There are two methods for changing variables on the edges:
+/// SetVariableState and SetVariableValue
+template <typename TSize = uint64_t> class GraphView {
```
antonio2368, line 14: static assert for the type

cpp/mg_utility/mg_utils.hpp
```diff
@@ -0,0 +1,209 @@
...
+  graph->CreateNode(memgraph_id);
+}
+
+template <typename TSize>
+void CreateGraphEdge(mg_graph::Graph<TSize> *graph,
+                     const mgp_vertex *vertex_from,
+                     const mgp_vertex *vertex_to) {
+  // Get Memgraph internal ID property
+  TSize memgraph_id_from = mgp_vertex_get_id(vertex_from).as_int;
+  TSize memgraph_id_to = mgp_vertex_get_id(vertex_to).as_int;
```
antonio2368, line 30:
```suggestion
  auto memgraph_id_from = mgp_vertex_get_id(vertex_from).as_int;
  auto memgraph_id_to = mgp_vertex_get_id(vertex_to).as_int;
```

cpp/mg_utility/mg_utils.hpp
```diff
@@ -0,0 +1,209 @@
...
+/// with iterative node indices. Graph object stores information about node and
+/// edge mappings, alongside with connection information
+///
+///@param graph Memgraph graph
+///@param result Memgraph result object
+///@param memory Memgraph storage object
+///@return mg_graph::Graph
+///
+template <typename TSize = uint64_t>
+mg_graph::Graph<TSize> *GetGraphView(const mgp_graph *memgraph_graph,
```
antonio2368, line 74: Avoid raw pointers as much as possible.
You can return `std::unique_ptr<mg_graph::Graph<TSize>>` here and bellow initialize with 
```
auto graph = std::make_unique<mg_graph::Graph<TSize>>();
```

cpp/connectivity_module/connectivity_module.cpp
jmatak: @gitbuda What is your opinion on this one, since I am no C++ expert?

antonio2368 (approved):
Beautiful

PR #17, opened 2021-03-24
Biconnected components module implementation

jmatak:
Both implementation and unit-tests for biconnected components algorithm.


cpp/biconnected_components_module/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+set(biconnected_components_module_src
```
jmatak, line 1: Change the name of algorithm. `"CALL cpp_biconnected_components_module.biconnected_components()"` is way too long.

cpp/CMakeLists.txt
```diff
@@ -22,12 +22,48 @@ set(CMAKE_CXX_FLAGS_RELWITHDEBINFO
 set(CMAKE_CXX_FLAGS_RELEASE "-O2 -DNDEBUG")
 set(CMAKE_SHARED_LIBRARY_PREFIX "")
 
+# Install testing dependencies
+find_package(Threads REQUIRED)
+
+# Install gtest
```
jmatak, line 28: Install dependencies with respect to different flags: RELEASE/DEBUG/TEST

.github/workflows/test.yml
```diff
@@ -30,21 +30,33 @@ jobs:
     - name: Run Memgraph MAGE image
       run: |
         docker run -d -p 7687:7687 memgraph:mage --telemetry-enabled=False
+    - name: Set up C++
```
jmatak, line 33: Should we run tests in Docker? Requires 2 Docker images, first for Memgraph query modules, second only with algorithms?

.github/workflows/test.yml
```diff
@@ -30,21 +30,33 @@ jobs:
...
     - name: Set up Python
       uses: actions/setup-python@v2
       with:
         python-version: '3.7'
-    - name: Install dependencies
-      run:
+    - name: Build C++ modules
+      run: |
+        mkdir -p cpp/build
+        (cd cpp/build && cmake .. && make -j8)
```
antonio2368, line 44: Why do you put everything in one row if you can change them just by calling the cmds row by row

antonio2368 (commented):
Some setup comments first.

cpp/CMakeLists.txt
jmatak: What is the purpose of every time building with tests on MAGE? If someone is purely interested in running the MAGE, tests shouldn't be something that is generated once the build is initiated?

.github/workflows/test.yml
jmatak: I propose making this efficient in the separate ease of use PR, and wait till that one is solved?

antonio2368: For sure!

cpp/CMakeLists.txt
antonio2368: As the users will build on very different machines, you want to ensure them that they did everything correctly (and that the modules can be correctly built on their machines). Also, you copy the libraries in a different folder so the extra test files shouldn't matter. 
If the user still has some problems with the module, we can be more confident that there is a bug in the implementation of that specific module.

cpp/biconnected_components_module/algorithm/biconnected_components.cpp
```diff
@@ -0,0 +1,75 @@
...
+    edge_stack->push(edge);
+    BccDFS(next_id, node_id, state, edge_stack, bcc, graph);
+    state->low_link[node_id] = std::min(state->low_link[node_id], state->low_link[next_id]);
+
+    // Articulation point check
+    if ((root && root_count > 1) ||  // special case for root
+        (!root && state->low_link[next_id] >= state->discovery[node_id])) {
+      // obsdata::Edges until (node, next) on stack form a BCC
+      bcc->emplace_back();
+      while (edge_stack->top().id != edge.id) {
```
antonio2368, line 36: What if the stack is empty? Can it happen?

cpp/biconnected_components_module/algorithm/biconnected_components.cpp
```diff
@@ -0,0 +1,75 @@
...
+  }
+}
+
+std::vector<std::vector<mg_graph::Edge<>>> bcc_algorithm::GetBiconnectedComponents(const mg_graph::GraphView<> *graph) {
+  bcc_utility::NodeState state;
+  size_t node_size = graph->Nodes().size();
+  state.visited.resize(node_size, false);
+  state.discovery.resize(node_size, 0);
+  state.low_link.resize(node_size, 0);
+  state.counter = 0;
```
antonio2368, line 52: Create a constructor that accepts the size for this.

cpp/biconnected_components_module/algorithm/biconnected_components.hpp
```diff
@@ -0,0 +1,44 @@
...
+///@param node_id Starting Node ID
+///@param parent_id Parental Node ID
+///@param state Current node state
+///@param edge_stack Current state of edge stack
+///@param bcc Current biconnected components
+///@param graph Graph to work on
+///
+void BccDFS(std::uint64_t node_id, std::uint64_t parent_id, bcc_utility::NodeState *state,
+            std::stack<mg_graph::Edge<>> *edge_stack, std::vector<std::vector<mg_graph::Edge<>>> *bcc,
+            const mg_graph::GraphView<> *graph);
```
antonio2368, line 30: const reference for graph?

cpp/biconnected_components_module/biconnected_components_test.cpp
```diff
@@ -0,0 +1,283 @@
...
+  auto G = std::make_unique<mg_graph::Graph<>>();
+  for (std::uint64_t i = 0; i < nodes; ++i) G->CreateNode(i);
+  for (auto &p : edges) G->CreateEdge(p.first, p.second);
+
+  return G;
+}
+
+/// Generates random undirected graph with a given numer of nodes and edges.
+/// The generated graph is not picked out of a uniform distribution.
+inline std::unique_ptr<mg_graph::Graph<>> GenRandomGraph(std::uint64_t nodes, std::uint64_t edges) {
```
antonio2368, line 39: Why do you inline these functions?

cpp/biconnected_components_module/biconnected_components_test.cpp
```diff
@@ -0,0 +1,283 @@
...
+  std::uniform_int_distribution<std::uint64_t> dist(0, nodes - 1);
+  std::set<std::pair<std::uint64_t, std::uint64_t>> E;
+  for (std::uint64_t i = 0; i < edges; ++i) {
+    std::int64_t u, v;
+    do {
+      u = dist(rng);
+      v = dist(rng);
+      if (u > v) std::swap(u, v);
+    } while (u == v || E.find({u, v}) != E.end());
+    E.insert({u, v});
```
antonio2368, line 51:
```suggestion
  for (std::uint64_t i = 0; i < edges; ++i) {
    std::optional<std::pair<std::uint64_t, std::uint64_t>> edge;
    do {
      edge = std::minmax(dist(rng), dist(rng));
    } while (edge->first == edge->second || E.find(*edge) != E.end());
    E.insert(*edge);
```

cpp/biconnected_components_module/biconnected_components_test.cpp
```diff
@@ -0,0 +1,283 @@
...
+  }
+  return BuildGraph(nodes, std::vector<std::pair<std::uint64_t, std::uint64_t>>(E.begin(), E.end()));
```
antonio2368, line 53:
```suggestion
  return BuildGraph(nodes, {E.begin(), E.end()});
```

cpp/biconnected_components_module/biconnected_components_test.cpp
antonio2368: You can also make an alias for the std::pair

cpp/biconnected_components_module/algorithm/biconnected_components.cpp
jmatak: Since Dino was the author of this algorithm I am not sure, because I haven't jumped into implementation details. However, in line 27, there is adding into a stack which means it will not be empty when recursion continues after line 29. Finally, I will include a check of not being empty in the if-clause.

cpp/biconnected_components_module/algorithm/biconnected_components.hpp
jmatak: Why shouldn't it be const, nothing within the graph should be changed inside the function?

cpp/biconnected_components_module/biconnected_components_test.cpp
jmatak: Previous implementation error I think. Have no clue what is the reason for this function being inlined.

cpp/biconnected_components_module/biconnected_components_test.cpp
jmatak: Where does this alias go in common practice? Inside of the header or implementation file?

cpp/biconnected_components_module/algorithm/biconnected_components.hpp
antonio2368:
```suggestion
            const mg_graph::GraphView<> &graph);
```
I meant this.

cpp/biconnected_components_module/biconnected_components_test.cpp
antonio2368: you can have it locally in this function for example, if you add it globally you could have naming conflicts and a lot of compiler troubles.

cpp/biconnected_components_module/algorithm/biconnected_components.cpp
antonio2368: Oh, I see, we have a guarantee that at least the lowest member will be `== edge.id`.

cpp/biconnected_components_module/biconnected_components_module.cpp
```diff
@@ -0,0 +1,77 @@
+#include <iostream>
+#include <map>
+#include <optional>
+#include <sstream>
+#include <unordered_set>
```
antonio2368, line 5: Don't think you need all of these headers... (`iostream` for example)

.github/workflows/test.yml
antonio2368: Why 2 docker images?

cpp/biconnected_components_module/algorithm/biconnected_components.hpp
```diff
@@ -0,0 +1,58 @@
...
+
+/// Simple struct that keeps the state of nodes in algorithms that rely on
+/// DFS traversal.
+struct NodeState {
+  std::vector<bool> visited;
+  std::vector<std::uint64_t> discovery, low_link, depth;
+  std::vector<std::uint64_t> parent;
+  uint64_t counter;
+
+  NodeState(std::uint64_t number_of_nodes) {
```
antonio2368, line 17: Make it explicit. Good practice to use that keyword as much as possible so you don't create objects when you didn't want to.

cpp/biconnected_components_module/biconnected_components_test.cpp
```diff
@@ -0,0 +1,284 @@
...
+};
+
+/// Builds the graph from a given number of nodes and a list of edges.
+/// Nodes should be 0-indexed and each edge should be provided in both
+/// directions.
+std::unique_ptr<mg_graph::Graph<>> BuildGraph(std::uint64_t nodes,
+                                              std::vector<std::pair<std::uint64_t, std::uint64_t>> edges) {
+  auto G = std::make_unique<mg_graph::Graph<>>();
+  for (std::uint64_t i = 0; i < nodes; ++i) G->CreateNode(i);
+  for (auto &p : edges) G->CreateEdge(p.first, p.second);
```
antonio2368, line 32: `for (const auto [from, to] : edges)` or whatever name fits the best here.
Same in the other places, better than using those weird `first` and `second` names from pair. :smile:

cpp/biconnected_components_module/algorithm/biconnected_components.hpp
jmatak: After reading about `explicit`, is there an example where we do not want to use it?

antonio2368: I would correct myself a bit and say use the keyword ALWAYS for the single argument constructors to avoid accidental conversions.

For example, for the copy constructor it doesn't make much sense to have it explicit in any case.

For user defined constructors that take multiple arguments `explicit` will force you to define the name of function every time and you can't use the shorthand version without the class name.
https://godbolt.org/z/n6GchhfdY

cpp/biconnected_components_module/algorithm/biconnected_components.cpp
```diff
@@ -0,0 +1,69 @@
...
+        edge_stack->pop();
+      }
+      bcc->back().push_back(edge_stack->top());
+      edge_stack->pop();
+    }
+  }
+}
+
+std::vector<std::vector<mg_graph::Edge<>>> bcc_algorithm::GetBiconnectedComponents(const mg_graph::GraphView<> &graph) {
+  size_t number_of_nodes = graph.Nodes().size();
```
antonio2368, line 45: `std::size_t`

antonio2368 (approved):
Looks fine to me.

PR #18, opened 2021-03-25
Graph coloring algorithm

SuzanaPratljacic:


python/mage/graph_coloring_module/operators/mutations/random_mutation.py
```diff
@@ -0,0 +1,52 @@
...
+    def mutate(
+        self, graph: Graph, indv: Individual, parameters: Dict[str, Any] = None
+    ) -> Tuple[Individual, List[int]]:
+        """Mutate the given individual and returns the new individual and nodes that were changed."""
+
+        random_mutation_probability = param_value(
+            graph, parameters, "random_mutation_probability"
+        )
+        random_mutation_probability_2 = param_value(
+            graph, parameters, "random_mutation_probability_2"
```
jmatak, line 30: Try validating this other probability too. In  `validate`

python/mage/graph_coloring_module/population_factory.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+logger = logging.getLogger("telco")
+
+
+@validate("population_size", "no_of_colors")
+def generate_individuals(
+    graph: Graph, parameters: Dict[str, Any] = None
+) -> List[Individual]:
+    """Creates a list of individuals in which some individuals are the result of the given algorithms.
```
jmatak, line 20: What is the reason of providing a list of algorithms?

jmatak (changes requested):
I have made some requests mostly on the architecture and Python side since I am not familiar enough to comment on the algorithm and optimization part.

python/graph_coloring.py
```diff
@@ -43,9 +43,9 @@ def _run_algorithm(graph: Graph, parameters) -> List[int]:
         "no_of_processes": 1,
         "no_of_chunks": 1,
         "communication_delay": 10,
-        "max_iterations": max_iterations,
+        "max_iterations": 10,
```
jmatak, line 46: These are default parameters, where are input parameters used in this chunk of code?

jmatak (changes requested):

python/mage/graph_coloring_module/population_factory.py
SuzanaPratljacic: It is a way to give initial solutions. Greedy algorithms can be used to initialize the initial population in order to improve the convergence of a population-based algorithm. Each algorithm in the list represents instructions for initializing one individual.

python/mage/graph_coloring_module/algorithms/meta_heuristics/parallel_algorithm.py
```diff
@@ -0,0 +1,127 @@
...
+    ) -> None:
+        """A function that executes an algorithm."""
+        pass
+
+    def _get_queues(
+        self, ind: int, no_of_chunks: int, queues: mp.Queue
+    ) -> Tuple[Optional[mp.Queue], Optional[mp.Queue], Optional[mp.Queue]]:
+        """Returns my_q, prev_q, next_q for the process on a given index."""
+        if no_of_chunks == 1:
+            return None, None, None
```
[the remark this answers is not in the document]
SuzanaPratljacic: It is used in functions for writing and reading messages. When the queue is None it means that there is no need to send messages because we only have one chunk of the population.

python/mage/graph_coloring_module/components/chain_chunk.py
```diff
@@ -0,0 +1,82 @@
...
+        super().__init__(graph, individuals)
+        self._prev_indv = prev_indv
+        self._next_indv = next_indv
+        self._set_correlations()
+
+    def _get_prev_correlation_ind(self, ind: int) -> int:
+        """Returns the index of the correlation with the previous individual in the chain of individuals."""
+        return ind - 1 if ind - 1 >= 0 else self.size
+
+    def _get_next_correlation_ind(Self, ind: int) -> int:
```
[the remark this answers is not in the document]
SuzanaPratljacic: That function returns the index of the correlation between the individual on index ind and the next individual in the chain of individuals. The correlation stored on the index ind is the correlation between individual on index ind and individual on index ind+1.

python/graph_coloring.py
```diff
@@ -1,19 +1,23 @@
...
+    return getattr(mage.graph_coloring_module, name)
 
 
 @mgp.read_proc
 def color_graph(
     context: mgp.ProcCtx, parameters: mgp.Map
 ) -> mgp.Record(node=str, color=str):
+    """
+    Example:
+    CALL graph_coloring.color_graph({}) YIELD *;
```
jmatak, line 19: If you don't want your procedure call to look like this, simply put the default parameter in the function call
```parameters: mgp.Map = {}```
This would look much better when calling with default parameters:
```CALL graph_coloring.color_graph() YIELD *;```

File: python/mage/graph_coloring_module/population_factory.py [hunk 7]
```diff
@@ -0,0 +1,80 @@
...
+        [
+            Individual(no_of_colors, graph)
+            for _ in range(population_size - len(individuals))
+        ]
+    )
+    return individuals
+
+
+@validate("population_size", "no_of_chunks")
+def create(graph: Graph, parameters: Dict[str, Any] = None) -> Optional[Population]:
```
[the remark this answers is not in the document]
jbajic: Sry for the delay, forgot to push the comment, written it 3 days ago...

python/graph_coloring.py
```diff
@@ -0,0 +1,176 @@
...
+    nodes = []
+    adj_list = defaultdict(list)
+
+    for v in vertices:
+        context.check_must_abort()
+        nodes.append(v.id)
+
+    for e in edges:
+        context.check_must_abort()
+        weight = e.properties.get("weight", 1)
```
jmatak, line 168: This might as well be another module parameter (the name of property between nodes). Otherwise this is great solution by defaulting to 1.

python/mage/graph_coloring_module/population_factory.py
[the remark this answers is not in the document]
jmatak: Thanks for the reply. This is simply for the purpose of extensibility, meaning some other population method doesn't necessarily require `population_size` and `no_of_chunks` parameters. Also, both parameters are validated when entering the function's scope.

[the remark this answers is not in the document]
SuzanaPratljacic: It seems that for some components it is not easy to define a common interface. For example, an interface for mutation, because different implementations of the mutation have different parameters that we don't know in advance.

jmatak (approved):

jbajic (approved):

Josipmrden (approved):

python/mage/graph_coloring_module/algorithms/meta_heuristics/quantum_annealing.py
```diff
@@ -0,0 +1,168 @@
...
+from mage.graph_coloring_module.algorithms.meta_heuristics.parallel_algorithm import (
+    ParallelAlgorithm,
+)
+
+
+logger = logging.getLogger("graph_coloring")
+
+
+class QA(ParallelAlgorithm):
+    """A class that represents quantum annealing algorithm."""
```
MasterMedo, line 21: *the* quantum

python/mage/graph_coloring_module/components/chain_chunk.py
```diff
@@ -0,0 +1,126 @@
...
+        """Returns the index of the correlation with the next individual in the chain of individuals."""
+        return ind
+
+    def get_prev_individual(self, ind: int) -> Individual:
```
MasterMedo, line 82: any reason the variable is called *ind* instead of *i*, or *index*?

python/mage/graph_coloring_module/components/chain_chunk.py
```diff
@@ -0,0 +1,126 @@
...
+    def set_prev_individual(self, indv: Individual) -> None:
+        """Sets the unit that precedes the current piece of chain."""
+        self._cumulative_correlation -= self._correlation[self.size]
+        self._correlation[self.size] = self._calculate_correlation(
+            indv, self.individuals[0]
+        )
+        self._cumulative_correlation += self._correlation[self.size]
+        self._prev_indv = indv
+
+    def set_next_individual(self, indv: Individual) -> None:
```
MasterMedo, line 119: rename the variable: indv -> individual

MasterMedo (commented):
Overall the code quality seems good, I'm in a spot of bother regarding code styling though.
Variable names should be unambiguous, I shouldn't have to infer their meaning, write full names not abbreviations.
Documentation is well written, gives a good insight in what the functions/classes do without getting into details.

On the other hand the code structure is super complex, often layers upon layers of nested function calls, I dislike that greatly but if that's the standard of the code base I guess it's too late to argue about it.

I strongly advise you to read pep 8 down to the letter and follow it diligently in the future.

Great (first?) PR!

e2e/graph_coloring_test/test_graph_init_algorithms/test.yml
```diff
@@ -0,0 +1,10 @@
+query: >
+    CALL graph_coloring.color_graph(
+        {init_algorithms: ["SDO", "LDO"]})
+        YIELD node, color
+    MATCH (a:Cell) WHERE toString(id(a)) = node 
+    RETURN COUNT(a.id) AS id_count, COUNT(color) AS color_count;
```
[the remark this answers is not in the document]
jmatak: This is a good point. I think the best idea would be to put this in new PR where only this issue is raised.

python/mage/graph_coloring_module/algorithms/greedy/SDO.py
```diff
@@ -0,0 +1,61 @@
...
+from mage.graph_coloring_module.utils.validation import validate
+from mage.graph_coloring_module.parameters import Parameter
+
+
+class SDO(Algorithm):
+    """A class that represents SDO greedy algorithm. This algorithm sorts nodes considering
+    their saturation degrees and then colors them sequentially. The saturation degree of a
+    node is defined as the number of its differently colored neighbors. If it is not possible
+    to uniquely determine the color, color is chosen randomly. If coloring the node with any
+    possible color would cause conflicts, then the color is chosen randomly."""
```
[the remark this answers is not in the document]
jmatak: Again, another PR with the suggestion of a format checker?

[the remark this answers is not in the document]
MasterMedo: Mmmm sure, btw I think reading some peps should definitely be a part of required reading before developing in python. I see the core team has a nice notion page about required literature. I propose creating our own.

PR #21, opened 2021-04-06
[T017 <- T024] Organize building tool

jmatak:


.github/workflows/test.yml
```diff
@@ -41,22 +41,32 @@ jobs:
     - name: Build C++ modules
       run: |
         mkdir -p cpp/build
-        (cd cpp/build && cmake .. && make -j8)
+        pushd cpp/build
+        cmake ..
+        make -j8
+        popd
```
antonio2368, line 47: I think github actions start each step from the root directory, so you don't have to return at the end.
```suggestion
        cd cpp/build
        cmake ..
        make -j8
```

.github/workflows/test.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     - name: Download Memgraph Docker image
       if: steps.cache-memgraph-docker.outputs.cache-hit != 'true'
       env:
-        MG_VERSION: '1.3.0'
+        MG_VERSION: '1.4.0'
```
antonio2368, line 20: move env above steps and use the MG_VERSION everywhere where a version is mentioned. In your case, check the cache step, you use 1.3.0 for the key while you changed to 1.4.0 here. 
Instead, you can write
```
key: cache-memgraph-v${{ env.MG_VERSION }}-docker-image
```

build
```diff
@@ -0,0 +1,86 @@
...
+
+# Make build directory
+CPP_BUILD_DIRECTORY = f"{CPP_DIRECTORY}/build"
+os.makedirs(CPP_BUILD_DIRECTORY, exist_ok=True)
+
+# Start CMake building
+os.chdir(CPP_BUILD_DIRECTORY)
+subprocess.run(["ls", "-l"])
+subprocess.run(["cmake", ".."])
+subprocess.run(["make"])
```
antonio2368, line 63: It would be nice if you could fetch the number of cores and use that for the `-j` flag.
```
subprocess.run(["make", f"-j{core_count}"])
```

test_e2e
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env python3
+
+import os
+import subprocess
+
+WORK_DIRECTORY = os.getcwd()
+E2E_DIRECTORY = f"{WORK_DIRECTORY}/e2e"
+
+#################################################
+#           Unit Test Python modules            #
```
antonio2368, line 10: E2E test?

.github/workflows/test.yml
```diff
@@ -41,22 +42,29 @@ jobs:
...
     - name: Install Python dependencies
       run: |
         python -m pip install --upgrade pip
         pip install -r ./python/tests/requirements.txt
         pip install -r ./python/requirements.txt
     - name: C++ Modules unit tests
       run: |
-        for f in ./cpp/build/*/*_test; do $f; done
+        pushd cpp/build
+        ls
```
antonio2368, line 56: why `ls` and `popd`?
and why use `pushd` instead of `cd`?

test_unit
```diff
@@ -0,0 +1,28 @@
...
+#            Unit Test C++ modules              #
+#################################################
+CPP_BUILD_DIRECTORY = f"{CPP_DIRECTORY}/build"
+if not os.path.exists(CPP_BUILD_DIRECTORY):
+    print("No C++ build directory in ./cpp/build !")
+    exit(1)
+
+# Run ctest
+os.chdir(CPP_BUILD_DIRECTORY)
+subprocess.run(["ctest"])
```
antonio2368, line 20: ctest also has `-j` flag.

.github/workflows/test.yml
```diff
@@ -13,14 +16,12 @@ jobs:
       uses: actions/cache@v1
       with:
         path: ~/memgraph
-        key: cache-memgraph-v1.3.0-docker-image
+        key: cache-memgraph-v${{env.MG_VERSION}}-docker-image
```
antonio2368, line 19:
```suggestion
        key: cache-memgraph-v${{ env.MG_VERSION }}-docker-image
```
I think this is the convention for using variables in Github Actions. Don't forget to change it everywhere else.

test_unit
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env python3
```
antonio2368, line 1: Where is this file used?

antonio2368: I see it mentioned in README... Nice!

jmatak: I wanted to separate C++ and Python tests in the Github actions for the purpose of easier debugging, however included this script for anyone wanting easy check on all the modules in the repository. I have put the same information in the README file.

antonio2368 (approved):

PR #22, opened 2021-04-09
Find bridges algorithm

jmatak:


cpp/bridges_module/algorithm/bridges.hpp
```diff
@@ -0,0 +1,38 @@
...
+    counter = 0;
+  }
+
+  void Update(std::uint64_t node_id) {
+    counter++;
+    visited[node_id] = true;
+    discovery[node_id] = counter;
+    low_link[node_id] = counter;
+  }
+};
```
antonio2368, line 28: This can be put into utils now right?

cpp/bridges_module/bridges_test.cpp
```diff
@@ -0,0 +1,272 @@
...
+  Timer() : start_time_(std::chrono::steady_clock::now()) {}
+
+  template <typename TDuration = std::chrono::duration<double>>
+  TDuration Elapsed() const {
+    return std::chrono::duration_cast<TDuration>(std::chrono::steady_clock::now() - start_time_);
+  }
+
+ private:
+  std::chrono::steady_clock::time_point start_time_;
+};
```
antonio2368, line 22: This can be put into utils or maybe even create test_utils. :thinking:

cpp/bridges_module/algorithm/bridges.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+void BridgeDfs(uint64_t node_id, uint64_t parent_id, bridges_util::NodeState *state,
+               std::vector<mg_graph::Edge<>> *bridges, const mg_graph::GraphView<> &graph);
+}  // namespace bridges_util
```
antonio2368, line 32: What's the different in the DFS implementation between this one and the Bcc?

cpp/bridges_module/algorithm/bridges.hpp
jmatak: Not sure that this is useful for each algorithm, only ones that are copied from previous implementation repository

cpp/bridges_module/bridges_test.cpp
jmatak: This is a good idea, might be useful to have test utils for creating test graphs and having a timer.

cpp/bridges_module/algorithm/bridges.hpp
jmatak: There is some kind of difference in the internal logic, I might want to check on this one further because a lot of algorithms are going to use DFS. Might be useful if DFS only employs basic logic in mg_graph, but then uses the `consumer` pattern to do custom stuff that each different algorithm requires.

antonio2368: Yep, I was thinking about something like that. But, if the common part is mixed with custom parts you will end up with a mess, and repeating the DFS logic for each algorithm would be better.

jmatak: I think the common part largely intersects with the custom part, which would be difficult to implement as a utility.

cpp/bridges_module/bridges_test.cpp
```diff
@@ -0,0 +1,213 @@
...
+std::uint64_t CountComponents(const mg_graph::Graph<> &graph) {
+  std::uint64_t n_components = 0;
+  std::vector<bool> visited(graph.Nodes().size());
+
+  for (auto &node : graph.Nodes()) {
+    if (visited[node.id]) continue;
+    ++n_components;
+
+    std::queue<std::uint64_t> queue;
+    visited[node.id] = true;
```
jbajic, line 21: If the queue contains node ids, which are defined as TSize, then it should be of type whichever is Graph, something like this:
```suggestion
template <typename TSize = std::uint64_t>
std::uint64_t CountComponents(const mg_graph::Graph<TSize> &graph) {
  std::uint64_t n_components = 0;
  std::vector<bool> visited(graph.Nodes().size());

  for (auto &node : graph.Nodes()) {
    if (visited[node.id]) continue;
    ++n_components;

    std::queue<TSize> queue;
    visited[node.id] = true;
```

cpp/bridges_module/bridges_test.cpp
```diff
@@ -0,0 +1,213 @@
...
+/// Checks if obtained list of bridges is correct.
+bool CheckBridges(std::vector<mg_graph::Edge<>> user, std::vector<std::pair<std::uint64_t, std::uint64_t>> correct) {
+  std::set<std::pair<std::uint64_t, std::uint64_t>> user_bridge_set, correct_bridge_set;
+  for (auto &p : correct) {
+    correct_bridge_set.insert(p);
+    correct_bridge_set.insert({p.second, p.first});
+  }
+  for (auto &edge : user) {
+    user_bridge_set.insert({edge.from, edge.to});
+    user_bridge_set.insert({edge.to, edge.from});
```
jbajic, line 64:
```suggestion
    user_bridge_set.emplace_back(edge.from, edge.to);
    user_bridge_set.emplace_back(edge.to, edge.from);
```

cpp/mg_utility/mg_generate.hpp
```diff
@@ -0,0 +1,58 @@
...
+  return BuildGraph(num_nodes, {rand_edges.begin(), rand_edges.end()});
+}
+
+/// Generates a random undirected tree with a given number of nodes.
+/// The generated tree is not picked out of a uniform distribution.
+std::unique_ptr<mg_graph::Graph<>> GenRandomTree(std::uint64_t num_nodes) {
+  auto seed = std::chrono::high_resolution_clock::now().time_since_epoch().count();
+  std::mt19937 rng(seed);
+  std::vector<std::pair<std::uint64_t, std::uint64_t>> edges;
+  for (std::uint64_t i = 1; i < num_nodes; ++i) {
```
jbajic, line 51: For num_nodes it would be better to use `size_t`

cpp/mg_utility/mg_test_utils.hpp
```diff
@@ -0,0 +1,20 @@
...
+ public:
+  explicit Timer() : start_time_(std::chrono::steady_clock::now()) {}
+
+  template <typename TDuration = std::chrono::duration<double>>
+  TDuration Elapsed() const {
+    return std::chrono::duration_cast<TDuration>(std::chrono::steady_clock::now() - start_time_);
+  }
+
+ private:
+  std::chrono::steady_clock::time_point start_time_;
```
jbajic, line 17: I assume there is not use case to ever change the start_time, since its being set in constructor
```suggestion
  const std::chrono::steady_clock::time_point start_time_;
```

cpp/bridges_module/bridges_test.cpp
```diff
@@ -0,0 +1,213 @@
...
+}
+
+std::vector<std::pair<std::uint64_t, std::uint64_t>> GetBridgesBruteforce(mg_graph::Graph<> *graph) {
+  using IntPair = std::pair<std::uint64_t, std::uint64_t>;
+
+  std::uint64_t comp_cnt = CountComponents(*graph);
+  std::vector<IntPair> bridges;
+  auto edges = graph->ExistingEdges();
+  for (const auto &edge : edges) {
+    auto from = edge.from, to = edge.to;
```
jbajic, line 44:
```suggestion
  for (const auto &[id, from, to] : edges) {
```

cpp/bridges_module/bridges_test.cpp
antonio2368: That's a nice catch!
But these functions are restricted only to this TU, so you don't need the flexibility of the templates that much. Every created `Graph` here uses the default argument, and it's okay to have a function that only accepts that.
If he tried to call this function with any other TSize, it would be a compilation error, so it wouldn't go undetected.
This boils down to not overcomplicate things if not necessary, in C++ it's mostly trying not to use templates if you can. :D

Also, because Graph objects of different template arguments cannot be converted to each other, this showcases nicely how you don't need the default argument here.
```
template <typename TSize>
std::uint64_t CountComponents(const mg_graph::Graph<TSize> &graph)
```
`TSize` will be deduced from the graph type. 
If you call this function with a `Graph<>` object, the TSize will be the default type of the Graph template argument.

cpp/bridges_module/bridges_test.cpp
antonio2368: For primitive types, like here, it's better to prefer copying over creating a reference.

cpp/mg_utility/mg_test_utils.hpp
antonio2368: I think it's generally accepted that having const member variables are more trouble than they're worth.
First of all, it disables move, because const values cannot be moved (this is why you should never return a const value from a function).
Second, you cannot use copy assignment as it copies field by field, so you cannot copy the start_time from one timer into another.
You can create API around that, but it's better to have a simple nonconst private member and control the access using const getters/methods.

cpp/bridges_module/bridges_test.cpp
jmatak: Are you sure that std::set contains this function (_emplace_back_)? Also, we are working with pairs here

jbajic: You are right its `emplace`, sry :smile:

cpp/bridges_module/algorithm/bridges.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+    if (state->low_link[next_id] > state->discovery[node_id]) {
+      if (graph.GetEdgesBetweenNodes(edge.from, edge.to).size() == 1) bridges->push_back(edge);
+    }
+  }
+}
+
+std::vector<mg_graph::Edge<>> bridges_alg::GetBridges(const mg_graph::GraphView<> &graph) {
+  auto number_of_nodes = graph.Nodes().size();
+  bridges_util::NodeState state(number_of_nodes);
```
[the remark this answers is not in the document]
jmatak: Suggesting that _NodeState_ parameter should be of type _size_t_

cpp/mg_utility/mg_generate.hpp
jmatak: `size_t` is used for looping? But what if our container size does not fit a 32-bit integer? This is something for my education

cpp/bridges_module/bridges_test.cpp
jmatak: What is the example of copying here?

antonio2368: `for (const auto [id, from, to] : edges)`

cpp/bridges_module/algorithm/bridges.cpp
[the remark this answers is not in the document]
jbajic: Not `NodeState` but the parameter in `NodeState` constructor (`bridges.hpp`), maybe I should have commented there...

cpp/mg_utility/mg_generate.hpp
jbajic: `size_t` is used whenever you are referring to any container size (arrays, vectors, sets..), and they will assume the appropriate type which can best express the container size (by using size_t you do not have to worry if the container size will be expressed in uin64 or uint32 or anything else)

cpp/bridges_module/bridges_test.cpp
jbajic: The idea is to remove the reference, since it makes no sense in referencing primitive types, like this:
```suggestion
  for (const auto [id, from, to] : edges) {
```

cpp/mg_utility/mg_generate.hpp
jmatak: This is helpful, thanks!

cpp/bridges_module/algorithm/bridges.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+    bridges_util::BridgeDfs(next_id, node_id, state, bridges, graph);
+    state->low_link[node_id] = std::min(state->low_link[node_id], state->low_link[next_id]);
+
+    const auto &edge = graph.GetEdge(neigh.edge_id);
+    if (state->low_link[next_id] > state->discovery[node_id]) {
+      if (graph.GetEdgesBetweenNodes(edge.from, edge.to).size() == 1) bridges->push_back(edge);
+    }
+  }
+}
```
jbajic, line 40: Since all these functions are defined in `bridefs_util` namespace you can write it like this:
```suggestion
namespace bridges_util {
  NodeState::NodeState(std::size_t number_of_nodes) {
...
}
```
 in order to avoid verbose definitions like this`bridges_util::NodeState::Update`

cpp/bridges_module/bridges_test.cpp
```diff
@@ -0,0 +1,211 @@
...
+}
+
+TEST(Bridges, SimpleNeighborCycle) {
+  auto graph = mg_generate::BuildGraph(2, {{0, 1}, {1, 0}});
+  auto bridges = bridges_alg::GetBridges(*graph);
+  ASSERT_EQ(0, bridges.size());
+}
+
+TEST(Bridges, NeighborCycle) {
+  auto graph = mg_generate::BuildGraph(6, {{0, 1}, {1, 2}, {2, 3}, {3, 0}, {3, 4}, {3, 4}, {4, 5}});
```
jbajic, line 184: Why do we duplicate `{3,4}` edge?

jmatak: This is a wild catch. Although there is no special reason for this, rather that it is a mistake made by the original code author.

jmatak: Update: seems like this test was checking a neighboring cycle, meaning two vertices connected with two edges. This shouldn't result as bridge in the algorithm, meaning that code was correct.

jbajic (approved):

cpp/bridges_module/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+target_link_libraries(bridges PRIVATE mg_utility)
+target_include_directories(bridges PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
+
+# Module tests
+include(GoogleTest)
+set(bridges_test_src 
+    bridges_test.cpp
+    algorithm/bridges.cpp)
+
+add_executable(bridges_test  "${bridges_test_src}")
```
antonio2368, line 17:
```suggestion
add_executable(bridges_test "${bridges_test_src}")
```

cpp/bridges_module/bridges_module.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include <mg_exceptions.hpp>
+#include <mg_utils.hpp>
+
+#include "algorithm/bridges.hpp"
+
+namespace {
+
+// const char *fieldEdgeID = "edge_id";
+const char *fieldNodeFrom = "node_from";
+const char *fieldNodeTo = "node_to";
```
antonio2368, line 11: `kFieldNodeFrom` and `kFieldNodeTo`? Not sure what is used for constants in this code but it's good to have convention for that.

antonio2368 (approved):
Looks good, only minor comments.

PR #23, opened 2021-04-12
C++ Testing utility

jmatak:


cpp/bridges_module/bridges_test.cpp
```diff
@@ -66,54 +51,10 @@ std::vector<std::pair<std::uint64_t, std::uint64_t>> GetBridges(mg_graph::Graph<
...
-  for (std::uint64_t i = 1; i < nodes; ++i) {
-    std::uniform_int_distribution<std::uint64_t> dist(0, i - 1);
-    auto dad = dist(rng);
-    edges.emplace_back(dad, i);
-  }
-  return BuildGraph(nodes, edges);
-}
+}  // namespace bruteforce_alg
 
+namespace test_bridges {
```
antonio2368, line 57: Why have namespaces at all?
You can put every free function here (including the ones under `bruteforce_alg`) in an anonymous namespace so you don't expose their symbols.
```
namespace {
}
```

cpp/mg_utility/mg_generate.hpp
```diff
@@ -0,0 +1,58 @@
...
+
+#include "mg_graph.hpp"
+
+namespace mg_generate {
+
+/// Builds the graph from a given number of nodes and a list of edges.
+/// Nodes should be 0-indexed and each edge should be provided in both
+/// directions.
+std::unique_ptr<mg_graph::Graph<>> BuildGraph(std::uint64_t num_nodes,
+                                              std::vector<std::pair<std::uint64_t, std::uint64_t>> edges) {
```
antonio2368, line 17: Why are you sending vector by value and not const reference?

cpp/biconnected_components_module/biconnected_components_test.cpp
```diff
@@ -4,69 +4,13 @@
...
-  std::mt19937 rng(seed);
-  std::vector<std::pair<std::uint64_t, std::uint64_t>> edges;
-  for (std::uint64_t i = 1; i < nodes; ++i) {
-    std::uniform_int_distribution<std::uint64_t> dist(0, i - 1);
-    std::uint64_t dad = dist(rng);
-    edges.emplace_back(dad, i);
-  }
-  return BuildGraph(nodes, edges);
-}
+namespace test_bcc {
```
antonio2368, line 13: Same comment as for the bridges_test.cpp

cpp/bridges_module/bridges_test.cpp
jmatak: I had namespaces only because there were similarly named functions in other modules, thanks for the tip!

antonio2368 (approved):

PR #24, opened 2021-04-12
Update Memgraph base docker image

MasterMedo:
Since Memgraph is in DockerHub now, this should be updated.


jmatak (approved):
Finally on Docker Hub!

MasterMedo:
I have no merging rights :(

PR #27, opened 2021-04-22
Cycles detection algorithm

jmatak:


cpp/cycles_module/cycles_test.cpp
```diff
@@ -0,0 +1,285 @@
...
+    std::rotate(user_cycle.begin(), std::min_element(user_cycle.begin(), user_cycle.end()), user_cycle.end());
+    user_cycles.push_back(user_cycle);
+  }
+
+  if (user_cycles.size() != correct.size()) {
+    return false;
+  }
+
+  for (size_t i = 0; i < user_cycles.size(); ++i) {
+    user_cycles[i].push_back(*user_cycles[i].begin());
```
jbajic, line 33:
```suggestion
    asssert(user_cycles[i].size() > 0);
    user_cycles[i].push_back(user_cycles[i][0]);
```
or just 
```suggestion
    user_cycles[i].push_back(user_cycles[i][0]);
```
since these are tests...

cpp/cycles_module/algorithm/cycles.cpp
```diff
@@ -0,0 +1,229 @@
...
+#include <unordered_set>
+
+#include <mg_exceptions.hpp>
+#include <mg_graph.hpp>
+
+#include "cycles.hpp"
+
+namespace cycles_util {
+
+NodeState::NodeState(size_t number_of_nodes) {
```
jbajic, line 13: Just to be consistent, since we are using everything from std, we could also use std::size_t
```suggestion
NodeState::NodeState(std::size_t number_of_nodes) {
```

cpp/cycles_module/algorithm/cycles.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+  bool IsVisited(std::uint64_t node_id);
+
+  void SetParent(std::uint64_t parent_id, std::uint64_t node_id);
+
+  std::uint64_t GetParent(std::uint64_t node_id);
+
+  void SetDepth(std::uint64_t node_id, std::uint64_t node_depth);
+
+  std::uint64_t GetDepth(std::uint64_t node_id);
```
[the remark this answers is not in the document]
jmatak: In addition to this. Is it necessary to have an implementation in another file, once when implementation is trivial as is in this example?

cpp/cycles_module/algorithm/cycles.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+  return cycles;
+}
+
+std::vector<std::pair<mg_graph::Node<>, mg_graph::Node<>>> GetNeighbourCycles(const mg_graph::GraphView<> &graph) {
+  std::vector<std::pair<mg_graph::Node<>, mg_graph::Node<>>> cycles;
+  std::set<std::pair<std::uint64_t, std::uint64_t>> multi_edges;
+  for (const auto &edge : graph.Edges()) {
+    const auto [from, to] = std::minmax(edge.from, edge.to);
+    std::pair<std::uint64_t, std::uint64_t> e = {from, to};
```
jbajic, line 194:
```suggestion
    std::pair<std::uint64_t, std::uint64_t> e{from, to};
```

cpp/cycles_module/algorithm/cycles.cpp
```diff
@@ -0,0 +1,207 @@
...
+    for (auto next_node : adj_list[curr_node]) {
+      if (!visited[next_node]) curr_node = next_node;
+    }
+  }
+
+  for (auto node : nodes) {
+    if (!visited[node]) {
+      return;
+    }
+  }
```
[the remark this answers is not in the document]
jmatak: In line 126 a cycle is constructed, the rest of the code is there to check out whether the cycle is a valid one or not, therefore, I wouldn't separate it.

cpp/cycles_module/cycles_module.cpp
```diff
@@ -0,0 +1,60 @@
+#include <mg_procedure.h>
+#include <mg_utils.hpp>
+
+#include "algorithm/cycles.hpp"
+
+namespace {
+
+const char *kFieldCycleId = "cycle_id";
+const char *kFieldNode = "node";
```
jbajic, line 9:
```suggestion
constexpr char const *kFieldCycleId = "cycle_id";
constexpr char const *kFieldNode = "node";
```

cpp/cycles_module/algorithm/cycles.cpp
```diff
@@ -0,0 +1,201 @@
...
+  for (std::size_t i = 0; i < fundamental_cycles.size(); ++i) {
+    if ((mask & (1 << i)) == 0) continue;
+    for (std::size_t j = 1; j < fundamental_cycles[i].size(); ++j) {
+      auto edge = std::minmax(fundamental_cycles[i][j], fundamental_cycles[i][j - 1]);
+      edge_cnt[edge]++;
+    }
+  }
+
+  std::map<std::uint64_t, std::vector<std::uint64_t>> adj_list;
+  std::set<std::uint64_t> nodes;
```
antonio2368, line 102: Use `unordered_map` and `unordered_set` if you don't care about the order, and it seems you don't.

cpp/cycles_module/algorithm/cycles.cpp
```diff
@@ -0,0 +1,201 @@
...
+    nodes.insert(from);
+    nodes.insert(to);
+  }
+
+  // deg(v) = 2 for all vertices in a cycle.
+  for (const auto node : nodes) {
+    if (adj_list[node].size() != 2) return;
+  }
+
+  std::map<std::uint64_t, bool> visited;
```
antonio2368, line 118: You mean `std::unordered_set` where true means that the value is present in the set? :D

jmatak: Legacy implementation leads to legacy issues :smiley:

cpp/cycles_module/algorithm/cycles.hpp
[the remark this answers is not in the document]
antonio2368: Yeah, sometimes it's annoying to put a definition into a cpp but I would still do that with the exception of the methods whose implementation are really obvious in general. Example of this would be getter functions 
```
struct Foo {
 const auto &field() const { return field_; }
 private:
  int field_;
}; 
```
It's obvious that it's a getter function, they will always return a direct reference to the field and you can use `const auto &` for the return type.

For all of this, it's not that obvious what the implementation is. What if you want to add checks later on for the ids? Every time you add something new to the header forces recompilation of every file that includes that header. If you kept it into cpp you would only have to recompile that one file.

jbajic (approved):

cpp/cycles_module/algorithm/cycles.cpp
```diff
@@ -0,0 +1,201 @@
...
+  for (const auto node : nodes) {
+    if (adj_list[node].size() != 2) return;
+  }
+
+  std::unordered_set<std::uint64_t> visited;
+  std::vector<mg_graph::Node<>> cycle;
+  auto curr_node = *nodes.begin();
+
+  while (visited.find(curr_node) == visited.end()) {
+    cycle.push_back({curr_node});
```
antonio2368, line 123: `emplace_back`

cpp/cycles_module/algorithm/cycles.cpp
```diff
@@ -0,0 +1,201 @@
...
+    visited.insert(curr_node);
+    for (const auto next_node : adj_list[curr_node]) {
+      if (visited.find(next_node) == visited.end()) curr_node = next_node;
```
antonio2368, line 126: why not `break` here?

cpp/cycles_module/algorithm/cycles.cpp
```diff
@@ -0,0 +1,201 @@
...
+  }
+
+  // deg(v) = 2 for all vertices in a cycle.
+  for (const auto node : nodes) {
+    if (adj_list[node].size() != 2) return;
+  }
+
+  std::unordered_set<std::uint64_t> visited;
+  std::vector<mg_graph::Node<>> cycle;
+  auto curr_node = *nodes.begin();
```
antonio2368, line 120: Now it's a random element from a list of nodes, is this okay?

cpp/cycles_module/algorithm/cycles.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+  if (cycle.size() > 2) {
+    cycles->emplace_back(cycle);
+  }
+}
+
+void GetCyclesFromFundamentals(const std::vector<std::vector<uint64_t>> &fundamental_cycles,
+                               const mg_graph::GraphView<> &graph, std::vector<std::vector<mg_graph::Node<>>> *cycles) {
+  auto size = 1 << fundamental_cycles.size();
+  for (std::int32_t mask = 1; mask < size; ++mask) {
```
antonio2368, line 144: I would use `std::uint32_t` for mask stuff. It's more natural to use unsigned types when doing bit stuff.

cpp/cycles_module/cycles_test.cpp
```diff
@@ -0,0 +1,285 @@
...
+                                   {10, 1, 4, 5, 0, 11, 6, 7, 8, 9},
+                                   {10, 1, 0, 5, 6, 11},
+                                   {10, 1, 2, 3, 4, 5, 0, 11},
+                                   {10, 1, 2, 3, 4, 5, 0, 11, 6, 7, 8, 9},
+                                   {10, 1, 2, 3, 4, 5, 6, 11},
+                                   {10, 1, 4, 5, 6, 7, 8, 9}}));
+}
+
+TEST(Cycles, NeighbouringCycles) {
+  auto graph = new mg_graph::Graph<>();
```
antonio2368, line 257: Where is `delete` for this? Use `std::unique_ptr`.

cpp/cycles_module/algorithm/cycles.cpp
jmatak: Is this a question from the algorithm standpoint? In this implementation, there is no guarantee that elements should be ordered within the cycle, and there is no clear starting point.

antonio2368: Yeah, from the algorithm standpoint if it's okay.
Okay then. :+1:

cpp/cycles_module/algorithm/cycles.cpp
jmatak: If emplace_back takes various number of elements, I should explicitly construct Node like this: `cycle.emplace_back((mg_graph::Node<>){curr_node});`, or create a constructor for Node. Which one is usually a preferred way?

antonio2368: Oh it doesn't have a constructor, you would prefer the second way of creating a constructor. But this is a really simple object so no need for that. You can leave push_back.

cpp/cycles_module/algorithm/cycles.hpp
```diff
@@ -11,7 +11,7 @@ namespace cycles_util {
 struct NodeState {
   std::vector<bool> visited;
   std::vector<std::uint64_t> discovery, depth;
-  std::vector<std::uint64_t> parent;
+  std::vector<std::int64_t> parent;
```
jmatak, line 14: @antonio2368 What do you think of this. This should be a container that keeps the track of parent nodes within the zero-indexed nodes in the graph. The problem is the initialization of that container, the right thing to do is setting it to -1, but it can be accomplished only by using `int64_t` which is incompatible to the `uint64_t` in which nodes are stored.

antonio2368: `std::optional<uint64_t>` :smile:

cpp/cycles_module/algorithm/cycles.cpp
```diff
@@ -0,0 +1,208 @@
...
+void FindNonSpanningTreeEdges(std::uint64_t node_id, const mg_graph::GraphView<> &graph, NodeState *state,
+                              std::set<std::pair<uint64_t, uint64_t>> *non_st_edges) {
+  std::unordered_set<std::uint64_t> unique_neighbour;
+  state->SetVisited(node_id);
+  for (const auto &neigh : graph.Neighbours(node_id)) {
+    auto next_id = neigh.node_id;
+
+    // Check if is returning edge or already visited neighbour
+    if (state->GetParent(node_id) != std::nullopt &&
+        (next_id == state->GetParent(node_id) || unique_neighbour.find(next_id) != unique_neighbour.end())) {
```
antonio2368, line 37: `std::optional` has an implicit conversion to bool so you don't have to compare it to `!= std::nullopt` in ifs.
```suggestion
    if (const auto parent = state->GetParent(node_id); 
        parent && (next_id == parent || unique_neighbour.find(next_id) != unique_neighbour.end()) {
```

cpp/cycles_module/algorithm/cycles.cpp
```diff
@@ -0,0 +1,208 @@
...
+  // climb until a and b reach the same depth:
+  //
+  //       ()               ()
+  //      /  \             /  \.
+  //    ()   (b)   -->   (a)  (b)
+  //   /  \             /   \.
+  // (a)  ()           ()   ()
+  cycle.emplace_back(node_a);
+  while (state.depth[node_a] > state.depth[node_b]) {
+    if (state.GetParent(node_a) == std::nullopt) {
```
antonio2368, line 78:
```suggestion
    if (!state.GetParent(node_a)) {
```

cpp/cycles_module/algorithm/cycles.cpp
```diff
@@ -0,0 +1,208 @@
...
+      throw mg_exception::InvalidIDException();
+    }
+    node_a = state.GetParent(node_a).value();
+    cycle.emplace_back(node_a);
```
antonio2368, line 82: optional has really similar usage to the `std::unique_ptr`. You can use * operator to get underlying element or -> operator to call a method on the data saved inside the `std::optional` (if it exists of course).

Here's a typical usage for your use case.
```suggestion
    const auto maybe_parent = state.GetParent(node_a)
    if (!maybe_parent) {
      throw mg_exception::InvalidIDException();
    }
    node_a = *maybe_parent;
    cycle.emplace_back(node_a);
```

antonio2368 (approved):
:+1:

jbajic (approved):

PR #35, opened 2021-05-10
Directed GraphView implementation. mg_utility documentation

jmatak:
- [x] Directed GraphView
- [x] Better documentation for mg_utility


cpp/mg_utility/mg_generate.hpp
```diff
@@ -10,20 +10,32 @@
 
 namespace mg_generate {
 
-/// Builds the graph from a given number of nodes and a list of edges.
+///
+/// @brief Builds the undirected graph from a given number of nodes and a list of edges.
 /// Nodes should be 0-indexed and each edge should be provided in both
```
SuzanaPratljacic, line 15: Edges should be in both directions to create an undirected graph?

jmatak: To create an undirected graph, one should specify edges that will act like undirected. For example, when we say `BuldGraph(2, {{0,1}, {1,0}}` we are creating two undirected edges, basically the same case as `BuldGraph(2, {{0,1}, {1,0}}`.

cpp/connectivity_module/connectivity_module.cpp
```diff
@@ -63,31 +58,24 @@ static void Weak(const mgp_list *args, const mgp_graph *memgraph_graph,
     }
 
     for (const auto &p : vertex_component) {
-
       auto vertex_id = graph->GetMemgraphNodeId(p.first);
       auto component_id = p.second;
```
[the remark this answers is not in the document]
jmatak: Will add this change in other PR, since it is not explicitly related to this one (this module has been in  review a long time ago)

cpp/connectivity_module/connectivity_module.cpp
```diff
@@ -63,31 +58,24 @@ static void Weak(const mgp_list *args, const mgp_graph *memgraph_graph,
     }
 
     for (const auto &p : vertex_component) {
```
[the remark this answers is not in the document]
jmatak: I Will add this change in other PR since it is not explicitly related to this one (this module has been in review a long time ago)

cpp/mg_utility/mg_graph.hpp
```diff
@@ -158,6 +155,30 @@ class Graph : public GraphView<TSize> {
...
+
+    if (from < 0 || to < 0 || from >= nodes_.size() || to >= nodes_.size()) {
+      throw mg_exception::InvalidIDException();
+    }
+    auto id = edges_.size();
+    edges_.push_back({id, from, to});
+    adj_list_[from].push_back(id);
+    neighbours_[from].emplace_back(to, id);
+    nodes_to_edge_.insert({std::minmax(from, to), id});
+    return id;
```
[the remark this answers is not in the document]
jmatak: Fixed

jbajic (approved):

cpp/mg_utility/mg_graph.hpp
```diff
@@ -134,26 +132,39 @@ class Graph : public GraphView<TSize> {
...
   ///
   /// @param[in]  from  The from node identifier
   /// @param[in]  to    The to node identifier
   ///
   /// @return     Created edge id
-  TSize CreateEdge(std::uint64_t memgraph_id_from,
-                   std::uint64_t memgraph_id_to) {
-    auto from = memgraph_to_inner_id_[memgraph_id_from];
-    auto to = memgraph_to_inner_id_[memgraph_id_to];
+  TSize CreateUndirectedEdge(std::uint64_t memgraph_id_from, std::uint64_t memgraph_id_to) {
```
antonio2368, line 141: hmm, have a single function:
```
TSize CreateEdge(std::uint64_t from, std::uint64_t to, bool is_undirected)
```
Then you have more or less the logic from the CreateDirectedEdge function with the check at the end
```
if (is_undirected) {
  adj_list_[to]...
   ...
}
```
or if you want to be extra performance freaks (to eliminate the if), you can have compile time parameter
`template <bool is_undirected>`
and have 
```
if constexpr (is_undirected) {
  adj_list_[to]...
}
```
but then you can't eliminate the switch statements above.

jmatak: Separate functions were made on purpose. I wanted to emphasize that there exist 2 different functions for edge creation and raise awareness for the final user. 

However, since there might be more use-cases where a switch would be needed, I'll refactor this.

jmatak: @antonio2368 
For the purpose of readability and easier extension, do you think it is better to model graph type with `bool is_undirected`, or to use custom enumeration `mg_graph::GraphType graph_type`

antonio2368: yeah, enum is much better as it's more explicit + you should be able to use the enum as template argument if you want.
I just don't like repeating (e.g. calling `GetInnerNodeId` twice), you can avoid that with `if`/`if constexpr`.

jmatak: I updated following your comment. I chose a solution without templatizing the CreateEdge. I wanted to eliminate all unnecessary confusion for the end-user.

cpp/mg_utility/mg_graph.hpp
```diff
@@ -13,20 +13,20 @@
 
 namespace mg_graph {
 
+enum class GraphType : std::uint8_t { kDirectedGraph, kUndirectedGraph };
```
antonio2368, line 16: Do you need k before the names if it's part of the enum?
k is used for constants, enum values are something else.

jmatak: https://google.github.io/styleguide/cppguide.html#Enumerator_Names 

Followed this guide.

antonio2368: Interesting, nice catch!

antonio2368 (approved):

PR #39, opened 2021-06-08
Publish MAGE Docker Hub image

jmatak:
I have tested this workflow with this branch [T034-MAGE_mage-docker-hub]

All went well! 
**Image**: https://hub.docker.com/r/memgraph/memgraph-mage
**Action**: https://github.com/memgraph/mage/actions/runs/918565808

I can easily change the name. This is just a workflow example.


.github/workflows/docker_publish.yml
```diff
@@ -0,0 +1,22 @@
...
+      - name: Log in to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+      - name: Push to Docker Hub
+        uses: docker/build-push-action@v2
+        with:
+          push: true
+          tags: memgraph/memgraph-mage:latest
```
jbajic, line 22: How are we going to version memgraph-mage? We should at least have every `memgraph-mage for every memgraph version on dockerhub, like `memgraph/memgraph-mage:1.4.0` & `memgraph/memgraph-mage:1.5.0`. What do you say, we would specify memgraph versions for mage releases and build and upload every version for which we are doing a release for mage (e.g last 3 versions always).

jbajic: Also it should have two tags, latest and memgraph version, maybe something like this:
```suggestion
          tags: 
            - memgraph/memgraph-mage:latest
            - memgraph/memgraph-mage: ${{ env.MG_VERSION }}
```

jmatak: I think that version should depend on the version of Memgraph image it was built from. That means changing the tagging in Dockerfile each time a new version of Memgraph comes.

jmatak: Right now we build it from `memgraph/memgraph:latest`

jmatak: Should we add this topic to the issue and push the workflow as it is right now?

jbajic (approved):

PR #42, opened 2021-06-09
Add VRP (vehicle routing problem) module

Josipmrden:


Josipmrden:
I forgot to add tests tho lol, coming in next commit.

python/mage/constraint_programming/vrp_cp_solver.py
```diff
@@ -0,0 +1,208 @@
...
+        )
+
+    def _add_objective(self):
+        intermedias = []
+        for edge, variable in self._edge_chosen_vars.items():
+            duration = self.get_distance(edge)
+            mul = self._model.Intermediate(duration * variable)
+            intermedias.append(mul)
+
+        self._model.Obj(sum(intermedias))
```
jbajic, line 197:
```suggestion
        intermedias_sum = 0
        for edge, variable in self._edge_chosen_vars.items():
            duration = self.get_distance(edge)
            intermedias_sum += self._model.Intermediate(duration * variable)

        self._model.Obj(intermedias_sum)
```

python/mage/constraint_programming/vrp_cp_solver.py
```diff
@@ -0,0 +1,208 @@
...
+
+    def _add_options(self):
+        self._model.options.SOLVER = 1
+
+    def print_time_vars(self):
+        for time, variable in self._time_vars.items():
+            print(f"{time}={variable.value}")
+
+    def print__edge_chosen_vars(self):
```
jbajic, line 206:
```suggestion
    def print_edge_chosen_vars(self):
```

python/vrp.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    global __depot_index
+
+    if __depot_index is not None:
+        return __depot_index
+
+    for i, vertex in enumerate(vertices):
+        if depot_label in vertex.labels:
+            __depot_index = vertices[i].id
+            break
```
jbajic, line 44:
```suggestion
    for vertex in vertices:
        if depot_label in vertex.labels:
            __depot_index = vertex.id
            break
```

python/vrp.py
```diff
@@ -0,0 +1,82 @@
...
+
+    if __depot_index == None:
+        raise Exception("No depot specified!")
```
jbajic, line 47: Define our exception `DepotUnspecifiedException`

jbajic (changes requested):

python/vrp.py
Josipmrden: Thanks, I actually meant to write `__depot_index = i` but since I deleted everything from the database, id got equal to index

python/mage/constraint_programming/vrp_cp_solver.py
```diff
@@ -0,0 +1,364 @@
...
+
+    def __init__(self, no_vehicles: int, distance_matrix: np.array, depot_index: int):
+        self._model = GEKKO(remote=False)
+
+        self.no_vehicles = no_vehicles
+        self.distance_matrix = distance_matrix
+        self.depot_index = depot_index
+
+        self._edge_chosen_vars = dict()
+        self._time_vars = dict()
```
jmatak, line 24: Where is `_time_vars` set? Or it is used only as a container in constraint?

python/mage/constraint_programming/vrp_cp_solver.py
```diff
@@ -0,0 +1,364 @@
...
+        ):
+            return 0
+
+        return self.distance_matrix[node_from][node_to]
+
+    def print_results(self):
+        # Results
+        print("\nResults")
+        print(f"Obj={self._model.options.objfcnval}")
+        self.print_time_vars()
```
jmatak, line 99: Do you need printing result?

Josipmrden: Yeah that's a valid point.

python/mage/constraint_programming/vrp_cp_solver.py
Josipmrden:
```
    def _initialize_location_node(self, node_index: int):
        self._time_vars[node_index] = self._model.Var(value=0, lb=0, integer=False)
```

python/vrp.py
```diff
@@ -0,0 +1,94 @@
+from mage.geography import create_distance_matrix, LATITUDE, LONGITUDE
+from mage.constraint_programming import VRPConstraintProgrammingSolver
+from typing import Dict, List
+
+import mgp
+
+
+__distance_matrix = None
```
jmatak, line 8: It might be useful to stress test this. Keeping the distance matrix in memory for the whole time of execution might be pretty expensive.

python/vrp.py
```diff
@@ -0,0 +1,94 @@
...
+    global __distance_matrix
+
+    if __distance_matrix is not None:
+        return __distance_matrix
+
+    vertex_positions: List[Dict[str, float]] = []
+    for vertex in vertices:
+        vertex_positions.append(
+            {
+                LATITUDE: vertex.properties.get(LATITUDE),
```
jmatak, line 25: What happens if there is no property named `latitude`?

python/vrp.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+    global __depot_index
+
+    if __depot_index is not None:
+        return __depot_index
+
+    for i, vertex in enumerate(vertices):
+        if depot_label in vertex.labels:
+            __depot_index = i
+            break
```
jmatak, line 47: Is there a better way of finding vertices with depot_labels, rather than O(n)

jmatak: Maybe the argument to procedure should be list of depot vertices. But if your design choice is to keep the label, it's also fine :+1:

jmatak (commented):
Left some mostly semantic comments.

python/vrp.py
Josipmrden: I don't think this can be better than O(n) since I never know which vertex is going to be the depot vertex and I need to find it inside the distance matrix.

I like the second approach though, it is then not possible not to have a depot node, and not rely on labels.

python/vrp.py
Josipmrden: I will handle the exception in next commit.

python/vrp.py
Josipmrden: We can see with buda how to approach this. Or have a maximum matrix size after which it deletes itself so it does not stay in memory the whole time.

jbajic (approved):
Small feedback, all good, gj!

python/mage/constraint_programming/vrp_cp_solver.py
```diff
@@ -0,0 +1,353 @@
...
+    def _add_objective(self):
+        intermedias_sum = 0
+        for edge, variable in self._edge_chosen_vars.items():
+            duration = self.get_distance(edge)
+            intermedias_sum += self._model.Intermediate(duration * variable)
+
+        self._model.Obj(intermedias_sum)
+
+    def _add_options(self):
+        self._model.options.SOLVER = 1
```
jmatak, line 182: Add documentation for this solver, it is uncertain what you mean by `SOLVER = 1`

python/mage/geography/distance_calculator.py
```diff
@@ -55,3 +58,11 @@ def calculate_distance_between_points(
         distance *= KM_MULTIPLIER
 
     return distance
+
+
+class InvalidCoordinatesException(Exception):
```
jmatak, line 63: Could be useful to give the user the explanation of why coordinates are wrong invalid

jmatak (approved):
Fix these minor details and I think we are ready to go

jmatak:
One more detail to add: 

- [ ] End to end tests
- [ ] README update with all the algorithms

python/mage/geography/distance_calculator.py
Josipmrden: I did leave a message in the code above?

python/mage/constraint_programming/vrp_cp_solver.py
```diff
@@ -0,0 +1,356 @@
...
+        return edges_vars
+
+    def _add_adjacent_input_edge_variables(
+        self, node_index: int
+    ) -> List[Tuple[int, int]]:
+        edges_vars = []
+
+        for adjacent_node in range(len(self.distance_matrix)):
+            if adjacent_node == self.depot_index:
+                continue
```
jmatak, line 149: There are two possible types of labels in this VRP scenario? Everything that is not considered a depot label is considered a location?

jmatak (approved):

PR #47, opened 2021-06-17
Mofified set cover result for better usability

Josipmrden:
2 Changes:
* Set remote execution of GEKKO to false (now executed locally) - still need maybe to add this as configurable as probably one needs a graphics card for constraint solving
* modified result of set cover for better usability and compliance with docs


jmatak (approved):
:fire:

PR #49, opened 2021-07-22
Add online node2vec

SuzanaPratljacic:
Online node2vec algorithm.


python/mage/node2vec_online_module/w2v_learners.py
```diff
@@ -0,0 +1,83 @@
...
+        self.num_epochs = 1
+        self.embeddings = None
+
+    def __str__(self):
+        return "gensimw2v_dim%i_lr%0.4f_neg%i_sg%i" % (
+            self.embedding_dims,
+            self.lr_rate,
+            self.neg_rate,
+            self.sg,
+        )
```
jmatak, line 41: Is string representation necessary for implementation?

python/mage/node2vec_online_module/walk_sampling.py
```diff
@@ -0,0 +1,135 @@
...
+        walk.append(node_)
+        while True:
+            if (
+                random.uniform(0, 1) < 1 / (cent_ * self.beta + 1)
+                or (node_ not in self.G)
+                or len(walk) >= self.max_len
+            ):
+                break
+            sum_ = cent_ * random.uniform(0, 1)
+            sum__ = 0
```
jmatak, line 78: Variable naming is not intuitive

python/node2vec_online.py
```diff
@@ -0,0 +1,166 @@
...
+            k,
+            full_walks,
+            embedding_dims,
+            learning_rate,
+            sg,
+            negative_rate,
+            threads,
+        )
+
+    embeddingd = context.learner.get_embedding_vectors()
```
jmatak, line 122: Naming

python/requirements.txt
```diff
@@ -1,2 +1,4 @@
 gekko==0.2.8
-networkx==2.5
\ No newline at end of file
+networkx==2.5
+gensim==4.0.0
+python-Levenshtein
```
jmatak, line 4: This one is failing to install in Github Action

python/node2vec_online.py
```diff
@@ -0,0 +1,166 @@
...
+    beta: float = 0.9,
+    cuttoff: int = 604800,
+    k: int = 4,
+    full_walks: bool = False,
+    embedding_dims: int = 128,
+    learning_rate: float = 0.01,
+    sg: bool = True,
+    negative_rate: float = 10,
+    threads: int = MAX_THREADS,
+):
```
jmatak, line 144: The update can occur only when an edge comes into the system?

jmatak (commented):
First iteration of comments

python/node2vec_online.py
```diff
@@ -0,0 +1,166 @@
...
+    @learner.setter
+    def learner(self, learner):
+        self._learner = learner
+
+    @start_time.setter
+    def start_time(self, start_time):
+        self._start_time = start_time
+
+    def is_initialized(self):
+        return self._learner is None or self._updater is None or self._start_time is None
```
jmatak, line 43: `is_initialized` is `true` once some of attributes is None, is this expected behaviour?

Dockerfile
```diff
@@ -4,7 +4,7 @@ FROM memgraph-mage
 USER root
 
 RUN apt-get update && \
-    apt-get --yes install curl git cmake g++ clang python3-dev && \
+    apt-get --yes install curl git cmake g++ clang python3-dev python3-setuptools && \
```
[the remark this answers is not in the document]
SuzanaPratljacic: python-Levenshtein

python/mage/node2vec_online_module/w2v_learners.py
```diff
@@ -34,44 +34,35 @@ def __init__(
...
         self.model.build_vocab(sentences, update=True)
         self.model.train(
-            sentences, epochs=self.num_epochs, total_examples=self.model.corpus_count)
-        self.embeddings = self.get_embedding_vectors()
+            sentences, epochs=self.num_epochs, total_examples=self.model.corpus_count
+        )
 
     def get_embedding_vectors(self):
+        if self.model is None:
+            return {}
```
jmatak, line 65: In which cases the model is None?

python/node2vec_online.py
```diff
@@ -72,38 +74,42 @@ def set(
...
     if not context.is_initialized():
         return
 
     embeddings = context.learner.get_embedding_vectors()
 
     return [
-        mgp.Record(node=node, embedding=embeddings)
-        for node, embedding in embeddings.items()
+        mgp.Record(
+            node_id=ctx.graph.get_vertex_by_id(str(node_id)), embedding=embeddings
```
jmatak, line 110: Is the name of record's value `node_id`, but the returned value is node?

python/mage/node2vec_online_module/w2v_learners.py
SuzanaPratljacic: When there is no data (because data is needed for model construction; look at the partial fit function)

jmatak (commented):
Further activity involves testing the module :D

python/node2vec_online.py
```diff
@@ -0,0 +1,248 @@
...
+    @start_time.setter
+    def start_time(self, start_time):
+        self._start_time = start_time
+
+    def is_initialized(self):
+        return (
+                self._learner is not None
+                and self._updater is not None
+                and self._start_time is not None
+        )
```
jmatak, line 47: Is mode uninitialized if any of these are None?

jmatak (approved):
Seems like in this iteration the setting of parameters is a little bit different than it was before, but I think I like it more as it is now. Although it is more difficult to set parameters because there are two methods, the user is at least aware of what he is doing and what needs to be set in order to get the algorithm running.

python/node2vec_online.py
antoniofilipovic: No, you are right, just updater and learner.

python/node2vec_online.py
```diff
@@ -0,0 +1,249 @@
...
+    """
+        This function resets parameters and embeddings already learned.
+    """
+
+    global node2vec_context
+
+    node2vec_context.start_time = None
+    node2vec_context.updater = None
+    node2vec_context.learner = None
+    return mgp.Record(message="Updater and learner are ready to be set again.")
```
jmatak, line 163: Name `reset` in this case is not doing what is usually intended to, maybe we should come up with the new one

antoniofilipovic: maybe `set_clean_state()` I don't have better ideas

jmatak: Let there be a reset, for now, I neither have a better idea

jmatak:
I think this is ready for shipping as soon as we agree on the format of E2E tests. Great work :+1:

jmatak (changes requested):
Thing that is left to do is merging with main and including `setup` and `cleanup` on E2E testing :smile:

jmatak (approved):

PR #51, opened 2021-07-26
Node similarity

Arya513:
Implementation of node similarity with 3 different algorithms:
- Jaccard similarity
- Overlap similarity
- Cosine similarity

- [x] Implementation
- [ ] Unit tests
- [x] End to end tests
- [x] Docs documentation


python/node_similarity.py
```diff
@@ -0,0 +1,289 @@
...
+# not implemented for directed graphs
+# not implemented for multigraphs
+# not taking into account edge weights
+
+@mgp.read_proc
+def jaccard(
+        context: mgp.ProcCtx,
+        node1: mgp.Any,
+        node2: mgp.Any,
+        mode: str = 'c'
```
jmatak, line 17: Although this is way shorter, maybe we should allow both 'c' and 'cartesian' as arguments?

python/node_similarity.py
```diff
@@ -0,0 +1,289 @@
...
+    if isinstance(node2, mgp.Vertex):
+        nodes2 = tuple([node2])
+    elif isinstance(node2, tuple):
+        nodes2 = node2
+    else:
+        raise TypeError("Invalid type of second argument.")
+
+    result = _calculate_similarity(nodes1, nodes2, _calculate_cosine, mode)
+
+    return [mgp.Record(node1=n1, node2=n2, similarity=similarity) for n1, n2, similarity in result]
```
jmatak, line 144: Since this is pretty repetitive, maybe it would be better to export it in a single function, where the only thing that differs is method of choice for calculation

python/node_similarity.py
```diff
@@ -0,0 +1,289 @@
...
+
+    denominator = sqrt(len(neighbours1) * len(neighbours2))
+
+    if denominator != 0:
+        cosine_similarity = len(neighbours1 & neighbours2) / denominator
+
+    return cosine_similarity
+
+
+def _calculate_similarity(nodes1: tuple, nodes2: tuple, method: Callable, mode: str) -> List:
```
jmatak, line 227: Specify return value to `List[float]`

python/node_similarity.py
```diff
@@ -0,0 +1,289 @@
...
+    :param str mode: Can be `p` for pairwise similarity calculation or `c` for calculating the similarity of the
+    Cartesian product of given tuples
+    """
+
+    result: list
+
+    if not isinstance(nodes1, tuple) or not isinstance(nodes2, tuple):
+        raise TypeError("Arguments should be tuples.")
+
+    if mode not in ['p', 'c']:
```
jmatak, line 250: Define constants on higher level rather than using 'p' or 'c' in the code

python/node_similarity.py
```diff
@@ -0,0 +1,289 @@
...
+        :rtype: Set
+
+    The input arguments consist of
+
+        :param mgp.Vertex node: A node
+    """
+
+    neighbors = set()
+
+    if isinstance(node, mgp.Vertex):
```
jmatak, line 280: Since you have type-checked the methods, you can remove this line here

jmatak (commented):
What's left are unit-tests and end-to-end tests. Since you have been working with mgp module inside similarity functions, unit-testing would be impossible (or try to mock it - https://changhsinlee.com/pytest-mock/), try focusing on writing more end-to-end tests

e2e/node_similarity_test/test_node_list_cosine_p/test.yml
```diff
@@ -0,0 +1,13 @@
...
+        WITH collect(n) as nodes, m
+        
+        CALL node_similarity.cosine(m, nodes) YIELD *
+          
+        RETURN node1.id as node1, node2.id as node2, similarity
+        
+
+exception: >
+        Incompatible lengths of given arguments
+
```
jmatak, line 13: Name of this test is `test_list_cosine_p`. On the other side, there is a testing of exception?

e2e/node_similarity_test/test_node_list_jaccard_p/test.yml
```diff
@@ -0,0 +1,13 @@
...
+        WITH collect(n) as nodes, m
+        
+        CALL node_similarity.cosine(m, nodes) YIELD *
+          
+        RETURN node1.id as node1, node2.id as node2, similarity
+        
+
+exception: >
+        Incompatible lengths of given arguments
+
```
jmatak, line 13: If this query is testing exception, write it in the name of test

e2e/node_similarity_test/test_node_list_overlap_p/test.yml
```diff
@@ -0,0 +1,12 @@
...
+        WHERE n.id < 2
+        WITH collect(n) as nodes, m
+        
+        CALL node_similarity.overlap(m, nodes) YIELD *
+          
+        RETURN node1.id as node1, node2.id as node2, similarity
+
+exception: >
+        Incompatible lengths of given arguments
+
```
jmatak, line 12: This one too

python/node_similarity.py
```diff
@@ -0,0 +1,302 @@
...
+        raise TypeError("Invalid type of first argument.")
+
+    if isinstance(node2, mgp.Vertex):
+        nodes2 = tuple([node2])
+    elif isinstance(node2, tuple):
+        nodes2 = node2
+    else:
+        raise TypeError("Invalid type of second argument.")
+
+    mode_constants = Mode()
```
jmatak, line 261: No need to create `Mode()` object. Simply call it by class: `Mode.PAIRWISE`

python/node_similarity.py
```diff
@@ -0,0 +1,302 @@
...
+
+    neighbors = set()
+
+    for edge in node.in_edges:
+        neighbors.add(edge.from_vertex)
+
+    for edge in node.out_edges:
+        neighbors.add(edge.to_vertex)
+
+    neighbors_dict[node.id] = neighbors
```
jmatak, line 300: Add a comment that this is used to preserve state. Also, this needs to be updated! Be careful, when calling the module again, after creating more nodes, this code would return the previous state, because as any other global static variable, this one is living in the global scope once is created.

jmatak (changes requested):

e2e/node_similarity_test/test_node_list_cosine_p/test.yml
Arya513: The name of this `test is test_node_list_cosine_p`. One argument is a node, the other is a list (which has a length greater than 1). For the pairwise mode, both need to be the same length. Since they're not, an exception occurs.

python/node_similarity.py
```diff
@@ -0,0 +1,285 @@
...
+
+        for edge in node.in_edges:
+            neighbors.add(edge.from_vertex)
+
+        for edge in node.out_edges:
+            neighbors.add(edge.to_vertex)
+
+        neighbors_dict[node.id] = neighbors
+
+        return neighbors
```
antepusic, line 285: If changes have been made to the graph since the previous query, this function wouldn’t reflect them because there’s no option to specify that the nodes’ neighbors are to be re-fetched.

Arya513: The dictionary `neighbors_dict` is reset at the beginning of each execution and all the neighbors are calculated anew.

python/node_similarity.py
```diff
@@ -0,0 +1,285 @@
...
+from typing import Set, List, Callable, Tuple, Union
+from itertools import product
+import mgp
+
+# not implemented for directed graphs
+# not implemented for multigraphs
+# not taking into account edge weights
+
+neighbors_dict = dict()
+neighbors_dict.clear()
```
jmatak, line 11: This is not clearing the cache, because this won't be be executed when module is started. This should be added to the `_calculate_similarity` since it is called often.

jmatak (changes requested):

python/node_similarity.py
Arya513: Okay, done.

python/node_similarity.py
```diff
@@ -0,0 +1,287 @@
...
+            return [
+                mgp.Record(node1=n1, node2=n2, similarity=method(n1, n2))
+                for n1, n2 in zip(nodes1, nodes2)
+            ]
+        else:
+            raise ValueError("Incompatible lengths of given arguments.")
+    elif mode.lower() in Mode.CARTESIAN:
+        return [
+            mgp.Record(node1=n1, node2=n2, similarity=method(n1, n2))
+            for n1, n2 in product(nodes1, nodes2)
```
Josipmrden, line 256: Have we checked how does it perform on a large number of nodes? Does it return a list of tuples, is the memory overhead too big? Maybe stress test this for 1000 or 10 000 nodes vs. a common for loop, for time benchmark

Josipmrden (commented):
Small comment about an external method we use, otherwise I think it's good

jmatak (commented):
One more thing is left, updating the README :+1:

python/node_similarity.py
Arya513: There's no significant difference in nested for loops vs. itertools.product time benchmark.

jmatak (approved):
:rocket: 
That's it

python/node_similarity.py
```diff
@@ -0,0 +1,285 @@
...
+
+        for edge in node.in_edges:
+            neighbors.add(edge.from_vertex.id)
+
+        for edge in node.out_edges:
+            neighbors.add(edge.to_vertex.id)
+
+        neighbors_dict[node.id] = neighbors
+
+        return neighbors
```
antoniofilipovic, line 285: There is no need for else branch here. Thing is that you already made a return statement and now you have a line of indent more.

python/node_similarity.py
```diff
@@ -0,0 +1,285 @@
...
+            ]
+        else:
+            raise ValueError("Incompatible lengths of given arguments.")
+    elif mode.lower() in Mode.CARTESIAN:
+        return [
+            mgp.Record(node1=n1, node2=n2, similarity=method(n1, n2))
+            for n1, n2 in product(nodes1, nodes2)
+        ]
+    else:
+        raise ValueError("Invalid mode.")
```
antoniofilipovic, line 257: This should go up in my opinion.
if mode.lower() not in [MODE.CARTESIAN, MODE.PAIRWISE):
     raise ValueError("You have selected invalide mode") #litle bit more friendly message
if mode.lower() in MODE.PAIRWISER:
   ... logic here
    return statement
logic for mode.CARTESIAN

python/node_similarity.py
```diff
@@ -0,0 +1,285 @@
...
+
+    nodes1: tuple
+    nodes2: tuple
+
+    if isinstance(node1, mgp.Vertex):
+        nodes1 = tuple([node1])
+    elif isinstance(node1, tuple):
+        nodes1 = node1
+    else:
+        raise TypeError("Invalid type of first argument.")
```
antoniofilipovic, line 234: Like my comment from below, I think this can go up (needs to be checked first)

python/node_similarity.py
```diff
@@ -0,0 +1,285 @@
...
+
+    jaccard_similarity: float
+
+    neighbors1: set
+    neighbors2: set
+
+    if node1 == node2:
+        return 1.0
+    else:
+        jaccard_similarity = 0.0
```
antoniofilipovic, line 114: No need for else statement here

python/node_similarity.py
```diff
@@ -0,0 +1,285 @@
...
+
+    return cosine_similarity
+
+
+def _calculate_similarity(
+    node1: Union[mgp.Vertex, Tuple[mgp.Vertex]],
+    node2: Union[mgp.Vertex, Tuple[mgp.Vertex]],
+    method: Callable,
+    mode: str,
+) -> List[Tuple[mgp.Vertex, mgp.Vertex, float]]:
```
antoniofilipovic, line 201: I am not an expert but I think 4 values as parameters are a little bit too much, think if you can somehow reduce it.

antoniofilipovic: Maybe mode can be something like default value.

python/node_similarity.py
```diff
@@ -0,0 +1,285 @@
...
+    Constants for the mode parameter
+    """
+
+    PAIRWISE = ["p", "pairwise"]
+    CARTESIAN = ["c", "cartesian"]
+
+
+@mgp.read_proc
+def jaccard(
+    context: mgp.ProcCtx, node1: mgp.Any, node2: mgp.Any, mode: str = "cartesian"
```
antoniofilipovic, line 24: Does it make sense that default is pairwise as it reduces complexity for calculation

python/node_similarity.py
```diff
@@ -0,0 +1,285 @@
...
+    for calculating the similarity of the Cartesian product of given tuples.
+    The default value is set to be a Cartesian product.
+    :type mode: str, optional
+
+    :return: This procedure returns a mgp.Record with 3 fields:
+    the first node, the second node and the overlap similarity between them
+    :rtype: mgp.Record
+    """
+
+    return _calculate_similarity(node1, node2, _calculate_overlap, mode)
```
antoniofilipovic, line 67: Here you can maybe even extract mode, so it is not `p` or `pairwise`, but that it is just `pairwise`. Same for `c` or `cartesian`. That way you have a clean state, and you can save it maybe as attribute, like current_mode

antoniofilipovic (changes requested):
I left few comments for structural changes, please check them and let me know if there are some discussion points 😄

python/node_similarity.py
Arya513: I prefer the way I wrote it.

python/node_similarity.py
Arya513: And again, I prefer the way I wrote it. I think it's easier to read.

python/node_similarity.py
Arya513: I don't see a problem with 4 arguments.

python/node_similarity.py
Arya513: I don't think it matters. In the end, the user will decide how they want to use the function.

python/node_similarity.py
Arya513: I don't understand what the problem here is.

python/node_similarity.py
Arya513: Okay.

python/node_similarity.py
Arya513: Okay.

python/node_similarity.py
```diff
@@ -0,0 +1,285 @@
...
+            ]
+        else:
+            raise ValueError("Incompatible lengths of given arguments.")
+    elif mode == Mode.CARTESIAN:
+        return [
+            mgp.Record(node1=n1, node2=n2, similarity=method(n1, n2))
+            for n1, n2 in product(nodes1, nodes2)
+        ]
+    else:
+        raise ValueError("Invalid mode.")
```
antoniofilipovic, line 257: I think I left somewhere else a comment, but more user-friendly message can go here
For example This is the invalid mode, but you can use one of the following: pairwise, cartesian

Arya513: Didn't see the comment. Changed the error message.

antepusic (commented):
Well done! 
The only suggestion I would make is to give the user the choice whether to reload `neighbors_dict`, e.g. in case that some edges have been deleted, but only if there’s a clear use case for it.

antepusic (approved):
👍

python/node_similarity.py
```diff
@@ -0,0 +1,285 @@
...
+    for calculating the similarity of the Cartesian product of given tuples.
+    The default value is set to be a Cartesian product.
+    :type mode: str, optional
+
+    :return: This procedure returns a mgp.Record with 3 fields:
+    the first node, the second node and the Jaccard similarity between them
+    :rtype: mgp.Record
+    """
+
+    return _calculate_similarity(node1, node2, _calculate_jaccard, Mode(mode))
```
jmatak, line 44: What happens if mode is something that is not in ["pairwise", "cartesian"]

Arya513: The enum throws an error. I fixed that so that the called function throws an error.

python/node_similarity.py
```diff
@@ -0,0 +1,304 @@
...
+            ]
+        else:
+            raise ValueError("Incompatible lengths of given arguments.")
+    elif mode == Mode.CARTESIAN:
+        return [
+            mgp.Record(node1=n1, node2=n2, similarity=method(n1, n2))
+            for n1, n2 in product(nodes1, nodes2)
+        ]
+    else:
+        raise ValueError("Invalid mode. Valid modes are: pairwise, cartesian.")
```
antoniofilipovic, line 276: This can be updated with according to this comment:
https://memgraph.slack.com/archives/C01L7JJL1L5/p1628255422010800

antoniofilipovic (approved):
Looks all good - ship it 😄

PR #53, opened 2021-07-26
Add tested stand-alone implementation of union-find

antepusic:
I added a stand-alone implementation of the union-find data structure and associated algorithms, which is to be used in a future union-find query module.

- [x] Implementation
- [x] Unit tests
- [x] End to end tests
- [x] Docs documentation


python/union_find.py
```diff
@@ -0,0 +1,126 @@
...
+        """
+        return self.nodes[node_id].parent
+
+    def grandparent(self, node_id: int) -> int:
+        """
+        Returns given node's grandparent's ID.
+
+        :param node_id: Node ID
+        """
+        return self.parent(self.parent(node_id))
```
jmatak, line 55: What happens when there is no grandparent for a certain node?

python/union_find.py
```diff
@@ -0,0 +1,126 @@
...
+        """
+        Returns the representative node's ID for the component that given node is member of.
+        Uses path splitting (https://en.wikipedia.org/wiki/Disjoint-set_data_structure#Finding_set_representatives)
+        in order to keep trees representing connected components flat.
+
+        :param node_id: Node ID
+        """
+        while node_id != self.parent(node_id):
+            self.nodes[node_id].parent = self.grandparent(node_id)  # path splitting
+            node_id = self.parent(node_id)
```
jmatak, line 75: Having hard time to understand this block. We are iterating as long as `node_id` is different than `node_id`'s parent, and at the line `#75` we are setting it to that exact value?

python/tests/union_find/test_find.py
```diff
@@ -0,0 +1,22 @@
+from python.union_find import UnionFind
+import pytest
+
+IDs = [i for i in range(10)]
+
+
+@pytest.fixture
+def ds():
+    return UnionFind(node_ids=IDs)
```
jmatak, line 9: Add support for testing other complex graph structures.

python/union_find.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def union(self, node1_ids: List[int], node2_ids: List[int]) -> None:
+        """
+        Executes a sequence of union operations on each pair of same-index nodes from two equal-length lists.
+
+        :param node1_ids: First list of node IDs
+        :param node2_ids: Second list of node IDs
+        """
+        for node1_id, node2_id in zip(node1_ids, node2_ids):
+            self.union_pair(node1_id, node2_id)
```
jmatak, line 87: Can you implement `union` to be able to get `Union[int, List[int]]`, for easier calls with one argument, which is in my opinion more common way. If this API has more sense to you, you can leave it as it is :smile:

python/union_find.py
antepusic: At initialization, each node is assigned with itself as its parent (and conversely its grandparent, see line 55).
Since there are no actions that would clear a node’s `.parent` attribute, at all times every node has a grandparent.

python/union_find.py
antepusic: This code moves down the tree towards its root.

Suppose that `node_id` is initially 9 in [this tree](github url comment). Its parent is 5 and thus `node_id` is then set to 5.
In the next iteration, `node_id` is 5 and `self.parent(node_id)` returns 2. The traversal continues until `node_id` reaches the root, whose parent is by convention identical to itself.

(For the sake of brevity, this explanation disregards line 74 but it applies in that case as well.)

python/union_find.py
antepusic: No problem, will do!

python/mage/union_find/__init__.py
```diff
@@ -0,0 +1,4 @@
+from mage.union_find.union_find import (  # noqa: F401, F402, F403
+    UnionFind,
+    Node,
+)
```
jmatak, line 4: Delete `python/__init__.py`. Couldn't make comment there since it is empty.

python/mage/union_find/union_find.py
```diff
@@ -0,0 +1,114 @@
...
+        :param node_id_1: First node's ID or list of node IDs
+        :param node_id_2: Second node's ID or list of node IDs
+        """
+        if type(node_id_1) is int and type(node_id_2) is int:
+            self.union_pair(node_id_1, node_id_2)
+        else:
+            for x, y in zip(node_id_1, node_id_2):
+                self.union_pair(node_id_1=x, node_id_2=y)
+
+    def union_pair(self, node_id_1, node_id_2) -> None:
```
jmatak, line 69: Missing type checks

python/mage/union_find/union_find.py
```diff
@@ -0,0 +1,114 @@
...
+        self, node_id_1: Union[int, List[int]], node_id_2: Union[int, List[int]]
+    ) -> None:
+        """
+        Executes a sequence of union operations on each pair of same-index nodes.
+        Supports two nodes or two equal-length lists thereof.
+
+        :param node_id_1: First node's ID or list of node IDs
+        :param node_id_2: Second node's ID or list of node IDs
+        """
+        if type(node_id_1) is int and type(node_id_2) is int:
```
jmatak, line 63: If you prefer, there is function called `isinstance()`

python/mage/union_find/__init__.py
antepusic: Done!

python/mage/union_find/union_find.py
antepusic: Added.

python/mage/union_find/union_find.py
antepusic: Thanks! :slightly_smiling_face:

jmatak (commented):
Nice work! Keep on working with end2end tests. Try to take a look at scalability and performance of this algorithm.

python/union_find.py
```diff
@@ -0,0 +1,94 @@
...
+                for node1, node2 in zip(nodes1, nodes2)
+            ]
+    elif mode.lower() in Mode.CARTESIAN:
+        return [
+            mgp.Record(
+                node1=node1,
+                node2=node2,
+                connected=disjoint_set.connected(node1_id=node1.id, node2_id=node2.id),
+            )
+            for node1, node2 in product(nodes1, nodes2)
```
Josipmrden, line 91: Does this perform better than 2 for loops?

Josipmrden (commented):
Small comment

e2e/union_find_test/test_cartesian/test.yml
```diff
@@ -0,0 +1,53 @@
+query: >
+  MATCH (m {id: 0}), (n {id: 2}), (o {id: 1}), (p {id: 3})
+  CALL union_find.connected([m, o], [n, p], "c") YIELD *
+  RETURN node1, node2, connected;
```
jmatak, line 4:
```suggestion
query: >
  MATCH (m {id: 0}), (n {id: 2}), (o {id: 1}), (p {id: 3})
  CALL union_find.connected([m, o], [n, p], "c") YIELD *
  RETURN node1.id as node1_id, node2.id as node2_id, connected
  ORDER BY node1_id, node2_id;
```

I suggest these changes because of output simplicity, as well as deterministic order

e2e/union_find_test/test_pairwise/test.yml
```diff
@@ -0,0 +1,29 @@
+query: >
+  MATCH (m {id: 0}), (n {id: 2}), (o {id: 1}), (p {id: 3})
+  CALL union_find.connected([m, o], [n, p], "p") YIELD *
+  RETURN node1, node2, connected;
```
jmatak, line 4: Change to returnin ids

jmatak (changes requested):
Change these e2e test with comments.

python/union_find.py
```diff
@@ -0,0 +1,94 @@
...
+) -> mgp.Record(node1=mgp.Vertex, node2=mgp.Vertex, connected=bool):
+    """
+    Returns whether two nodes (or each pair in the product of two node lists) belong to the same connected component
+    of the graph.
+    :param nodes1: Node or tuple of nodes
+    :type nodes1: Union[mgp.Vertex, Tuple[mgp.Vertex]]
+    :param nodes2: Node or tuple of nodes
+    :type nodes2: Union[mgp.Vertex, Tuple[mgp.Vertex]]
+    :param mode: Mode of operation: `p` / `pairwise` for pairwise similarity calculation or `c` / `cartesian`
+    for calculating the similarity of the Cartesian product of given tuples. Default value is `pairwise`.
```
Arya513, line 37: This part of the documentation needs to be fixed.

e2e/union_find_test/test_pairwise/test.yml
antepusic: Changed!

python/union_find.py
antepusic: Done!

e2e/union_find_test/test_cartesian/test.yml
antepusic: Nice catch, done!

jmatak (approved):
:+1: 
Think we are good here

python/union_find.py
antepusic: List comprehensions tend to be 5% faster.

Arya513 (approved):
Great job!

python/union_find.py
```diff
@@ -0,0 +1,96 @@
...
+            return [
+                mgp.Record(
+                    node1=node1,
+                    node2=node2,
+                    connected=disjoint_set.connected(
+                        node1_id=node1.id, node2_id=node2.id
+                    ),
+                )
+                for node1, node2 in zip(nodes1, nodes2)
+            ]
```
antoniofilipovic, line 84: This else branch here is not needed I would say. You can just use return statement.

python/union_find.py
```diff
@@ -0,0 +1,96 @@
...
+        return [
+            mgp.Record(
+                node1=node1,
+                node2=node2,
+                connected=disjoint_set.connected(node1_id=node1.id, node2_id=node2.id),
+            )
+            for node1, node2 in product(nodes1, nodes2)
+        ]
+    else:
+        x = f"Invalid mode of operation specified {mode.lower()}, {Mode.PAIRWISE}."
```
antoniofilipovic, line 95: Here a more friendly message could be given to use, e.g.
Mode "x" is invalid, please specify one of the following: pairwise, cartesian.

antepusic: Done!

python/union_find.py
antepusic: Handled this in my last commit.

PR #55, opened 2021-08-06
[main < T044-MAGE] Add optional -p flag to copy files to folder

antoniofilipovic:
This way user can set a flag, and then there is no need for two steps: build + copy.


rust/rsmgp-example/Cargo.lock
```diff
@@ -1,7 +1,5 @@
 # This file is automatically @generated by Cargo.
 # It is not intended for manual editing.
-version = 3
-
```
antoniofilipovic, line 4: Don't know why this happened 😄 I can revert this if needed

build
```diff
@@ -109,3 +126,14 @@ for project in filter(
...
+#################################################
+#            Optional copy dist folder          #
+#################################################
+
+arguments = get_arguments()
+
+if arguments.path is not None:
+    print('Copying files from folder {src} to {dst} folder.'.format(src=QUERY_MODULES_DIRECTORY, dst=arguments.path))
+    copy_dist(arguments.path)
+    print('Copying done!')
```
antoniofilipovic, line 139: This structure didn't make much sense to me, but I continued with what was already there. I would suggest that we add 
 main() function and that this part getArguments() is first part and then copy at the end.

jmatak: Reason why it was written as a script was mimicking the bash script. If more functionalities are to be adapted here, we might benefit from  reconstruction

rust/rsmgp-example/Cargo.lock
jmatak: Yes, revert it, the reason is probably difference between Buda's and your version of Cargo

build
antoniofilipovic: Makes sense

build
```diff
@@ -109,3 +126,14 @@ for project in filter(
...
     dst_file = os.path.join(QUERY_MODULES_DIRECTORY, module)
     shutil.copy2(src_file, dst_file)
+
+#################################################
+#            Optional copy dist folder          #
+#################################################
+
+arguments = get_arguments()
+
+if arguments.path is not None:
```
jmatak, line 136: Is there a better way to check if parameter is existing in argparse?

antoniofilipovic: I will check this out.

antoniofilipovic: I found a slightly better way.

antoniofilipovic: Check if it is okay now.

build
```diff
@@ -7,6 +7,8 @@ import glob
 import fnmatch
 import multiprocessing as mp
 
+import argparse
+
```
jmatak, line 11: The thing that came to my mind: is the `argparse` part of the standard Python library, because if it isn't, it would introduce unnecessary dependency for users trying to build MAGE.

antoniofilipovic: I didn't have to install anything - and also it is library on Python docs so I think it is
https://docs.python.org/3/library/argparse.html

antoniofilipovic: I double-checked - it is

antoniofilipovic: Awesome 😄 put it on approve 🛩️

jmatak (changes requested):
One more thing, add this to the README, so users are aware that this exists

jmatak (approved):

PR #56, opened 2021-08-17
Add black-box script for local build, reduce Docker image size

antoniofilipovic:
Create a black-box script for local development to handle:

- [x] build (and optional copy of mage/dist
- [x] setting up different query_modules directory
- [x] setting up mage/dist as query module directory and doing build in one command -> this is what user wants, easy to use command.

This PR  also contains a Docker development image build and Docker production image build.

README.md updated to follow new changes.


setup
```diff
@@ -0,0 +1,212 @@
...
+    if arguments.action == "modules":
+        if arguments.path is None:
+            path = f"{pathlib.Path().resolve()}/dist"
+        else:
+            path = arguments.path
+
+        if arguments.mg_conf_path:
+            mg_conf_path = arguments.mg_conf_path
+        else:
+            mg_conf_path = "/etc/memgraph/memgraph.conf"
```
jmatak, line 203: Logic is different for these two. At first, default item is in the first `if` branch, and later it is under else `branch`

README.md
```diff
@@ -31,51 +31,94 @@
...
 ```
 
-### Installing MAGE locally
-**1.** Run the `build` script. It will generate a `dist` directory with all the needed files. If you want to skip step 2)
-you can add optional -p (--path) flag which represents where will contents of `dist` directory be copied.
+This will build any new algorithm added to MAGE, and load it inside Memgraph.
+
+### 2. Installing MAGE locally with Linux package of Memgraph
+
+In second option to run MAGE, you need Linux based Memgraph. Since Memgraph needs to load MAGE's modules, there is the
```
jmatak, line 92: Add link to this type of installation in MG documentation. This seems hard to follow for someone not understanding what "Linux based Memgraph" actually is

setup
```diff
@@ -0,0 +1,210 @@
...
+    log.log(level=log.INFO, msg="Starting building of source code...")
+    build()
+    log.log(level=log.INFO, msg="Building done.")
+
+    if copy_dist_dest_path is None:
+        log.log(level=log.INFO,
+                msg="Skipping copy of contents of 'dist' folder to specified folder since --path is not specified")
+        return
+
+    log.log(level=log.INFO,
```
jmatak, line 171: Did you fix "copying from None"?

jmatak (commented):
I have one more idea, maybe we can filter out which algorithms do we want in our build, like build only CPP or build only Python .Something to discuss, I need to test this first

setup
```diff
@@ -0,0 +1,210 @@
...
+    if modules_storage_path_param is None:
+        modules_storage_path = f"{pathlib.Path().resolve()}/dist"
+    else:
+        modules_storage_path = modules_storage_path_param
+
+    log.log(level=log.INFO,
+            msg="--query-modules-dir flag in {memgraph_conf} will be set to {path}".format(path=modules_storage_path,
+                                                                                       memgraph_conf=MG_CONF_PATH))
+
+    subprocess.run(['sed', '-E', '-i', "s@.*(--query-modules-directory).*@\\1={path}@".format(path=modules_storage_path), MG_CONF_PATH  ])
```
jmatak, line 189: Black magic
Is this changing the flag or appending to the previously set flag?

setup
antoniofilipovic: Yes 😄

setup
antoniofilipovic: This is completely changing it. 
What about black magic 😄 do you think this is not good or it can somehow be better?

MasterMedo:
1. Can you write a high level overview in the pull request description of what exactly the script is supposed to do?
2. Why isn't this a bash script? This is definitely an example of a bad use-case for python.
3. Should this be compatible with all operating systems?

jmatak:
> 1. Can you write a high level overview in the pull request description of what exactly the script is supposed to do? (step-by-step process)
> 2. Why isn't this a bash script? This is definitely an example of a bad use-case for python.
> 3. Should this be compatible with all operating systems?

The decision to use Python previously was based on a fact that we want to support as many operating systems as possible. However, since there is no local Windows and Mac installation, there is no need to put focus on that till we have working Win and Mac instances.

jmatak (approved):

PR #57, opened 2021-08-19
Add UUID generator module

g-despot:
A very simple query module that generates string UUIDs.


cpp/uuid_module/uuid_module.cpp
```diff
@@ -0,0 +1,40 @@
...
+    if (record == nullptr) {
+      throw mg_exception::NotEnoughMemoryException();
+    }
+
+    mg_utility::InsertStringValueResult(record, kFieldUuid, string, memory);
+  } catch (const std::exception &e) {
+    // We must not let any exceptions out of our module.
+    mgp_result_set_error_msg(result, e.what());
+    return;
+  }
```
jmatak, line 28: Is the goal of this module to obtain a uuid for each node in graph?

g-despot: The goal is to return a newly generated string UUID just like in [this](https://neo4j.com/labs/apoc/4.1/graph-updates/uuid/#manual-uuids) APOC procedure.

jmatak: Alright that's cool, please write an e2e test before merging this to `main`

g-despot: The e2e test is written and committed, you can try it out.

jmatak (approved):

PR #60, opened 2021-09-07
Add filtering to E2E tests. 

jmatak:
- [x] Filter e2e tests with `-k` parameter
- [ ] Add gqlalchemy dependency


test_e2e
```diff
@@ -10,6 +11,16 @@ E2E_DIRECTORY = f"{WORK_DIRECTORY}/e2e"
 #                End to end tests               #
 #################################################
 
-os.environ["PYTHONPATH"] = E2E_DIRECTORY
-os.chdir(E2E_DIRECTORY)
-subprocess.run(["pytest", ".", "-vv"])
+def main(test_filter: str = None):
+    os.environ["PYTHONPATH"] = E2E_DIRECTORY
+    os.chdir(E2E_DIRECTORY)
+    command = ["pytest", ".", "-vv"]
```
antoniofilipovic, line 17: This can maybe go as const
`E2E_COMMAND = ["pytest", ".", "-vv"]`

test_e2e
```diff
@@ -10,6 +11,16 @@ E2E_DIRECTORY = f"{WORK_DIRECTORY}/e2e"
...
+    if test_filter:
+        command.extend(["-k", test_filter])
+    subprocess.run(command)
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='Test MAGE E2E, make sure to have your Memgraph instance ready.')
+    parser.add_argument('-k', help='Add filter to tests you want to run', type=str, required=False)
+    args = parser.parse_args()
+    main(args.k)
```
antoniofilipovic, line 26: What if we add more arguments, does it make sense to send to the main function `args` reference?

antoniofilipovic: Actually, I am not sure now, since this way we can use an easier main function as part of the external module, there is no need to send whole args reference then

test_e2e
```diff
@@ -10,6 +11,16 @@ E2E_DIRECTORY = f"{WORK_DIRECTORY}/e2e"
...
+    os.chdir(E2E_DIRECTORY)
+    command = ["pytest", ".", "-vv"]
+    if test_filter:
+        command.extend(["-k", test_filter])
+    subprocess.run(command)
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='Test MAGE E2E, make sure to have your Memgraph instance ready.')
+    parser.add_argument('-k', help='Add filter to tests you want to run', type=str, required=False)
+    args = parser.parse_args()
```
antoniofilipovic, line 25: This can maybe be part of function `parse_arguments()`

antoniofilipovic (commented):
It seems fine for me, left few comments just for discussion

test_e2e
jmatak: I thought about it and left it here because it is easier to notice it when debugging, and also at some point in time we would like to make this path "." modifiable.

test_e2e
jmatak: This is intended to be a standalone script, and not being used anywhere else. But the arguments could be extracted in a cleaner way

test_e2e
jmatak: You are correct, that could be done

README.md
```diff
@@ -102,6 +102,12 @@ python3 test_unit
 # Running end-to-end tests
 python3 test_e2e
 ```
+
+Furthermore, to test only specific end-to-end tests, you can add argument `-k` with substring refering to the algorithm that needs to be tested.
```
antoniofilipovic, line 106: Here I think we are missing, "For example to test module named `weakly_connected_components`, you would `write:python3 test_e2e -k weakly_connected_components`

antoniofilipovic: Or to change:
python3 test_e2e -k `<module_name>`

antoniofilipovic (approved):
There is just one comment on the usage part in README.md. I will put approve but consider changing this on README.md

test_e2e
antoniofilipovic: Okay 👍🏻

test_e2e
antoniofilipovic: Good point 👍🏻

README.md
```diff
@@ -102,6 +102,12 @@ python3 test_unit
 # Running end-to-end tests
 python3 test_e2e
 ```
+
+Furthermore, to test only specific end-to-end tests, you can add argument `-k` with substring refering to the algorithm that needs to be tested. To test module named `<query_module>`, you would have to run `python3 test_e2e -k <query_module>` where `<query_module>` is the name of the specific module you want to test.
```
MasterMedo, line 106:
```suggestion
Furthermore, to test only specific end-to-end tests, you can add argument `-k` with substring referring to the algorithm that needs to be tested. To test a module named `<query_module>`, you would have to run `python3 test_e2e -k <query_module>` where `<query_module>` is the name of the specific module you want to test.
```

MasterMedo (approved):

PR #61, opened 2021-09-07
Fix betweenness centrality algorithm name typo

### LINKED ISSUE #59: betweenness_centrality returns a score with typo
I'm not sure if there is a specific reason for it, but the `betweenness_centrality` module returns the score as `betweeenness_centrality` (3 e instead of 2).
On the [docs page](https://docs.memgraph.com/mage/query-modules/cpp/betweenness-centrality/) the same thing is repeated which made me question if it was accidental or intentional.

jmatak:
Return value was called `betweeenness_centrality` which has a typo

Closes #59

### DIFF
--- a/cpp/betweenness_centrality_module/betweenness_centrality_module.cpp
+++ b/cpp/betweenness_centrality_module/betweenness_centrality_module.cpp
@@ -7,7 +7,7 @@
 
 namespace {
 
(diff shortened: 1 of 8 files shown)


antonio2368 (approved):
Very nice :smile:

PR #65, opened 2021-09-16
Reduce docker image size, add black-box script for Docker

### LINKED ISSUE #64: [ENHANCEMENT] Docker image size getting bigger and bigger
Currently, the Docker image size is around 3GB and will probably get bigger because of dependencies.

We could solve that problem by introducing Docker stages build so that we can only build `python`, `cpp` or `rust` algorithms. If someone needs all algorithms to build, that would also be possible.

antoniofilipovic:
Goal of this PR is to solve following issues:

- [x] reduce docker image size
- [ ] add black-box script to import files to docker container, install python dependencies, and run build
- [ ] add versioning with memgraph **(new PR)**

Closes  #64

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,5 +4,5 @@
 htmlcov
 dist
 **/__pycache__
-**/*.pyc
 cpp/build
+Dockerfile
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,26 +12,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
-    - name: Cache Memgraph Docker image
-      id: cache-memgraph-docker
-      uses: actions/cache@v1
-      with:
-        path: ~/memgraph
-        key: cache-memgraph-v${{ env.MG_VERSION }}-docker-image
-    - name: Download Memgraph Docker image
-      if: steps.cache-memgraph-docker.outputs.cache-hit != 'true'
-      run: |
-        mkdir ~/memgraph
-        curl -L https://memgraph.com/download/memgraph/v${{ env.MG_VERSION }}/docker/memgraph-${{ env.MG_VERSION }}-community-docker.tar.gz > ~/memgraph/memgraph-docker.tar.gz
-    - name: Load Memgraph Docker image
-      run: |
-        docker load -i ~/memgraph/memgraph-docker.tar.gz
     - name: Build project
       run: |
-        docker build -t memgraph:mage --target development .
+        docker build -t memgraph:mage --target dev .
     - name: Run Memgraph MAGE image
       run: |
-        docker run -d -p 7687:7687 memgraph:mage --telemetry-enabled=False
+        docker run -d -p 7687:7687 memgraph:mage
     - name: Set up C++
       run: |
         sudo apt update
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,7 +1,6 @@
 name: Build and Test
 
 env:
-  MG_VERSION: '1.4.0'
   PY_VERSION: '3.7'
   CORE_COUNT: '8'
 
@@ -12,12 +11,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
-    - name: Build project
+    - name: Build Memgraph MAGE dev image
       run: |
-        docker build -t memgraph:mage --target dev .
+        docker build -t memgraph-mage .
     - name: Run Memgraph MAGE image
       run: |
-        docker run -d -p 7687:7687 memgraph:mage --telemetry-enabled=False
+        docker run -d -p 7687:7687 --name mage memgraph-mage --telemetry-enabled=False
     - name: Set up C++
       run: |
         sudo apt update
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,6 +3,7 @@ name: Build and Test
 env:
   PY_VERSION: '3.7'
   CORE_COUNT: '8'
+  MAGE_CONTAINER: 'mage'
 
 on: [push]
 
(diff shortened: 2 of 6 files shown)


Dockerfile
```diff
@@ -1,37 +1,48 @@
...
+    libpython3.7    `memgraph` \
+    libssl1.1       `memgraph` \
+    openssl         `memgraph` \
+    python3         `mage-memgraph` \
+    python3-pip     `mage-memgraph` \
+    --no-install-recommends \
+    && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+#Memgraph
+RUN pip3 install networkx==2.4 numpy==1.19.2 scipy==1.5.2 \
```
jmatak, line 20: What are these dependencies, and why are they required here?

Dockerfile
```diff
@@ -1,37 +1,48 @@
...
+    && curl https://download.memgraph.com/memgraph/v1.6.1/debian-10/memgraph_1.6.1-community-1_amd64.deb --output memgraph.deb \
```
jmatak, line 21: The MG version should be a parameter, and not hardcoded

.github/workflows/test.yml
```diff
@@ -12,26 +12,12 @@ jobs:
...
-      run: |
-        docker load -i ~/memgraph/memgraph-docker.tar.gz
     - name: Build project
       run: |
-        docker build -t memgraph:mage .
+        docker build -t memgraph:mage --target dev .
     - name: Run Memgraph MAGE image
       run: |
-        docker run -d -p 7687:7687 memgraph:mage --telemetry-enabled=False
+        docker run -d -p 7687:7687 memgraph:mage
```
jmatak, line 20: Disable telemetry, we do not want to have traffic by GithubActions

Dockerfile
```diff
@@ -1,37 +1,76 @@
...
-
-# Build all necessary modules
-RUN python3 build
-RUN cp -r /mage/dist/* /usr/lib/memgraph/query_modules/
-
-# It's required to install python3 because auth module scripts are going to be
-# written in python3.
-RUN python3 -m  pip install -r /mage/python/requirements.txt
+#Memgraph
+RUN curl https://download.memgraph.com/memgraph/v1.6.1/debian-10/memgraph_1.6.1-community-1_amd64.deb --output memgraph.deb \
```
jmatak, line 24: Add MG version as parameter in Dockerfile

Dockerfile
```diff
@@ -1,37 +1,76 @@
...
 ENV LD_LIBRARY_PATH /usr/lib/memgraph/query_modules
+
 # Memgraph listens for Bolt Protocol on this port by default.
 EXPOSE 7687
+
 # Snapshots and logging volumes
 VOLUME /var/log/memgraph
 VOLUME /var/lib/memgraph
+# Configuration volume
 VOLUME /etc/memgraph
```
jmatak, line 37: Seems these are not necessary

python/requirements.txt
```diff
@@ -1,2 +1,4 @@
+numpy==1.21.1
+scipy==1.7.1
```
jmatak, line 2: How this came up here?

setup
```diff
@@ -177,16 +177,15 @@ def run_build_action(copy_dist_dest_path: str):
...
 
     log.log(level=log.INFO,
             msg="--query-modules-dir flag in {memgraph_conf} will be set to {path}".format(path=modules_storage_path,
-                                                                                       memgraph_conf=MG_CONF_PATH))
+                                                                                           memgraph_conf=MG_CONF_PATH))
 
-    subprocess.run(['sed', '-E', '-i', "s@.*(--query-modules-directory).*@\\1={path}@".format(path=modules_storage_path), MG_CONF_PATH  ])
+    subprocess.run(
+        ['sed', '-E', '-i', "s@.*(--query-modules-directory).*@\\1={path}@".format(path=modules_storage_path),
+         MG_CONF_PATH])
```
jmatak, line 188: We should talk about using `sed` here. Or discuss using bash script for the whole process

Dockerfile
antoniofilipovic: Of course, didn't forgot about it as I am still working on it.

Dockerfile
antoniofilipovic: Yes it does but it is how it is on [here](https://github.com/memgraph/memgraph/blob/master/release/docker/memgraph_community.dockerfile) so I thought it would be good idea to add it here to. It serves just to show where are volumes, but it is not necessary.

python/requirements.txt
antoniofilipovic: Because of `nxalg` module. [Here](https://memgraph.com/docs/mage/query-modules/python/nxalg)

antoniofilipovic: Because of `nxalg` module. [Here](https://memgraph.com/docs/mage/query-modules/python/nxalg)

.dockerignore
```diff
@@ -4,5 +4,5 @@
 htmlcov
 dist
 **/__pycache__
-**/*.pyc
-cpp/build
\ No newline at end of file
+cpp/build
+Dockerfile
```
antoniofilipovic, line 8: Added this so that when I change Dockerfile, the complete MAGE folder is not copied again - the whole process is lot slower for local dev.

Dockerfile
antoniofilipovic: Of course, didn't forget about it, just still under development.

setup
antoniofilipovic: Yes of course, what do you mean?

antoniofilipovic: Yes of course, what do you mean?

.github/workflows/test.yml
```diff
@@ -12,23 +12,9 @@ jobs:
...
-      if: steps.cache-memgraph-docker.outputs.cache-hit != 'true'
-      run: |
-        mkdir ~/memgraph
-        curl -L https://memgraph.com/download/memgraph/v${{ env.MG_VERSION }}/docker/memgraph-${{ env.MG_VERSION }}-community-docker.tar.gz > ~/memgraph/memgraph-docker.tar.gz
-    - name: Load Memgraph Docker image
-      run: |
-        docker load -i ~/memgraph/memgraph-docker.tar.gz
     - name: Build project
       run: |
-        docker build -t memgraph:mage .
```
antoniofilipovic, line 31: @jmatak was this a bug? Why are we here building `memgraph:mage` and we are deploying `memgraph-mage`?

Dockerfile
jmatak: Remove them since they won't be used anymore, although you got the point of them being informatioal

Dockerfile
```diff
@@ -1,37 +1,82 @@
...
+
+#copy modules
+COPY --from=dev /usr/lib/memgraph/query_modules/ /usr/lib/memgraph/query_modules/
+
+#copy python build
+COPY --from=dev /usr/local/lib/python${PY_VERSION} / /usr/local/lib/python${PY_VERSION} /
+
+
+RUN rm -rf /mage \
+    && export PATH="/usr/local/lib/python3.7:${PATH}" \
```
jmatak, line 76: Py-version flag

.dockerignore
jmatak: https://docs.docker.com/develop/develop-images/build_enhancements/

This is how built might be enhanced in terms of speed

antoniofilipovic: I tested it out, maybe 40 seconds faster. I will push it to GitHub, maybe it helps there more.

Dockerfile
antoniofilipovic: Removed it.

jmatak (approved):
Let's merge this and continue on the other branch

PR #66, opened 2021-09-17
LabelRankT dynamic community detection algorithm

antepusic:
- [x] Standalone algorithm implementation
- [x] Query module
- [x] Unit tests
- [x] End-to-end tests
- [x] Weighted graph support
- [x] Docs documentation


cpp/community_detection_module/algorithm/label_propagation.cpp
```diff
@@ -0,0 +1,291 @@
...
+namespace LabelRankT {
+
+///@brief Checks if set is subset of another set.
+///
+///@tparam T -- set element data type
+///
+///@param set1 -- subset candidate
+///@param set2 -- set against which the check is performed
+///
+///@return -- whether set1 is subset of set2
```
jmatak, line 18: From the implementation, it sounds as we are working with vectors, not sets

cpp/community_detection_module/algorithm/label_propagation.cpp
```diff
@@ -0,0 +1,291 @@
...
+  return DEFAULT_WEIGHT;
+}
+
+void LabelRankT::set_structures(std::uint64_t node_id) {
+  times_updated[node_id] = 0;
+  std::unordered_map<std::uint64_t, double> node_label_Ps;
+
+  double sum_w_i = w_selfloop;
+  for (const auto node_j : graph->Neighbours(node_id))
+    sum_w_i += get_weight(node_j.node_id, node_id);
```
jmatak, line 42: Use curly brackets even when there is only one line of code in the for-loop. It makes code easier to read.

cpp/community_detection_module/algorithm/label_propagation.cpp
antepusic: Good catch, thanks!

cpp/community_detection_module/algorithm/label_propagation.cpp
```diff
@@ -0,0 +1,291 @@
...
+  }
+
+  return calculate_labels();
+}
+
+std::unordered_map<std::uint64_t, std::uint64_t> LabelRankT::calculate_labels(
+    std::uint64_t max_iterations, std::uint64_t max_updates,
+    std::unordered_set<std::uint64_t> changed_nodes,
+    std::unordered_set<std::uint64_t> to_delete) {
+  bool incremental = changed_nodes.size() > 1 ? true : false;
```
jmatak, line 242: Is it incremental if change_nodes.size() == 1?

cpp/community_detection_module/algorithm/label_propagation.cpp
```diff
@@ -0,0 +1,291 @@
...
+
+  if (incremental) {
```
jmatak, line 244: I need explanation for this if-statement

cpp/community_detection_module/algorithm/label_propagation.cpp
```diff
@@ -0,0 +1,291 @@
...
+      updated_label_Ps;
+
+  bool none_updated = true;
+  std::uint64_t most_updates = 0;
+
+  for (const auto node : graph->Nodes()) {
+    if (incremental) {
+      bool was_updated = changed_nodes.count(node.id) > 0 ? true : false;
+      if (not was_updated) continue;
+    }
```
jmatak, line 196:
```suggestion
    bool was_updated = changed_nodes.count(node.id) > 0 ? true : false;
    if (incremental && !was_updated) {
        continue;
    }
```

jmatak: It seems that you are trying to avoid counting nodes if it is not `incremental`, which also works

cpp/community_detection_module/algorithm/label_propagation.cpp
```diff
@@ -0,0 +1,291 @@
...
+  }
+
+  std::vector<std::uint64_t> most_probable_labels = {};
+
+  for (const auto [label, P] : label_Ps[node_id]) {
+    if (P == max_P) most_probable_labels.push_back(label);
+  }
+
+  return most_probable_labels;
+}
```
jmatak, line 103: Maybe ordered map where we check only maximum values?

cpp/community_detection_module/algorithm/label_propagation.cpp
```diff
@@ -0,0 +1,291 @@
...
+    node_i_degree++;
+
+    std::vector<std::uint64_t> labels_j = most_probable_labels(node_j.node_id);
+
+    if (is_subset(labels_i, labels_j)) label_similarity++;
+  }
+
+  if (label_similarity <= node_i_degree * similarity_threshold) return true;
+
+  return false;
```
jmatak, line 121: return 
```suggestion
  return label_similarity <= node_i_degree * similarity_threshold;
```

cpp/community_detection_module/algorithm/label_propagation.cpp
```diff
@@ -0,0 +1,291 @@
...
+
+  for (const auto [label, _] : node_label_Ps) {
+    double inflated_node_label_Ps = pow(node_label_Ps[label], exponent);
+    node_label_Ps[label] = inflated_node_label_Ps;
+    sum_Ps += inflated_node_label_Ps;
+  }
+
+  for (const auto [label, _] : node_label_Ps) node_label_Ps[label] /= sum_Ps;
+
+  return;
```
jmatak, line 157: No need for a return

jmatak (changes requested):
There are things to discuss :smile:

cpp/community_detection_module/community_detection_module.cpp
```diff
@@ -0,0 +1,260 @@
...
+constexpr char const *kFieldCommunity = "community";
+
+constexpr char const *kCreatedVertices = "createdVertices";
+constexpr char const *kCreatedEdges = "createdEdges";
+constexpr char const *kUpdatedVertices = "updatedVertices";
+constexpr char const *kUpdatedEdges = "updatedEdges";
+constexpr char const *kDeletedVertices = "deletedVertices";
+constexpr char const *kDeletedEdges = "deletedEdges";
+
+auto empty_graph = mg_generate::BuildGraph(0, {});
```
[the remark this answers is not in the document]
antepusic: I’ve defined `algorithm` as a global variable so that it can persist between function calls.
Simply declaring `LabelRankT::LabelRankT algorithm;` doesn’t work because the class lacks a default constructor, and in fact it can’t have one since the constructor takes a reference to the graph. To get around this, I initially define `algorithm` with a dummy empty graph.

Definitely open for a more elegant solution, tho!

cpp/community_detection_module/algorithm/label_propagation.cpp
antepusic: Thanks for the tip!

cpp/community_detection_module/algorithm/label_propagation.cpp
antepusic: Real eye for detail :eyes:

cpp/community_detection_module/algorithm/label_propagation.cpp
antepusic: If the algorithm is run normally, its data structures are initialized for the entire graph, hence `graph->Nodes()` in the iterator.
With incremental operation, most of the graph is unchanged so the data structure entries are only updated for the changed nodes (and deleted node entries are erased).

cpp/community_detection_module/algorithm/label_propagation.cpp
antepusic: That was indeed what I was going for!

cpp/community_detection_module/algorithm/label_propagation.cpp
antepusic: As `propagate()` is the bottleneck of the algorithm and heavily retrieves label probabilities, I ultimately made the design choice to use an unordered map because of it having O(1) retrieval.

cpp/community_detection_module/community_detection_module.cpp
```diff
@@ -0,0 +1,260 @@
...
+    if (initialized) {
+      std::vector<std::uint64_t> modified_nodes;
+      std::vector<std::pair<std::uint64_t, std::uint64_t>> modified_edges;
+      std::vector<std::uint64_t> deleted_nodes_;
+      std::vector<std::pair<std::uint64_t, std::uint64_t>> deleted_edges_;
+
+      // Did this one blindly, suggestions welcome
+      for (auto i = 0UL; i < mgp_list_size(created_nodes); i++)
+        modified_nodes.push_back(
+            mgp_vertex_get_id((mgp_vertex *)mgp_list_at(created_nodes, i))
```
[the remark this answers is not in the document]
antepusic: Would `mg_utils.hpp` be a good place for these?

cpp/community_detection_module/algorithm/label_propagation.cpp
jmatak: But you  are always having 2 iterations over whole set. In the other case, you would only have to iterate through first few similar elements.

cpp/community_detection_module/algorithm/label_propagation.cpp
```diff
@@ -0,0 +1,282 @@
...
+  return most_probable_labels;
+}
+
+bool LabelRankT::distinct_enough(mg_graph::Node<uint64_t> node_i,
+                                 double similarity_threshold) {
+  std::vector<std::uint64_t> labels_i = most_probable_labels(node_i.id);
+  std::uint64_t label_similarity = 0;
+
+  int node_i_degree = 0;
+  for (const auto node_j : graph->Neighbours(node_i.id)) {
```
jmatak, line 109: Take a note here and think about the dynamic case. The neighbours in this case will be in the state after processing the query, meaning you wouldn't have deleted edges, and would have newly added ones. Hope that is fine for the algorithm

cpp/community_detection_module/community_detection_module.cpp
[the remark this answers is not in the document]
jmatak: This is then a reasonable workaround, however, you might put it within algorithm implementation and not in the module, so the same structure can be used for unit-testing

jmatak (commented):
Now it's time to test the dynamic part of the algorithm :smile:

cpp/community_detection_module/community_detection_module.cpp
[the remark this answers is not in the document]
jmatak: Yes, exactly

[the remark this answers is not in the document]
jmatak: That would be the only place for it because we do not want these C-API procedures outside of there

cpp/community_detection_module/algorithm/label_propagation.cpp
antepusic: According to the paper (https://arxiv.org/pdf/1305.2006.pdf), the set that is iterated upon is very small thanks to the cutoff operation:
> More importantly, Φr is shown empirically to
reduce the space complexity efficiently, from quadratic to
linear. The average number of labels in each node is typically less than 3.0 for r = 0.1.

Does this check out or is an ordered map still preferable?

cpp/community_detection_module/algorithm/label_propagation.cpp
antepusic: This is in line with the algorithm design :+1:

cpp/community_detection_module/algorithm/label_propagation.cpp
jmatak: If less than 3, than totally fine :smile:

jmatak (approved):

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
```diff
@@ -0,0 +1,370 @@
...
+    if (P == max_P) most_probable_labels.push_back(label);
+  }
+
+  return most_probable_labels;
+}
+/* #endregion */
+
+/* #region label_propagation_steps */
+bool LabelRankT::DistinctEnough(std::uint64_t node_i_id,
+                                double similarity_threshold) {
```
Josipmrden, line 175: const I guess, since it won't be changed in this method?
Try adapting every variable to const which is not mutated in the methods

Josipmrden (commented):
Small comments

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
antepusic: Makes sense! I’m going to comb through the code and make parameters `const` where possible.

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
```diff
@@ -0,0 +1,370 @@
...
+    times_updated[node_id] = 0;
+  }
+}
+
+std::int64_t LabelRankT::NodeLabel(std::uint64_t node_id) {
+  std::int64_t node_label = -1;
+  double max_P = 0;
+
+  for (const auto [label, P] : label_Ps[node_id]) {
+    if (P > max_P || (P == max_P && (std::int64_t)label < node_label)) {
```
antaljanosbenjamin, line 117: As far as I see, `label` is type of `uint64_t`, thus `(std::int64_t)label < node_label` can only be true if `label > std::numeric_limits<int64_t>::max()`, because it will overflow. Am I missing something here?

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
```diff
@@ -0,0 +1,370 @@
...
+  }
+
+  return node_label;
+}
+
+std::unordered_map<std::uint64_t, std::int64_t> LabelRankT::AllLabels() {
+  std::unordered_map<std::uint64_t, std::int64_t> labels;
+
+  std::set<std::int64_t> labels_ordered;
+  std::map<std::int64_t, std::int64_t> lookup;
```
antaljanosbenjamin, line 130:
```suggestion
  std::unordered_map<std::int64_t, std::int64_t> lookup;
```
Why not `unordered_map`?

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
```diff
@@ -0,0 +1,370 @@
...
+
+  int64_t label_i = 1;
+  for (const auto label : labels_ordered) {
+    if (label == -1) {  // no label found
+      lookup[label] = label;
+    } else {
+      lookup[label] = label_i;
+      label_i++;
+    }
+  }
```
antaljanosbenjamin, line 144:
```suggestion
  bool insert_no_label = false;
  for (const auto label : labels_ordered) {
    if (label == -1) {  // no label found
      insert_no_label = true;
    } else {
      lookup[label] = label_i;
      label_i++;
    }
  }
  if (insert_no_label) {
    lookup[-1] = -1;
  }
```
I don't see a reason to constantly assign the same value.

antaljanosbenjamin: Or even better:
```suggestion
  labels_ordered.erase(-1);
  for (const auto label : labels_ordered) {
    lookup[label] = label_i;
    label_i++;
  }
  lookup[-1] = -1;
```

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
```diff
@@ -0,0 +1,370 @@
...
+
+std::vector<std::uint64_t> LabelRankT::MostProbableLabels(
+    std::uint64_t node_id) {
+  double max_P = 0;
+
+  for (const auto [label, P] : label_Ps[node_id]) {
+    if (P > max_P) max_P = P;
+  }
+
+  std::vector<std::uint64_t> most_probable_labels = {};
```
antaljanosbenjamin, line 163:
```suggestion
  std::vector<std::uint64_t> most_probable_labels;
```
Just  small style change, because this is the same but shorter.

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
```diff
@@ -0,0 +1,370 @@
...
+    double min_value) {
+  std::vector<std::uint64_t> to_be_removed;
+
+  for (const auto [label, P] : node_label_Ps) {
+    if (P < min_value) to_be_removed.push_back(label);
+  }
+
+  for (const auto label : to_be_removed) {
+    node_label_Ps.erase(label);
+  }
```
antaljanosbenjamin, line 238:
```suggestion
    for (auto it = node_label_Ps.begin(); it != node_label_Ps.end();) {
        if (it->second < min_value) {
            it = m.erase(it);
        } else {
            ++it;
        }
    }
```
This way there is no need to use an extra vector and the erase can be done in single loop.

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
```diff
@@ -0,0 +1,370 @@
...
+///
+///@param v1 -- subset candidate
+///@param v2 -- vector against which the check is performed
+///
+///@return -- whether v1 is subset of v2
+template <class T>
+bool IsSubset(std::vector<T> const& v1, std::vector<T> const& v2) {
+  for (const auto& element : v1) {
+    if (std::find(v2.begin(), v2.end(), element) == v2.end()) return false;
+  }
```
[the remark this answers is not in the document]
antepusic: This method is only used to check if a node’s labels form a subset of its neighbor’s labels. The number of labels stored for a node ought to be small – quoting the paper, “the average number of labels in each node is typically less than 3.0 for r = 0.1.” (https://arxiv.org/pdf/1305.2006.pdf)

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
```diff
@@ -0,0 +1,370 @@
...
+///@brief Checks if vector is subset of another vector.
+///
+///@tparam T -- set element data type
+///
+///@param v1 -- subset candidate
+///@param v2 -- vector against which the check is performed
+///
+///@return -- whether v1 is subset of v2
+template <class T>
+bool IsSubset(std::vector<T> const& v1, std::vector<T> const& v2) {
```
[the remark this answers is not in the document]
antepusic: Good catch, thanks!

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
```diff
@@ -0,0 +1,370 @@
...
+  }
+
+  for (const auto label : to_be_removed) {
+    node_label_Ps.erase(label);
+  }
+}
+
+std::pair<bool, std::uint64_t> LabelRankT::Iteration(
+    bool incremental, std::unordered_set<std::uint64_t> changed_nodes,
+    std::unordered_set<std::uint64_t> to_delete) {
```
antaljanosbenjamin, line 243: Try to use references for the containers, otherwise they will be copied all around.

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
```diff
@@ -0,0 +1,370 @@
...
+    std::unordered_map<std::uint64_t, double> updated_node_label_Ps =
+        Propagate(node_id);
+
+    // inflation
+    Inflate(updated_node_label_Ps, exponent);
+
+    // cutoff
+    Cutoff(updated_node_label_Ps, min_value);
+
+    updated_label_Ps.insert({node_id, updated_node_label_Ps});
```
antaljanosbenjamin, line 272:
```suggestion
    updated_label_Ps.insert({node_id, std::move(updated_node_label_Ps)});
```

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
```diff
@@ -0,0 +1,370 @@
...
+
+  return {none_updated, most_updates};
+}
+/* #endregion */
+
+std::unordered_map<std::uint64_t, std::int64_t> LabelRankT::CalculateLabels(
+    std::unique_ptr<mg_graph::Graph<>>& graph,
+    std::unordered_set<std::uint64_t> changed_nodes,
+    std::unordered_set<std::uint64_t> to_delete, bool persist) {
+  this->graph = std::move(graph);
```
antaljanosbenjamin, line 290: This move is really sneaky. I would use `std::unique_ptr<mg_graph::Graph<>>&&` as parameter type, therefore the callers are forced to move the unique pointer as kind of explicitly handling over the ownership.

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
```diff
@@ -0,0 +1,370 @@
...
+
+  if (persist) this->calculated = true;
+  ResetTimesUpdated();
+
+  return AllLabels();
+}
+
+std::unordered_map<std::uint64_t, std::int64_t> LabelRankT::GetLabels(
+    std::unique_ptr<mg_graph::Graph<>>& graph) {
+  std::unordered_map<std::uint64_t, std::int64_t> labels;
```
antaljanosbenjamin, line 321:
```suggestion
```

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
```diff
@@ -0,0 +1,370 @@
...
+                                              deleted_nodes.end());
+
+  for (const auto edge : modified_edges) {
+    changed_nodes.insert(edge.first);
+    changed_nodes.insert(edge.second);
+  }
+
+  for (const auto edge : deleted_edges) {
+    if (to_delete.count(edge.first) == 0) changed_nodes.insert(edge.first);
+    if (to_delete.count(edge.second) == 0) changed_nodes.insert(edge.second);
```
antaljanosbenjamin, line 365:
```suggestion
    changed_nodes.insert(edge.first);
    changed_nodes.insert(edge.second);
```

The insert will handle the duplications without any additional counts.

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
```diff
@@ -0,0 +1,370 @@
...
+    sum_w.clear();
+    times_updated.clear();
+    for (const auto node_id : NodesMemgraphIDs()) {
+      SetStructures(node_id);
+    }
+  }
+
+  for (std::uint64_t i = 0; i < this->max_iterations; i++) {
+    auto [none_updated, most_updates] =
+        Iteration(incremental, changed_nodes, to_delete);
```
antaljanosbenjamin, line 309: `NodesMemgraphIDs` is called in `Iteration`. Because it is a quite expensive operation, it might worth to precalculate it before `Iteration` is called and pass it as a parameter.

cpp/dynamic_community_detection_module/dynamic_community_detection_module.cpp
```diff
@@ -0,0 +1,291 @@
...
+constexpr char const *kDeletedVertices = "deletedVertices";
+constexpr char const *kDeletedEdges = "deletedEdges";
+
+LabelRankT::LabelRankT algorithm = LabelRankT::LabelRankT();
+bool initialized = false;
+
+auto saved_directedness = false;
+auto saved_weightedness = false;
+std::string saved_weight_property = "weight";
+double saved_default_weight = 1.0;
```
antaljanosbenjamin, line 34: I don't exactly see how the query module is meant to be used, but static variables might cause some problems, e.g.: reloading the query module might erase their value.

cpp/mg_utility/mg_graph.hpp
```diff
@@ -181,7 +215,7 @@ class Graph : public GraphView<TSize> {
       return false;
     }
     if (edges_[edge_id].id == k_deleted_edge_id_) {
-      return false;
+      return false;
```
antaljanosbenjamin, line 218:
```suggestion
      return false;
```

cpp/mg_utility/mg_graph.hpp
```diff
@@ -263,28 +297,35 @@ class Graph : public GraphView<TSize> {
...
 
   std::vector<TNode> nodes_;
   std::vector<TEdge> edges_;
+
+  std::vector<double> weights_;
+
   std::unordered_map<TSize, std::uint64_t> inner_to_memgraph_id_;
   std::unordered_map<std::uint64_t, TSize> memgraph_to_inner_id_;
-
+
```
antaljanosbenjamin, line 328:
```suggestion

```

cpp/mg_utility/mgp.hpp
```diff
@@ -421,4 +421,4 @@ void proc_add_deprecated_result(struct mgp_proc *proc, const char *name, struct
 
 bool must_abort(struct mgp_graph *graph) { return mgp_must_abort(graph); }
 
-}  // namespace mgp
+}  // namespace mgp
```
antaljanosbenjamin, line 424:
```suggestion
}  // namespace mgp

```

antaljanosbenjamin (commented):
I might got something wrong, in that case feel free to ignore any of my comments.

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.hpp
```diff
@@ -0,0 +1,241 @@
...
+#include <unordered_set>
+#include <vector>
+
+namespace LabelRankT {
+
+class LabelRankT {
+ private:
+  /* #region parameters */
+  /// default edge weight
+  double DEFAULT_WEIGHT = 1;
```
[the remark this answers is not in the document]
antepusic: Thanks for the suggestions! I’ll make the same change in the module code.

cpp/dynamic_community_detection_module/dynamic_community_detection_module.cpp
antepusic: I believe this is acceptable here.

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
```diff
@@ -0,0 +1,370 @@
...
+  }
+
+  return true;
+}
+
+std::unordered_set<std::uint64_t> LabelRankT::InNeighborsMemgraphIDs(
+    std::uint64_t node_id) {
+  std::unordered_set<std::uint64_t> neighbors;
+
+  auto neighbor_set = this->is_directed
```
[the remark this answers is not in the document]
antepusic: Learned something new, thanks!

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
antepusic: Thanks for this! I’ve now avoided calling `NodesMemgraphIDs` more than once per algorithm execution.

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
antepusic: Fair enough!

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
antepusic: Corrected in the meantime!

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
antepusic: Very elegant, thanks!

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
antepusic: Good idea, added it.

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
antepusic: What is the rationale for this change? Just wondering if there’s something I’m not seeing here

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
```diff
@@ -0,0 +1,370 @@
...
+                          ? graph->InNeighbours(graph->GetInnerNodeId(node_id))
+                          : graph->Neighbours(graph->GetInnerNodeId(node_id));
+  for (const auto node_i : neighbor_set) {
```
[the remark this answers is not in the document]
antepusic: I’ve verified that the size is indeed 16 bytes. Since this is now resolved, could you DM me with the reasoning for this decision? I’m curious :slightly_smiling_face:

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
antepusic: Whereas I’ve already used references for containers in some places, I’ve now made edits to use them consistently.

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
antepusic: If we delete both a node and its incident edge, the deleted node would be added to `changed_nodes` per your suggestion, even though it should belong to `deleted_nodes` and nothing else.
This ensues in a crash - changed nodes have their communities updated and this cannot be done for a deleted node.

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
```diff
@@ -0,0 +1,371 @@
...
+    const bool is_weighted, const double similarity_threshold,
+    const double exponent, const double min_value,
+    const std::string weight_property, const double w_selfloop,
+    const std::uint64_t max_iterations, const std::uint64_t max_updates) {
+  this->is_directed = is_directed;
+  this->is_weighted = is_weighted;
+  this->similarity_threshold = similarity_threshold;
+  this->exponent = exponent;
+  this->min_value = min_value;
+  this->weight_property = weight_property;
```
kostasrim, line 338:
```suggestion
  this->weight_property = std::move(weight_property);
```

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
```diff
@@ -0,0 +1,371 @@
...
+  if (!calculated) return CalculateLabels(graph, {}, {}, false);
+
+  return AllLabels();
+}
+
+std::unordered_map<std::uint64_t, std::int64_t> LabelRankT::SetLabels(
+    std::unique_ptr<mg_graph::Graph<>>& graph, const bool is_directed,
+    const bool is_weighted, const double similarity_threshold,
+    const double exponent, const double min_value,
+    const std::string weight_property, const double w_selfloop,
```
kostasrim, line 331:
```suggestion
   std::string weight_property, const double w_selfloop,
```

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
```diff
@@ -0,0 +1,371 @@
...
+  this->w_selfloop = w_selfloop;
+  this->max_iterations = max_iterations;
+  this->max_updates = max_updates;
+
+  return CalculateLabels(graph);
+}
+
+std::unordered_map<std::uint64_t, std::int64_t> LabelRankT::UpdateLabels(
+    std::unique_ptr<mg_graph::Graph<>>& graph,
+    const std::vector<std::uint64_t> modified_nodes,
```
kostasrim, line 348:
```suggestion
    const std::vector<std::uint64_t> &modified_nodes,
```

kostasrim: same on other lines :)

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
```diff
@@ -0,0 +1,371 @@
...
+  return neighbors;
+}
+
+void LabelRankT::SetStructures(const std::uint64_t node_id) {
+  times_updated[node_id] = 0;
+  std::unordered_map<std::uint64_t, double> node_label_Ps;
+
+  double sum_w_i = w_selfloop;
+  for (const auto node_j_id : InNeighborsMemgraphIDs(node_id)) {
+    sum_w_i += GetTotalWeightBetween(node_j_id, node_id);
```
kostasrim, line 55: maybe `std::accumulate` ?

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.hpp
```diff
@@ -0,0 +1,241 @@
...
+  ///
+  ///@param from_node_id -- source node’s Memgraph ID
+  ///@param to_node_id -- target node’s Memgraph ID
+  ///
+  ///@return -- total weight of edges between from_node_id and to_node_id
+  double GetTotalWeightBetween(const std::uint64_t from_node_id,
+                               const std::uint64_t to_node_id);
+
+  ///@brief Retrieves given edge’s weight.
+  double GetWeight(const std::uint64_t edge_id);
```
kostasrim, line 92:
```suggestion
  double GetWeight(const std::uint64_t edge_id) const;
```

kostasrim: consider const qualifying the rest of the member functions that don't modify data members.

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.hpp
```diff
@@ -0,0 +1,241 @@
...
+
+  ///@brief Resets each node’s number of updates.
+  void ResetTimesUpdated();
+
+  /// Returns given node’s current community label.
+  /// If no label was found (e.g. with the min_value set too high),
+  /// the assigned label number is -1.
+  ///
+  /// @return -- given node’s current community label
+  std::int64_t NodeLabel(const std::uint64_t node_id);
```
kostasrim, line 102: You can omit `const` on a declaration when you pass by value and add it only on its definition (that is the source file). That way,  API  users don't get confused (from their point of view, they don't care if the function modifies its arguments because of the by value semantics)

kostasrim (commented):
Really good work and big patch! Some comments from me and I hope they are helpful :)

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
antepusic: Casting `label` to `int64_t` didn’t cause any such behavior here. I’ve prepared a pastebin @ https://pastebin.com/kELFfyBt for you to verify – if you run it, `2 2 2` should be printed out.

antepusic: Furthermore, I’ve run the unit tests and verified that `(std::int64_t)label < node_label` is in fact true in some tests.

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
```diff
@@ -0,0 +1,371 @@
...
+  return new_label_Ps_i;
+}
+
+void LabelRankT::Inflate(
+    std::unordered_map<std::uint64_t, double>& node_label_Ps,
+    const double exponent) {
+  double sum_Ps = 0;
+
+  for (const auto [label, _] : node_label_Ps) {
+    const double inflated_node_label_Ps = pow(node_label_Ps[label], exponent);
```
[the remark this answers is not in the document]
antepusic: As probabilities by definition are in the [0, 1] range, that is not the case. Anyhow, `std::pow` returns an appropriate value (+inf) in case an overflow occurs.

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
```diff
@@ -0,0 +1,371 @@
...
+  }
+
+  for (const auto label : to_be_removed) {
+    node_label_Ps.erase(label);
+  }
+}
+
+std::pair<bool, std::uint64_t> LabelRankT::Iteration(
+    const bool incremental,
+    const std::unordered_set<std::uint64_t> changed_nodes,
```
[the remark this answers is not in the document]
antepusic: Already done, thanks!

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
```diff
@@ -0,0 +1,371 @@
...
+    const std::unordered_set<std::uint64_t> to_delete) {
+  std::unordered_map<std::uint64_t, std::unordered_map<std::uint64_t, double>>
```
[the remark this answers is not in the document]
antepusic: There’s no reason here to use values instead of references, I’ve made the necessary changes per yours and Benjo’s suggestions.

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
antepusic: Done in the meantime!

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
```diff
@@ -0,0 +1,371 @@
...
+
+  return labels;
+}
+
+std::vector<std::uint64_t> LabelRankT::MostProbableLabels(
+    const std::uint64_t node_id) {
+  double max_P = 0;
+
+  for (const auto [label, P] : label_Ps[node_id]) {
+    if (P > max_P) max_P = P;
```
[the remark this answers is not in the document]
antepusic: I did consider that, however the code becomes much more unreadable because of the need to implement a custom comparator.

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
```diff
@@ -0,0 +1,371 @@
...
+///@tparam T -- set element data type
+///
+///@param v1 -- subset candidate
+///@param v2 -- vector against which the check is performed
+///
+///@return -- whether v1 is subset of v2
+template <class T>
+bool IsSubset(const std::vector<T>& v1, const std::vector<T>& v2) {
+  for (const auto& element : v1) {
+    if (std::find(v2.begin(), v2.end(), element) == v2.end()) return false;
```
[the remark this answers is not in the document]
antepusic: Good idea to leverage lambda expressions :+1:

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
antepusic: Unfortunately, it would require an exceedingly complex lambda function.

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.hpp
antepusic: Done!

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.hpp
antepusic: If we take into account that this API is used by this query module only, should these still be omitted?

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
antaljanosbenjamin: Yes! by moving it you can spare a copy! The new item that is inserted into `updated_label_Ps` will "steal" the resources of `updated_node_label_Ps`. I will write a explanation about what move is shortly.

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
antaljanosbenjamin: Oh sorry, I missed the part that the count is called on a different container, sorry.

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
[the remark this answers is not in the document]
antaljanosbenjamin: I can share the details here: the reasoning is on modern x64 CPUs the parameters are passed in the registers if they are not bigger than the size of 2 registers. [Godbolted](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:2,lang:c%2B%2B,selection:(endColumn:137,endLineNumber:34,positionColumn:137,positionLineNumber:34,selectionStartColumn:137,selectionStartLineNumber:34,startColumn:137,startLineNumber:34),source:'%23include+%3Ccstdint%3E%0A%0Astruct+A+%7B%0A++++int64_t+a%3B%0A++++int64_t+b%3B%0A%7D%3B%0A%0Astruct+B+%7B%0A++++int64_t+a%3B%0A++++int64_t+b%3B%0A++++int64_t+c%3B%0A%7D%3B%0A%0Avoid+foo(int64_t+i)%3B%0A%0Avoid+foo(A+a)%3B%0A%0Avoid+foo(B+b)%3B%0A%0Avoid+foo2(B%26+b)%3B%0A%0A//+int64_t+is+passed+in+registers+for+sure%0Avoid+barI()+%7B%0A++++int64_t+i%3B%0A++++foo(i)%3B%0A%7D%0A%0A//+Without+knowing+the+binary+code,+you+can+see+that+A+is+copied+in+the+same+way+as+a+value+of+int64_t+%0Avoid+barA()+%7B%0A++++A+a%3B%0A++++foo(a)%3B%0A%7D%0A%0A//+But+when+a+function+called+with+B+passed+by+value,+you+can+see+the+function+call+itself+requires+more+preparation+in+the+machine+code%0Avoid+barB()+%7B%0A++++B+b%3B%0A++++foo(b)%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%232',t:'0')),k:49.66057441253264,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((g:!((h:compiler,i:(compiler:clang_trunk,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'0',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:2,lang:c%2B%2B,libs:!(),options:'-std%3Dc%2B%2B17+-O1+-DNDEBUG',selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:2,tree:'1'),l:'5',n:'0',o:'x86-64+clang+(trunk)+(C%2B%2B,+Editor+%232,+Compiler+%232)',t:'0')),header:(),l:'4',m:75.30864197530865,n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compiler:2,editor:2,fontScale:14,fontUsePx:'0',wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+clang+(trunk)+(Compiler+%232)',t:'0')),k:48.205332011952024,l:'4',m:24.691358024691358,n:'0',o:'',s:0,t:'0')),k:50.33942558746736,l:'3',n:'0',o:'',t:'0')),l:'2',n:'0',o:'',t:'0')),version:4) (you can try to change the optimization flags from `O1`  to `O2` or `O3`.

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
antaljanosbenjamin: First of all, I missed that `node_label` is overwritten, so it won't be -1 always. Therefore this comment of mine is wrong, sorry for that.

Second, the overflow only happens if the unsigned value is greater than the maximum value of the signed value, [check here](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:2,lang:c%2B%2B,selection:(endColumn:53,endLineNumber:9,positionColumn:53,positionLineNumber:9,selectionStartColumn:53,selectionStartLineNumber:9,startColumn:53,startLineNumber:9),source:'%23include+%3Ciostream%3E%0A%23include+%3Climits%3E%0A+%0Aint+main()+%7B%0A++++std::int64_t+x+%3D+2%3B%0A++++std::uint64_t+y+%3D+std::numeric_limits%3Cstd::uint64_t%3E::max()%3B%0A++++std::int64_t+z+%3D+(std::int64_t)y%3B%0A++++%0A++++std::cout+%3C%3C+x+%3C%3C+!'%5Cn!'+%3C%3C+y+%3C%3C+!'%5Cn!'+%3C%3C+z+%3C%3C+!'%5Cn!'%3B%0A+%0A++++return+0%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%232',t:'0')),k:49.66057441253264,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:clang_trunk,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'0',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:2,lang:c%2B%2B,libs:!(),options:'-std%3Dc%2B%2B17+-O1+-DNDEBUG',selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:2,tree:'1'),l:'5',n:'0',o:'x86-64+clang+(trunk)+(C%2B%2B,+Editor+%232,+Compiler+%232)',t:'0'),(h:output,i:(compiler:2,editor:2,fontScale:14,fontUsePx:'0',wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+clang+(trunk)+(Compiler+%232)',t:'0')),header:(),k:50.33942558746736,l:'4',m:100,n:'0',o:'',s:1,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4).

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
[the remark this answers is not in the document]
antonio2368: A quick follow-up question do you think a solution with a smaller complexity (using std::unordered_set) would work better here? :smile:

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
antaljanosbenjamin: @antepusic asked me for an explanation for this in private, so I will leave [this](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYgATKVpMGoZAFI9AIQuXSS%2BsgJ4BlRugDCqWgFcWDEJIA7KRuADJ4DJgAcr4ARpjEElwArKQADqgKhM4Mnj5%2BAcEZWU4C4ZExLPGJXCl2mA6lDEIETMQEeb7%2BQfWNOS1tBOXRcQlJqQqt7Z0FPZODw5XV4wCUdqjexMjsHBYAzBHIPlgA1GZ77s6TxJis59hmGgCCj08A9G8nACoIeAonAHc6LQTvETkwTgA3BIAT1BxDwmCoJwiBGIGG8jhyJ1QyJYqGhJyULGMTmQClIoMwom8ShArw%2BJxOXAAdN8EAlMOCbicWJiEFSCAQEidMKo0kZmE1/hsCCcCByeQrjCcAJInNDeWjoUXQ4iAhHC0gMz5MpmYFnAFkgE4IIVpBQgD7zZAAawJCSotFQAJZaBYbyYbz2XA0AE4wwAOLhvABssb2EcjYZNTL0bPV6AEYA4coBpPlqANhC5ClQbAVEWAJyzAIYKPrWFadH%2BeGRZbYsQwcIaSnBtBuTHQMNTZpreB1hBZqcs3jl6tYtqYaTScKIJ2AeEJEIUCBIctR6PQmKa8oQTAP/yYsTLPmFJ0i2wUCjaa9QqeMCgBIoBCDhaSYZ8qzPLlvAYPAAEdvEwAB9NI0VBOEzFjDRJnQEAQDAyDoLgtFzncABxYhlwQfD7nIvRYwsWMUX%2BeIhQSVMIjPP5RVUJhHCJQhvAvHJp2eRkfi5fFCWJUk8HJUVmFieh/nXSZbhBBUuRuMtNifHFkRVVBlP1VBYm0akCDZABZJgYXiVNlN5D0NQEa4TxIDUVTBKhWDoPA2kLE4YQ2elnleBzOKEM5AmsZ4xyECAVhOG1UVw4gzDCyIAQbAgIA0FYkoAEVCywiQIdCUFlM4Lnws49D0LA3K1OVHBIMxkncBgLD0c58py15IogNAGEmE4hGonFdJiuLBASpLLBStKIAAKh0xUWXi%2BDiCywJcsmrqxzNNCMM1OUyIuCq9DQVcNSIRKmpayr2q2plOois0osGyihoWhJRrSiawve4glvGlacryu7tt%2B/6CAS0rcoYLVaEBvZwqebadsKvaSsO9xjpErl6su5rWtux77vWkGAD8opizaie2tsTggGHaDhhDzmylmvpWymwpB5GmV24q51K9xyta9ALsBQg9wFrB6GFRr8ZuhHXm5s0gd7LkqaRnmxz5/bBeFyrRacoEFRrBpMFlq6CcV6mtel82uWWvDrc15GHpd0LWYC6muy8E4AAlAIANTECdlDRaK7L6g6wpOG4CE2esGaZ/UwDANnHcShGPZB1ETgI83g58TAI96/rJtj82E5OWaM/a7PqchVAJwGguxGgiBc8hNvME5/Ka4B5m9lyrui7rt27rSBEu%2BFfz3dRWb2adxHOud1N1Urf4jkAuSOS44ApU2LlcRswlcdogaIBeqjKJWNkoh9E4ZZzOSxFdcEuwF38L1TaihrYT9U5qjlP/Pq4J6zEBHtBCuVBOQMG2GyAA6hLEqwAm4mFNpgNIaUizWUgZgVMohhSoIRJgVs9YhZWBsGcZIlguCNVZjQtqyRcpPw4P8F8a5d4QO7hXSepDGAECvAwHU%2BlDKcWVAQVMqkNhbFIc5esYJJheEYIhYaiocQGSMmyISsc8HQNgRpNowlMgEFoCOASnxkCDmFDqWIcJaTAT5tjahlg9j0MpL%2BCSApWLaFpHKPUcIFAknaM5fq65uGjwsfom4cCuTrgiPgQhcTzxym0poxwz8K5qVkZvFypYiD0AYPxF4USnhUGFPqDeHicYqmceuGRkcrH2zYqwCUFolbPEbs3IQeAWBtIAEpIkvkNBQvcQaCV%2BK2f4X9LxLhXIwYCDAfTaMmefIgrRGZwhHs3Ch1grCUliALQgtFxmfHQMYYACQNj/CoOiFgIEiS4gIPmHkWAsj7xOGkBICgvlYmhHffS3Y2K/MEWeb%2B1NGTWWkepORx9kKoXhdyLkqDgLrkUQUlRdiwG/Q0WI4y7J8EQs%2BGiKBxsBQQiwlBWCK1KROUwNCes%2BIbgrIvJk38cJjmsSWXKVBGBCynM%2BdvbkGxhEnEpThFa0TGBPhtLSUsvS2mfLQRUmUiUiVStiTcpy1ksh9PoKE0htL9Ryt0Twm4MCYkaV/Co5SAroWyM0mK8CVLIYsDwMAO0VICrKIUUhSqIlgK3PLK1FExkQZMDnEWb0ohaAwX%2BGzJxHoICjMJi8EmXsnhdJ1P0wu0FBlUGGa9SiRIxnqp%2BKxKgYEsQCEfqgVArp/i%2BP6rQPArp1YoR6XqzA%2Bb4UHKOXKG42wtxyPhQoRFdi7V6PNQYw1D4dLgi4l2jV2w6a0h4ps5yjNMA6loHgu1SIZ233ZKxEB8aUKJuhIi88ckixgjlTqa19YIiN1dFWAV1lK1wKaGyJ4orvxztSqO1GIBsaIoSRJC8ciJEPNEUZZ%2B%2B7skaTdR6vMwIvVKMKSyggmSVRiglBJY5/o3UEDYIIAVQb7nWRjfQfUWQsDFLHBG9c0axBxqhgVIq2Nk1ZVXumkp7xPihHNpkpQXI9ypX/aIJ9DAX3JMwCwVOrxc4kgiBHDWY51M8x1ujMqR1WqdoGUiOW102rOy1iFMdZmtPAd1hjY6CgWQBwUIXUO4cVg2hDXZhzTmXPoDDsQNTunMY5mMyFqzyNtMCzs61dwYgW3oIM/QHtlsFaIy1oyJ4/xfLeHkUSTAYnORzsrOg1iPzqSeVoH2g657x3UnLOrSq8n4IjkquCcpjEbZmkZNR4C8LEvdqMyhBjPN%2Bv5u46mrWkWDpBeOqNpEFd47EEiOgYzVs0vWaKrZmbrVvNBxDn5tzHnWtecc3tltB2AuUxmyFq6YX1vE09u7Kbes9M3UtiltxehPvff2L9r7f2fv/aB4DkHAOwfvflqZ%2B7wNOu8xszpoWr29A5u7slyHE2RtEgxxF%2BHUXtuVV285/b/nopHb0C9zGhPfMk6u4j4Luxbu7HC9tZ70XKqxcZsBFHRc0cmex9tRkftCv/rcGKrBzEdWsHlfHXiNabjADaPgdB65R29tDeLRmVJlZMkZE1ggcImDtf1PC7nebBsaCq4CDk9YOyYC7MOQEpIb0Nlk6G40sOTiMiA5xj0V7ALeTvUoHU65qPq7JWA/sg4Hdij%2BE4Ew2vPefFg5xRqlgmD0LZJYOE/Ati9fPcB0DKEHnUZ/GhsETAXnJIvGcGrKEE%2BMmQ56sEevEQ6kNxUzd8Xqwfqrd%2Bk4v7g%2BrNYgq/cpJ3fuzHIyeINI%2BwOPQWkvF0zfjIAFNeqkKjsY6go%2BfNdcWDd2SwPXz4QSGggmMDqVStUyEak2M4WkoJvDAAcwn03A2C0XuLqMnj0PJ%2BfFf2NlNLORkbKVAUhBgHMOqcsNIOgTAA5akCNPsf9c8QkdcE1b3DCQvDQBPVnfHZHXNN/BbBObdVbVLbA3HabOnezU7Inc7GnMnCnIkag6nNzBgm7ZqO7EGcePjVMTLBdS/WgecFiBgN%2BD9Lwb0IEdBFfC5ORXcDYbUL1EkU4CXXeNALAWIQCQlATCuBXYgJXbvIseFDnZALUSDUIa8XsekOvKJVkE4dwc8EwOTeUGEL5R1YAYiNIAUbfXdM1A9S1LkCAaiGKcJPdKJM0adPwumX%2BG%2BYpRkdMfvdAHUdAkDX3Iva1HkQwsQYwowYUMw%2BIWgMdIvViEvdAGIz4PYH9K8LJWqHyDYNKI8TEbdGvVCAvFIjQbyHFT9atPqSkXSJEEgTQxkM0IcJtOUayACYiCsEUfXFw4%2BBUTILkTo6UMBNvBIpoj/RFdcZSPANVLQs0Z9VAQhPiVMSQNkQZL5SDQfLkPhSEO/f4DIVEf4MCJwEEbLCuDiAUHvL9bENIxwtwkiUcJkIo6xbdLDE4RY7EayP4jw8%2BblCuIdaENvSo8YqXCpKrAE9XIE24GxWtajGESkMsGonLIxMEm4OJIsOpXeZPVDOYucdEqQkwYCYwGEYrasNWBsMFOUTowgAQYpH%2BGhdPZhb4akBAcCXE9XKTfsAksEYo8WE2YUPpEgV8XFIyCkUUS0a0H%2BFCf/IZQaLmdaLKFCOAmfOJJrRU4gOEKkq8HkCJaCOiOEaqCNAQnohAe/CIKyVZUQPsJIzAmE%2BdR8UhF8c03kqJVPOhZhVMO0AgB0J0N4LMckFkN1KxTIJ5P0csN4RgAAWlpDeGQBXBzLzO8KLgUAzPPwzJtNIQzJuIUHXQzNzLSHOAADEbjMAAQWYWAFBIRkAMyuArCtDU8mFcpIzoyPhGA/QVxwjpULR/Q3gAR8y0g3hJRgAeJLk3g8EYIkliFzE%2ByaE3FBShzHQRyik6yJzYlUyAxZy6y3g5wPJ9c3hsYOA1haBOBkheB/AOAtBSBUBOBdkqFEMGs9g9BeACBNAHy1hXQAgwwWR4wNBkhAhJBJAuA9BkgALYxpAnyOBJBXyQLPzOBeBHQLdgL3yHzSA4BYAkB/RoCaNyBKAKKYDEgt4TAYISVhC%2BA6AVVKBYhsLYgIhXxOAeBSAiMBEAB5BgMxbCrAEkEwcQIi0gfAQdJwaER0GSsUakOcHYD81EBobCltWIYic0zwLAbCtEXpPi4ir0JgJ/QOREAEISr5N8/i/gQQEQMQdgKQGQQQRQFQdQGS3QAwRclAGwGwQwPAG8eANYVAeCHIJSjMtCFmcwShKwLgQIE4DMoSrgFK7KKIbKbASwOQAiPCs2Lo1wYRGYfwLgEIYRRYUYGodITIbIAQUqiQWqkoHIKqqoMYcq%2BwIyfoKYDoLwLoJqrqrogYdoNq5YTq3qxqiahYCIEYdqmoNYP8/QUgNEfLUyx858rCmSr8jgVQSMWMDM1CjURcumZi10GKCAH8/ZE4XAbk43L7AwWwqAuiiqAClYICkCtYCAJAWUNIOcainqZ6mjZa7dO65axy4QGNVy6QCGzytQbCgEdw9awwTa0gN8j8naoSucP6uUY%2BPag6o6hi6sCAM6i6zwPpF6/YPQd6laz6tYDkIcMYaKUgcC5IDQFkPYRMSQJC5ISMOC2MQIPYWMFGjC3gFgCQDQC3dG3gHavCpqwirQFYDajgQCtG7C2W2moipW0gPULIFwSQIAA%3D%3D%3D) here also to be sure it it clear enough for everybody.

antonio2368: In this, case pure by value parameter would be enough as `unique_ptr` is a move-only type (you cannot copy it, you would get a compile error). 
This is another case for why `unique_ptr` is a good thing. :smile:

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
antepusic: As label numbers are obtained from node IDs (starting at 0), overflow shouldn’t happen unless we’re dealing with a graph with ~2^63 nodes, and hardware is still a long way from there.

I’ll reassess how IDs work anyway so as to alleviate any possible confusion and future-proof the algorithm.

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
antepusic: Thanks @antaljanosbenjamin & @antonio2368!

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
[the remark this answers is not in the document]
antepusic: Unordered sets are perfectly suited to this, thanks!

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
antaljanosbenjamin: @antonio2368 yes, that's true. The only reason I did say reference, because it is "moved" from one function to another, and it is only used from one place. By using an rvalue reference you can spare that extra move constructor+destructors. This case I would say value parameter has only drawbacks, because the function is never called with an lvalue reference. Though in practice it doesn't matter for performance in my opinion, I think this is the absolute best way to do it.

However if the function would be called with oth lvalue and rvalue references, than the value parameter is the way to go, because it spare and extra overload.

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.hpp
kostasrim: Yes, the query module shall only care that the value does not change (since we pass by value) but the implementation of NodeLabel (in the source file) cares if the that argument is modified on its body. That way you get a clean separation. I hope that helps :)

antepusic: Thanks! I’ll be making this change in line with your comment and Google’s style guide which we try to follow at MAGE.

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
```diff
@@ -0,0 +1,370 @@
...
+///@param b -- set against which the check is performed
+///
+///@return -- whether a is subset of b
+template <class T>
+bool IsSubset(const std::unordered_set<T>& a, const std::unordered_set<T>& b) {
+  return std::includes(a.begin(), a.end(), b.begin(), b.end());
+}
+
+std::vector<std::uint64_t> LabelRankT::NodesMemgraphIDs() const {
+  std::vector<std::uint64_t> nodes;
```
antonio2368, line 24: you know how many nodes you will add so you can call [reserve](https://en.cppreference.com/w/cpp/container/vector/reserve).

cpp/dynamic_community_detection_module/algorithm/dynamic_community_detection.cpp
```diff
@@ -0,0 +1,370 @@
...
+    times_updated[node_id] = 0;
+  }
+}
+
+std::int64_t LabelRankT::NodeLabel(std::uint64_t node_id) {
+  std::int64_t node_label = -1;
+  double max_P = 0;
+
+  for (const auto [label, P] : label_Ps[node_id]) {
+    if (P > max_P || (P == max_P && (std::int64_t)label < node_label)) {
```
antonio2368, line 117: Prefer `static_cast` as it's much safer.

cpp/dynamic_community_detection_module/dynamic_community_detection_module.cpp
```diff
@@ -0,0 +1,329 @@
...
+constexpr char const *kWSelfloop = "w_selfloop";
+constexpr char const *kMaxIterations = "max_iterations";
+constexpr char const *kMaxUpdates = "max_updates";
+
+constexpr char const *kCreatedVertices = "createdVertices";
+constexpr char const *kCreatedEdges = "createdEdges";
+constexpr char const *kUpdatedVertices = "updatedVertices";
+constexpr char const *kUpdatedEdges = "updatedEdges";
+constexpr char const *kDeletedVertices = "deletedVertices";
+constexpr char const *kDeletedEdges = "deletedEdges";
```
antonio2368, line 26: Prefer `std::string_view`.

cpp/dynamic_community_detection_module/dynamic_community_detection_module.cpp
```diff
@@ -0,0 +1,329 @@
...
+
+LabelRankT::LabelRankT algorithm = LabelRankT::LabelRankT();
+bool initialized = false;
+
+auto saved_directedness = false;
+auto saved_weightedness = false;
+std::string saved_weight_property = "weight";
```
antonio2368, line 33: put them into an anonymous namespace so they are static.

cpp/dynamic_community_detection_module/dynamic_community_detection_module.cpp
antepusic: Whereas I understand we prefer `std::string_view` with constant strings, I don't know if that is feasible here. For one, `mgp::proc_add_opt_arg` requires a `const char *` argument, and no suitable conversion function from `const std::string_view` to `const char *` exists.

antepusic: And of course, thanks for all the suggestions!

antepusic: That works :+1:

cpp/dynamic_community_detection_module/dynamic_community_detection_module.cpp
```diff
@@ -0,0 +1,347 @@
...
+    const auto min_value = mgp::value_get_double(mgp::list_at(args, 4));
+    const auto weight_property = mgp::value_get_string(mgp::list_at(args, 5));
+    const auto w_selfloop =
+        weighted ? mgp::value_get_double(mgp::list_at(args, 6)) : 1.0;
+    const auto max_iterations = mgp::value_get_int(mgp::list_at(args, 7));
+    const auto max_updates = mgp::value_get_int(mgp::list_at(args, 8));
+
+    ::saved_directedness = directed;
+    ::saved_weightedness = weighted;
+    ::saved_weight_property = weight_property;
```
antonio2368, line 75:
```suggestion
    saved_directedness = directed;
    saved_weightedness = weighted;
    saved_weight_property = weight_property;
```

antonio2368 (changes requested):
You write a really clean code. Comments are again mostly connected to C++ specifics but overall this is a well-structured code!

cpp/dynamic_community_detection_module/dynamic_community_detection_module.cpp
antepusic: In past discussion some reviewers didn’t find it immediately evident that these variables were global so I’ve elected to use the scope resolution operator. As it doesn’t do any harm, I’ll keep it there for now unless there’s a reason not to use it.

antepusic: In past discussion some reviewers didn’t find it immediately apparent that the left-hand variables were global so I’ve elected to use the scope resolution operator. As it doesn’t do any harm, I would prefer to keep it here for added clarity.

antepusic:
> You write a really clean code. Comments are again mostly connected to C++ specifics but overall this is a well-structured code!

Thanks for the kind words! I’ve incorporated your suggestions into the code and code review has been a very educational experience.

jmatak:
@antonio2368 Requested changes need to be reviewed, and if these are addressed, we have a green light :smile:

antonio2368 (approved):

PR #69, opened 2021-09-20
Approximative streaming PageRank

jmatak:
Pull request for implementation of approximative streaming PageRank based on [Bahmani et. al.](http://snap.stanford.edu/class/cs224w-readings/bahmani10pagerank.pdf).


cpp/pagerank_module/algorithm_online/pagerank.cpp
```diff
@@ -0,0 +1,222 @@
...
+std::vector<std::pair<std::uint64_t, double>> GetPagerank(const mg_graph::GraphView<> &graph) {
+  if (pagerank_online_alg::context.IsEmpty()) {
+    return SetPagerank(graph);
+  }
+  if (IsIncosistent(graph)) {
+    throw std::runtime_error(
+        "Graph has been modified, therefore is incosistent with cached results, please update the Pagerank by calling "
+        "set/reset!");
+  }
+  return CalculatePageRank();
```
Josipmrden, line 194: spaces between lines

e2e/pagerank_online_test/test_online_double_call_set/test.yml
```diff
@@ -0,0 +1,15 @@
+- query: >-
```
Josipmrden, line 1: why are we only testing count?

cpp/pagerank_module/algorithm_online/pagerank.cpp
```diff
@@ -0,0 +1,222 @@
...
+namespace pagerank_online_alg {
+namespace {
+PageRankData context;
+std::uint64_t global_R;
+double global_epsilon;
+
+void NormalizeRank(std::vector<std::pair<std::uint64_t, double>> &rank) {
+  const double sum = std::accumulate(rank.begin(), rank.end(), 0.0, [](auto &sum, auto &p) { return sum + p.second; });
+  for (auto &[node_id, value] : rank) {
+    value /= sum;
```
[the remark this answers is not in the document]
jmatak: Yes, we are changing the value of that double primitive there

cpp/pagerank_module/algorithm_online/pagerank.cpp
```diff
@@ -0,0 +1,222 @@
...
+    // Average length of walk is 1/epsilon
+    if (distr(gen) < epsilon) {
+      break;
+    }
+
+    current_id = graph.GetInnerNodeId(next_id);
+  }
+}
+
+void UpdateCreate(const mg_graph::GraphView<> &graph, const std::pair<std::uint64_t, std::uint64_t> new_edge) {
```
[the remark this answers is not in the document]
jmatak: We are never updating neither graph nor edge, and therefore these objects are `const`

cpp/pagerank_module/algorithm_online/pagerank.cpp
```diff
@@ -0,0 +1,222 @@
...
+      return true;
+    }
+  }
+  return false;
+}
+}  // namespace
+
+std::vector<std::pair<std::uint64_t, double>> SetPagerank(const mg_graph::GraphView<> &graph, const std::uint64_t R,
+                                                          const double epsilon) {
+  pagerank_online_alg::global_R = R;
```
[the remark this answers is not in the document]
jmatak: There is no other way to preserve a global state in Memgraph

e2e/pagerank_online_test/test_online_double_call_set/test.yml
jmatak: Algorithm is stochastic, therefore we are unable to test the real output.

cpp/mg_utility/data_structures/graph_view.hpp
```diff
@@ -70,8 +68,15 @@ template <typename TSize = std::uint64_t> class GraphView {
...
+
+  virtual bool NodeExists(TSize node_id) const = 0;
 };
 
-template <typename TSize> inline GraphView<TSize>::~GraphView() {}
+template <typename TSize>
+inline GraphView<TSize>::~GraphView() {}
 
-} // namespace mg_graph
+}  // namespace mg_graph
```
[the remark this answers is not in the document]
antonio2368: It's from Google Style Guide:
https://google.github.io/styleguide/cppguide.html#Namespaces
They use 2 or 4 spaces between code and comment for almost everything.

cpp/pagerank_module/algorithm_online/pagerank.cpp
```diff
@@ -0,0 +1,222 @@
...
+}
+
+void UpdateCreate(const mg_graph::GraphView<> &graph, const std::pair<std::uint64_t, std::uint64_t> new_edge) {
+  auto [from, to] = new_edge;
+
+  std::unordered_set<std::uint64_t> walk_table_copy(pagerank_online_alg::context.walks_table[from]);
+  for (auto walk_index : walk_table_copy) {
+    auto &walk = pagerank_online_alg::context.walks[walk_index];
+
+    auto position = std::find(walk.begin(), walk.end(), from) + 1;
```
antonio2368, line 70: Is it guaranteed that from is contained inside the walk? 
What if you get vector::end as the result of the find? Adding 1 would give you something that is not equal to vector::end.

cpp/pagerank_module/algorithm_online/pagerank.hpp
```diff
@@ -0,0 +1,154 @@
...
+  ///
+  ///
+ public:
+  void Init() {
+    walks.clear();
+    walks_counter.clear();
+    walks_table.clear();
+
+    gen = new std::mt19937(42);
+    distr = new std::uniform_real_distribution<float>(0.0, 1.0);
```
antonio2368, line 20: When do you delete?
That's why you need to prefer `std::unique_ptr`.

cpp/pagerank_module/algorithm_online/pagerank.hpp
```diff
@@ -0,0 +1,154 @@
...
+  ///@brief Context for storing data for dynamic pagerank
+  ///
+  ///
+ public:
+  void Init() {
+    walks.clear();
+    walks_counter.clear();
+    walks_table.clear();
+
+    gen = new std::mt19937(42);
```
antonio2368, line 19: Why a hardcoded seed and not use random_device?
```
std::random_device rd;
std::mt19937 g(rd())
```

cpp/pagerank_module/algorithm_online/pagerank.hpp
```diff
@@ -0,0 +1,154 @@
...
+  /// Table that keeps the node appearance and walk ID
+  std::unordered_map<std::uint64_t, std::unordered_set<std::uint64_t>> walks_table;
+
+  std::mt19937 *gen;
+  std::uniform_real_distribution<float> *distr;
+};
+
+extern PageRankData context;
+extern std::uint64_t global_R;
+extern double global_epsilon;
```
antonio2368, line 40: Why are they global and not just static inside the `cpp` file. Who else needs to be aware of these variables?

kostasrim (commented):
Good job, minor comments!

cpp/mg_utility/mg_graph.hpp
```diff
@@ -241,23 +241,27 @@ class Graph : public GraphView<TSize> {
...
+  std::uint64_t GetMemgraphNodeId(TSize node_id) const override {
     if (inner_to_memgraph_id_.find(node_id) == inner_to_memgraph_id_.end()) {
       throw mg_exception::InvalidIDException();
     }
-    return inner_to_memgraph_id_[node_id];
+    return inner_to_memgraph_id_.at(node_id);
+  }
+
+  bool NodeExists(TSize memgraph_id) const override {
+    return memgraph_to_inner_id_.find(memgraph_id) != memgraph_to_inner_id_.end();
```
antaljanosbenjamin, line 264:
```suggestion
  bool NodeExists(std::uint64_t memgraph_id) const override {
    return memgraph_to_inner_id_.find(memgraph_id) != memgraph_to_inner_id_.end();
```
As this function expects memgraph ids, I think its parameter should be `std::uint64_t` instead of `TSize`.

cpp/pagerank_module/algorithm_online/pagerank.hpp
antaljanosbenjamin: I think actually it is not necessary. It should be possible initialize them with [default member initialization](https://en.cppreference.com/w/cpp/language/data_members#Member_initialization):
```cpp
class PageRankData {
  /// ....
  std::mt19937 gen{std::random_device{}()};
  std::uniform_real_distribution<float> distr{0.0, 1.0};
};
```

antaljanosbenjamin: Oh yes, I don't think currently they need to be reset in `Init`, because we don't gain any value by doing so. The "randomness" won't change when we reset them.

cpp/mg_utility/mg_generate.hpp
```diff
@@ -28,6 +28,24 @@ std::unique_ptr<mg_graph::Graph<>> BuildGraph(
...
+///
+std::unique_ptr<mg_graph::Graph<>> BuildGraph(
+    const std::vector<std::uint64_t> &nodes, const std::vector<std::pair<std::uint64_t, std::uint64_t>> &edges,
+    const mg_graph::GraphType graph_type = mg_graph::GraphType::kUndirectedGraph) {
+  auto G = std::make_unique<mg_graph::Graph<>>();
+  for (auto node_id : nodes) G->CreateNode(node_id);
+  for (const auto [from, to] : edges) G->CreateEdge(from, to, graph_type);
+
+  return G;
+}
```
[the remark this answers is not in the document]
jmatak: The reason behind having this in a .hpp file was to reduce the dependency of other things in mg_utility. For unit tests, we do not want to depend on Memgraph API. 
I remember having a discussion about this choice but can't remember the reasoning behind it.

cpp/mg_utility/mg_test_utils.hpp
```diff
@@ -90,12 +90,45 @@ double AverageAbsoluteError(const std::vector<T> &result, const std::vector<T> &
...
+  std::sort(correct.begin(), correct.end());
+
+  auto size = result.size();
+  for (std::size_t i = 0; i < size; i++) {
+    auto [result_id, result_value] = result[i];
+    auto [correct_id, correct_value] = correct[i];
+
+    if (result_id != correct_id) return false;
+    if (std::abs(result_value - correct_value) > absolute_error) return false;
+  }
```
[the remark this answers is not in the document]
jmatak: That's true, this needs to change. Also, it doesn't seem like a usable function for other purposes.

cpp/pagerank_module/pagerank_module.cpp
```diff
@@ -35,9 +36,8 @@ void PagerankWrapper(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *resu
 
     auto graph = mg_utility::GetGraphView(memgraph_graph, result, memory, mg_graph::GraphType::kDirectedGraph);
 
-    const auto &graph_edges = graph->Edges();
+    auto graph_edges = graph->Edges();
     std::vector<pagerank_alg::EdgePair> pagerank_edges;
-    pagerank_edges.reserve(graph_edges.size());
```
antaljanosbenjamin, line 40: What is the rational behind this?

I think the old implementation is better, because:
1. it doesn't copy the vector returned by `graph->Edges()`
2. it possibly preallocates enough memory to reduce allocations

Am I missing something?

cpp/pagerank_module/algorithm_online/pagerank.cpp
jmatak: It is guaranteed, but a priori we do not have the information about a position within the walk. That's why I move to the next position after calling a find in a walk.

cpp/pagerank_module/algorithm_online/pagerank.hpp
jmatak: It was a testing mistake, I wanted to make sure that tests are having similar output.

cpp/pagerank_module/algorithm_online/pagerank.cpp
jmatak: This is automatic formatting from clang-format

cpp/pagerank_module/pagerank_online_module.cpp
```diff
@@ -0,0 +1,221 @@
...
+    pagerank_online_alg::Reset();
+    InsertMessageRecord(result, memory, "Pagerank context is reset! Before running again it will run initialization.");
+  } catch (const std::exception &e) {
+    // We must not let any exceptions out of our module.
+    InsertMessageRecord(result, memory, "Reset failed: An exception occurred, please check your module!");
+  }
+}
+
+extern "C" int mgp_init_module(struct mgp_module *module, struct mgp_memory *memory) {
+  // Online approximate PageRank solution
```
antepusic, line 137:
```suggestion
  // Online approximative PageRank solution
```

cpp/pagerank_module/pagerank_online_module.cpp
```diff
@@ -0,0 +1,221 @@
...
+      // Query module output record
+      mgp::proc_add_result(pagerank_proc, kFieldNode, mgp::type_node());
+      mgp::proc_add_result(pagerank_proc, kFieldRank, mgp::type_float());
+
+    } catch (const std::exception &e) {
+      return 1;
+    }
+  }
+
+  // Online approximate PageRank get results
```
antepusic, line 160:
```suggestion
  // Online approximative PageRank get results
```

cpp/pagerank_module/pagerank_online_module.cpp
```diff
@@ -0,0 +1,221 @@
...
+      // Query module output record
+      mgp::proc_add_result(pagerank_proc, kFieldNode, mgp::type_node());
+      mgp::proc_add_result(pagerank_proc, kFieldRank, mgp::type_float());
+
+    } catch (const std::exception &e) {
+      return 1;
+    }
+  }
+
+  // Online approximate PageRank update state
```
antepusic, line 174:
```suggestion
  // Online approximative PageRank update state
```

cpp/pagerank_module/algorithm_online/pagerank.cpp
```diff
@@ -0,0 +1,222 @@
...
+    while (position != walk.end()) {
+      auto node_id = *position;
+      pagerank_online_alg::context.walks_table[node_id].erase(walk_index);
+      pagerank_online_alg::context.walks_counter[node_id]--;
+      position++;
+    }
+    walk.erase(std::find(walk.begin(), walk.end(), from) + 1, walk.end());
```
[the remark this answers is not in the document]
jmatak: It's similar scenario as above, this could be simplified to calculate only once

cpp/pagerank_module/algorithm_online/pagerank.cpp
```diff
@@ -0,0 +1,222 @@
...
+                *pagerank_online_alg::context.distr, *pagerank_online_alg::context.gen);
+  }
+}
+
+void UpdateCreate(const mg_graph::GraphView<> &graph, const std::uint64_t new_vertex) {
+  auto R = pagerank_online_alg::global_R;
+  auto eps = pagerank_online_alg::global_epsilon;
+
+  auto walk_index = pagerank_online_alg::context.walks.size();
+  for (std::uint64_t i = 0; i < R; i++) {
```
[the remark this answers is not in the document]
jmatak: What is the reasoning behind this? Is it just the recommendation?

cpp/pagerank_module/algorithm_online/pagerank.cpp
```diff
@@ -0,0 +1,222 @@
...
+    std::vector<std::uint64_t> walk;
```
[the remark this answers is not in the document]
jmatak: Because I am storing it's reference, not sure how to do it with the temporary vector

cpp/pagerank_module/algorithm_online/pagerank.cpp
```diff
@@ -0,0 +1,222 @@
...
+}
+
+void UpdateDelete(const mg_graph::GraphView<> &graph, const std::pair<std::uint64_t, std::uint64_t> removed_edge) {
+  auto [from, to] = removed_edge;
+
+  std::unordered_set<std::uint64_t> walk_table_copy(pagerank_online_alg::context.walks_table[from]);
+  for (auto walk_index : walk_table_copy) {
+    auto &walk = pagerank_online_alg::context.walks[walk_index];
+
+    auto position = std::find(walk.begin(), walk.end(), from) + 1;
```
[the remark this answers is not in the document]
jmatak: Yes, we are sure that somewhere there we would find `from`, and if it is on the last place, it would point to `walk.end()`

cpp/pagerank_module/algorithm_online/pagerank.cpp
```diff
@@ -0,0 +1,222 @@
...
+  }
+}
+
+void UpdateDelete(const mg_graph::GraphView<> &graph, const std::uint64_t removed_vertex) {
+  pagerank_online_alg::context.walks_table.erase(removed_vertex);
+  pagerank_online_alg::context.walks_counter.erase(removed_vertex);
+}
+
+bool IsIncosistent(const mg_graph::GraphView<> &graph) {
+  for (auto const [node_id] : graph.Nodes()) {
```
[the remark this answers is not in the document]
jmatak: Nice to know! As for the first comment, since `node_id` is primitive, is there a reason to keep const here?

cpp/pagerank_module/algorithm_online/pagerank.hpp
jmatak: At one point I expected `module` to be able to get the context information, however I can change that.

cpp/mg_utility/mg_graph.hpp
jmatak: Correct, this is a nice catch!

cpp/pagerank_module/algorithm_online/pagerank.cpp
```diff
@@ -0,0 +1,231 @@
...
+#include <unordered_map>
+
+#include <mg_graph.hpp>
+#include "pagerank.hpp"
+
+namespace pagerank_online_alg {
+namespace {
+PageRankData context;
+std::uint64_t global_R;
+double global_epsilon;
```
antonio2368, line 11: Isn't this unsafe in multithreaded env?

cpp/pagerank_module/algorithm_online/pagerank.cpp
```diff
@@ -0,0 +1,231 @@
...
+  }
+}
+
+std::vector<std::pair<std::uint64_t, double>> CalculatePageRank() {
+  std::vector<std::pair<std::uint64_t, double>> pageranks;
+
+  auto R = pagerank_online_alg::global_R;
+  auto eps = pagerank_online_alg::global_epsilon;
+
+  auto n = pagerank_online_alg::context.walks_counter.size();
```
antonio2368, line 39:
```suggestion
  auto n = pagerank_online_alg::context.walks_counter.size();
  pageranks.reserve(n);
```

cpp/pagerank_module/algorithm_online/pagerank.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+void UpdateCreate(const mg_graph::GraphView<> &graph, std::uint64_t new_vertex) {
+  auto R = pagerank_online_alg::global_R;
+  auto eps = pagerank_online_alg::global_epsilon;
+
+  auto walk_index = pagerank_online_alg::context.walks.size();
+  for (std::uint64_t i = 0; i < R; ++i) {
+    std::vector<std::uint64_t> walk;
+
+    walk.emplace_back(new_vertex);
```
antonio2368, line 105:
```suggestion
    std::vector walk{new_vertex};
```

cpp/pagerank_module/algorithm_online/pagerank.hpp
antonio2368: I would remove this.

cpp/pagerank_module/algorithm_online/pagerank.hpp
```diff
@@ -0,0 +1,146 @@
...
+
+///
+///@brief Deletes vertex from context. This is trivial because we are sure that no edge exists around that node.
+///
+///@param graph Graph for updating
+///@param removed_vertex Removed vertex
+///
+void UpdateDelete(const mg_graph::GraphView<> &graph, std::uint64_t removed_vertex);
+
+}  // namespace
```
antonio2368, line 98: This should be deleted and the docs moved above the implementations as it's not supposed to be exposed to the user.

jmatak: The pattern with anonymous namespaces is to have them only in cpp and not exposed in hpp file?

cpp/pagerank_module/algorithm_online/pagerank.cpp
jmatak: This is correct, and it is a big downside of these dynamic methods. Furthermore, when Memgraph is restarted, this is guaranteed to not work. We already discussed implementing a thread-safe environment for storing dynamic context, however, the cost of engineering would surpass the benefits of having it at this moment.

antonio2368: I would write a comment at least, just to show that you're aware of those problems.

cpp/pagerank_module/algorithm_online/pagerank.hpp
antonio2368: Not only is it a pattern, but it's a huge no-no to put anonymous namespaces in header files: 
https://wiki.sei.cmu.edu/confluence/display/cplusplus/DCL59-CPP.+Do+not+define+an+unnamed+namespace+in+a+header+file
:smile:

cpp/pagerank_module/algorithm_online/pagerank.cpp
jmatak: Done!

antonio2368 (approved):
Good job! Sorry for the delay.

PR #70, opened 2021-09-20
New query modules API

jmatak:
Changing the API and getting ready for the new Memgraph version :smile:


e2e/rsmgp_example_test/test_simple/test.yml
```diff
@@ -1,35 +0,0 @@
-query: >
```
gitbuda, line 1: @jmatak can you rollback this as well + ignore the rust test in the e2e run?

jmatak: I should then include exclusion on e2e tests, which will introduce unnecessary complexity in other parts of the code (for example, testing action). I would leave it as clean as possible

antaljanosbenjamin (approved):
Looks nice to me! I haven't checked everything thoroughly, only `mgp.hpp` is the only file I checked from top to bottom, it looks good. For the others I scrolled through them, tried to find bugs related to lifetime management, but I didn't find anything.

PR #76, opened 2021-10-11
Fix set_cover.greedy result type bug

tonilastre:
When I try to use `set_cover.greedy` the following error is thrown:

> Error: Query failed: set_cover.greedy: ValueError: Unable to insert field 'containing_set' with value: '<mgp.Vertex object at 0x7f83314cafd8>'; did you set the correct field type?

It is because the function tries to set `Vertex` into a result type `List[Vertex]` for key `containing_set`. Along with the fix, `set_cover.greedy` now has the same function call and response as `set_cover.cp_solve`.


jmatak:
@Josipmrden Here is a reminder that this requires attention

PR #89, opened 2021-10-25
Weighted graph methods built for dynamic community detection

antepusic:
Weighted graph methods for MAGE’s dynamic community detection query module (memgraph/mage#66), transcluded into separate PR for easier review per Matak’s suggestion.

### DIFF
--- a/cpp/mg_utility/mg_graph.hpp
+++ b/cpp/mg_utility/mg_graph.hpp
@@ -200,9 +200,8 @@ class Graph : public GraphView<TSize> {
 
   /// Creates a weighted edge.
   ///
-  /// Creates an directed/undirected edge in the graph depending on graph type.
-  /// Edge will contain information about the original node IDs. Throws an
-  /// exception if nodes are not contained in graph.
+  /// Creates a directed/undirected edge in the graph depending on graph type. Edge will contain information
+  /// about the original node IDs. Throws an exception if nodes are not contained in the graph.
   ///
   /// @param[in]  from  The from node identifier
   /// @param[in]  to    The to node identifier
--- a/cpp/mg_utility/mg_graph.hpp
+++ b/cpp/mg_utility/mg_graph.hpp
@@ -22,10 +22,9 @@ class Graph : public GraphView<TSize> {
   using typename GraphView<TSize>::TEdge;
   using typename GraphView<TSize>::TNeighbour;
 
-  static_assert(
-      std::is_unsigned_v<TSize>,
-      "mg_graph::Graph expects the type to be an unsigned integer type\n"
-      "only (uint8_t, uint16_t, uint32_t, or uint64_t).");
+  static_assert(std::is_unsigned_v<TSize>,
+                "mg_graph::Graph expects the type to be an unsigned integer type\n"
+                "only (uint8_t, uint16_t, uint32_t, or uint64_t).");
 
  public:
   /// Create object Graph
@@ -55,11 +54,9 @@ class Graph : public GraphView<TSize> {
   /// @param[in] second node id
   ///
   /// @return     Iterator range
-  std::vector<TSize> GetEdgesBetweenNodes(TSize first,
-                                          TSize second) const override {
+  std::vector<TSize> GetEdgesBetweenNodes(TSize first, TSize second) const override {
     std::vector<TSize> edge_ids;
-    const auto [range_start, range_end] =
-        nodes_to_edge_.equal_range(std::minmax(first, second));
+    const auto [range_start, range_end] = nodes_to_edge_.equal_range(std::minmax(first, second));
     edge_ids.reserve(std::distance(range_start, range_end));
     for (auto it = range_start; it != range_end; ++it) {
       if (IsEdgeValid(it->second)) {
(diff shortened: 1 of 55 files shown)


cpp/mg_utility/mg_graph.hpp
```diff
@@ -149,13 +184,33 @@ class Graph : public GraphView<TSize> {
     neighbours_[from].emplace_back(to, id);
     nodes_to_edge_.insert({std::minmax(from, to), id});
 
+    directedness = graph_type;
```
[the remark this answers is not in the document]
antepusic: Fixed!

jmatak (approved):

cpp/mg_utility/mg_graph.hpp
```diff
@@ -90,6 +90,28 @@ class Graph : public GraphView<TSize> {
...
+  ///
+  /// @param[in] node_id target node id
+  ///
+  /// @return vector of neighbours
+  const std::vector<TNeighbour> &InNeighbours(TSize node_id) const {
+    if (node_id < 0 && node_id >= nodes_.size()) {
+      throw mg_exception::InvalidIDException();
+    }
+
+    return in_neighbours_[node_id];
```
jbajic, line 112: Why the above method doesn't have the same condition? Maybe just check if there is node_id in `in_neighbours_`?

cpp/mg_utility/mg_graph.hpp
```diff
@@ -148,14 +183,32 @@ class Graph : public GraphView<TSize> {
...
+      in_neighbours_[to].emplace_back(from, id);
     }
+
     return id;
   }
 
+  /// Creates a weighted edge.
+  ///
+  /// Creates an directed/undirected edge in the graph depending on graph type. Edge will contain information
+  /// about the original node IDs. Throws an exception if nodes are not contained in graph.
```
jbajic, line 200:
```suggestion
  /// Creates a directed/undirected edge in the graph depending on graph type. Edge will contain information
  /// about the original node IDs. Throws an exception if nodes are not contained in the graph.
```

cpp/mg_utility/mg_graph.hpp
```diff
@@ -279,6 +333,9 @@ class Graph : public GraphView<TSize> {
 
   std::vector<std::vector<TSize>> adj_list_;
   std::vector<std::vector<TNeighbour>> neighbours_;
```
jbajic, line 335: Do neighbours_ contains both in and out neighbours or just out? If so make it explicit

cpp/mg_utility/mg_utils.hpp
```diff
@@ -80,13 +103,14 @@ class OnScopeExit {
 };
 
 ///@brief Method for mapping the Memgraph in-memory graph into user-based graph
-/// with iterative node indices. Graph object stores information about node and
-/// edge mappings, alongside with connection information. Created ghraph is
-/// zero-indexed, meaning that indices start with index 0.
+/// with iterative node indices. The graph object stores information about node
+/// and edge mappings alongside with connection information. Created graph is
+/// zero-indexed, i.e. indices start at index 0.
```
jbajic, line 108:
```suggestion
/// with iterative node indices. The graph object stores information about the node
/// and edge mappings alongside connection information. Created graph is
/// zero-indexed, i.e. indices start at index 0.
```

cpp/mg_utility/mg_graph.hpp
antepusic: Were you asking about `OutNeighbours()` not having the same condition? If so, that function solely calls `Neighbours()`, which does have the condition you’ve mentioned.

The check is handled by the `node_id >= nodes_.size()` condition since `in_neighbours_` contains data for each node in ascending order by `node_id`. This design is in line with how the contributors before me implemented `neighbours_`.

cpp/mg_utility/mg_graph.hpp
antepusic: Due to the way `neighbours_` is initialized, with directed graphs it only contains the out-neighbors.
Keeping in mind that in- and out-neighbors are identical for undirected graphs, I’m reluctant to rename the vector, but I’ll clarify this with a comment. Does this sound good to you?

jbajic: Yes, that sound ok for now. But I would definitely think about separating different versions of Graph, since now you have directed/undirected/weighted in the same class, and that causes undirected to have properties it doesn't use. What do you think about this @jmatak ?

antepusic: If we decide to separate them, undirected graphs would still have to implement certain interfaces because of generality. For example, if an algorithm requires the out-neighbors, we should be able to call `OutNeighbours` without having to check whether the graph is directed. Of course, for undirected graphs, `OutNeighbours` would just default to calling `Neighbours`. 

Does the above make sense? I believe we should definitely take it into account if it’s a valid suggestion.

jbajic: aha, so we are speaking in the context of algorithms running on both undirected and undirected graphs (sorry for the late response)?

antepusic: Right @jbajic!

PR #98, opened 2021-11-16
Add json_util module

katarinasupe:
### Description

Added load_from_path procedure for loading JSON from local file and load_from_url for loading JSON from remote address procedure. Also created one e2e test for load_from_url and updated README.md by adding json_util.py as Python module.

### Pull request type

- [x] Algorithm/Module

#### Module/Algorithm
- [x] Unit tests
- [x] End-to-end tests
- [x] Code documentation


python/json_util.py
```diff
@@ -0,0 +1,41 @@
...
+def load_from_path(ctx: mgp.ProcCtx, path: str) -> mgp.Record(objects=mgp.List[object]):
+    """
+    Procedure to load JSON from a local file. 
+
+    Parameters
+    ----------
+    path : str
+        Path to the JSON that is being loaded.
+    """
+    with open(path) as json_file:
```
jmatak, line 16: Check whether the path exists before opening. Also, I would recommend using `pathlib`, it comes from standard Python.
`from pathlib import Path`

python/json_util.py
```diff
@@ -0,0 +1,41 @@
...
+        objects = json.load(json_file)
+
+        if type(objects) is dict:
```
jmatak, line 19: What happens if objects are of type List?

python/json_util.py
```diff
@@ -0,0 +1,41 @@
...
+def load_from_url(ctx: mgp.ProcCtx, url: str) -> mgp.Record(objects=mgp.List[object]):
+    """
+    Procedure to load JSON from a remote address. 
+
+    Parameters
+    ----------
+    path : str
+        URL to the JSON that is being loaded.
+    """
+    with urllib.request.urlopen(url) as json_url:
```
jmatak, line 35: Since this is repeatable to an upper method, maybe the extraction of data could be exported to another function and reused in both query procedures.

jmatak (changes requested):

python/json_util.py
katarinasupe: Added that in the new commit, not sure if that's what you were thinking about.

python/json_util.py
katarinasupe: It's fine, they are list and I can easily UNWIND them in Cypher. Tested that on simple ["apple", "banana"] example, since JSON can be array or dict. Dict cannot stay dict, since return type is List and since I want to UNWIND that.

python/json_util.py
katarinasupe: Fixed that by adding the extract_objects() method and realizing that my code previously could have been much simpler. Thanks :)

e2e/json_util_test/test_load_from_url/test.yml
```diff
@@ -0,0 +1,9 @@
+query: >
+    CALL json_util.load_from_url("https://raw.githubusercontent.com/memgraph/mage/add-json-util/e2e/json_util_test/test_load_from_url/data.json")
```
jmatak, line 2: Is there a possibility of sharing it on some MG cloud rather than using GitHub as storage

python/json_util.py
```diff
@@ -0,0 +1,54 @@
...
+    ----------
+    url : str
+        URL to the JSON that is being loaded.
+    """
+    try:
+        response = urlopen(url)
+    except URLError:
+        print("There is no file " + url)
+    else:
+        objects = extract_objects(response)
```
jmatak, line 52: Haven't seen the usage of try-except-else yet, nice :smile:

e2e/json_util_test/test_load_from_url/test.yml
katarinasupe: Did that, next step is debugging the problem with opening that kind of file with the current request method. Will do asap :)

katarinasupe: The problem I had with making requests was the missing requests due to amazon external link I added https://download.memgraph.com/asset/mage/data.json. Then I figured that could happen with other kinds of links so I added header either way and now it works with raw github link, as well as the link that requires header.

python/json_util.py
```diff
@@ -0,0 +1,56 @@
+import mgp
+import json
+from urllib.request import Request, urlopen
+from urllib.error import URLError
+from pathlib import Path
+
+
+def extract_objects(file):
```
jmatak, line 8: Add type of the `file` variable

python/json_util.py
```diff
@@ -0,0 +1,56 @@
...
+    ----------
+    url : str
+        URL to the JSON that is being loaded.
+    """
+    request = Request(url)
+    request.add_header('User-Agent', 'MAGE module')
+    try:
+        content = urlopen(request)
+    except URLError:
+        print("There is no file " + url)
```
jmatak, line 52: We should raise an exception rather than print and ignore

e2e/json_util_test/test_load_from_url/data.json
```diff
@@ -0,0 +1,9 @@
+{
+    "name": "Michael",
+    "age": 41,
+    "children": [
+        "Selina",
+        "Rana",
+        "Selma"
+    ]
+}
```
jmatak, line 9: Can this be deleted now that you have it on MG download page

jmatak (changes requested):

python/json_util.py
jmatak: That's fine as long as it is written in the documentation

e2e/json_util_test/test_load_from_url/test.yml
jmatak: Nice!

jmatak:
Once more comment. Please use Black formatter when structuring Python code: https://black.readthedocs.io/en/stable/

e2e/json_util_test/test_load_from_url/data.json
katarinasupe: Yes, you're right, I'll delete it

python/json_util.py
katarinasupe: I added it, not sure if that's what you wanted, but I tested the wrong URL, and get
`urllib.error.HTTPError: HTTP Error 404: Not Found`
so I hope that's fine

katarinasupe:
Using Black formatter from now :smile:

python/json_util.py
jmatak: We should catch any kind of error that can occur there, raising it again will result with a trace written in console log

jmatak (approved):

PR #100, opened 2021-12-02
Add export/import util

katarinasupe:
### Description

Addition of import and export util. It adds the ability to import and export graphs in a specific formats.

### Pull request type

- [x] Algorithm/Module

#### Module/Algorithm
- [x] Unit tests
- [x] End-to-end tests
- [x] Code documentation


katarinasupe:
I have a few comments/concerns:

1. Theoretically speaking, I supposed it is enough to count `out_edges` of every vertex to get all edges? When I thought about it, I concluded that if a vertex has `in_edge`, then that edge is also an `out_edge` of another vertex.
2. I noticed an issue when exporting to a local `JSON` file. I need to change permissions locally in order for this to work. Not sure if that could be handled better.
3.  I still didn't write any tests since I'm not sure how to write them since the return type is `None` (an empty `mgp.Record`).
4. I know putting everything in the dictionary might seem silly, but `mgp.Vertex.properties` is not `JSON` serializable, as well as some other types, so I had to go this way until there is some kind of `toJSON()` method in `mgp`.

jmatak:
1. I think this is absolutely alright thing to do
2. Permissions and how do they work with `memgraph` user is something that @antonio2368 is more familiar with. It definitely has to do something with it
3. You won't be able to write e2e tests, but a working unit test (checking only the functionality of methods in a module) would work just fine. But it would require moving the functions to the `python/mage` directory
4. This would be an interesting suggestion for `mgp` module. I don't think it should contain `JSON`, but a transformation to Python `Dict` would be helpful

katarinasupe:
1. Yaay!
2. Okay, I will ask Antonio :) 
3. I'll try writing the unit test and will let you know when it's done
4. Will make an issue on memgraph repository and see how it goes

katarinasupe:
Asked Antonio about permissions, and the possible solution is:
Create folder "folder_name" which will be the folder for exported files. With 
`sudo chown memgraph "folder_name"`
give permission to the user _memgraph_, since that user has started the process.
After that, memgraph will be able to write to the files within that folder.
I tried that and it works, it just has to be described correctly in the documentation.

katarinasupe:
I added test here, and I don't see that I should do any error handling here, since I am only communicating with memgraph and writing to file. Everything necessary for the output file is defined in the docs, mentioned above.

python/export_util.py
```diff
@@ -0,0 +1,56 @@
...
+
+    for vertex in ctx.graph.vertices:
+        labels = [label.name for label in vertex.labels]
+        properties = dict()
+
+        for key in vertex.properties.keys():
+            properties[key] = vertex.properties.get(key)
+
+        node = {
+            "id": vertex.id,
```
jmatak, line 28: These seem like a constant, please write them like this example: https://github.com/memgraph/mage/blob/main/python/mage/graph_coloring_module/parameters.py

python/export_util.py
```diff
@@ -0,0 +1,56 @@
...
+    path : str
+        Path to the file where JSON will be saved.
+    """
+    nodes = list()
+    relationships = list()
+    graph = list()
+
+    for vertex in ctx.graph.vertices:
+        labels = [label.name for label in vertex.labels]
+        properties = dict()
```
jmatak, line 22: `properties = {key: vertex.properties.get(key) for key in vertex.properties.keys() }`

Just the proposal

python/export_util.py
```diff
@@ -0,0 +1,56 @@
...
+        nodes.append(node)
+
+        for edge in vertex.out_edges:
+            properties = dict()
+
+            for key in edge.properties.keys():
+                properties[key] = edge.properties.get(key)
+
+            relationship = {
+                "id": edge.id,
```
jmatak, line 42: Similar comments as above

python/export_util.py
katarinasupe: I like the proposal so I have changed it :)

python/export_util.py
katarinasupe: Currently, I have issues with this kind of enums, due to the sort_key=True when dumping to JSON. I'll check what I can do. Meanwhile, I have decided to create dataclasses for Node and Relationship, it looks nice. We can discuss it on today's sync :D

katarinasupe: This is solved with value :)

katarinasupe:
I think I added everything here :smile:

python/export_util.py
```diff
@@ -0,0 +1,84 @@
...
+                    edge.type.name,
+                    properties,
+                    edge.from_vertex.id,
+                ).get_dict()
+            )
+
+        graph = nodes + relationships
+
+    with open(path, "w") as outfile:
+        js.dump(graph, outfile, indent=4, default=str)
```
jmatak, line 82: Maybe it would be good to wrap this into try/catch because this can fail if the user has no permission to write in some folder

python/import_util.py
```diff
@@ -0,0 +1,93 @@
...
+def json(ctx: mgp.ProcCtx, path: str) -> mgp.Record():
+    """
+    Procedure to import the local JSON created by the export_util.json procedure.
+
+    Parameters
+    ----------
+    path : str
+        Path to the JSON that is being imported.
+    """
+    file = open(path)
```
jmatak, line 39: Use try/catch and use ` with` statement for opening, because file handler might remain open

python/import_util.py
```diff
@@ -0,0 +1,93 @@
...
+import mgp
+import json as js
+
+
+def create_vertex(ctx, properties, labels):
+    vertex = ctx.graph.create_vertex()
+    vertex_properties = vertex.properties
+
+    for key, value in properties.items():
+        vertex_properties[key] = value
```
jmatak, line 11: Does this mutate the object?

e2e/export_import_deprecated/test_online_export_import/test.yml
```diff
@@ -0,0 +1,12 @@
...
+  output: 
+    - count_after_export: 3
+- query: >-
+    MATCH (n) RETURN count(n) AS count;
+  output: 
+    - count: 0
+- query: >-
+    MATCH (n) RETURN count(n) AS count_after_import;
+  output:
+    - count_after_import: 3
```
jmatak, line 12: Sorry to bother, but could you write another test with other data types included. For example, having time data, null, list, or dict. Just to see if it works with complex datatypes

jmatak (changes requested):
These are the last changes, I swear

python/import_util.py
katarinasupe: I was following the [documentation](https://memgraph.com/docs/memgraph/2.1.1/reference-guide/query-modules/api/python-api#example-usage-1) and did this in the same way. I supposed that it changes the object inside Memgraph, and while testing it, I realized it does. Let me know if I understood your question well.

e2e/export_import_deprecated/test_online_export_import/test.yml
katarinasupe: It's great that you asked this, since I totally forgot about temporal types. I need to work on this a bit more.

katarinasupe: Info on my discovery so far:
- I think that there is no problem with lists and I added test to support that.
- I think that `null` is okay. If it's added as a property in the import_file, then it won't be created in the database, just like when running a Cypher query with a property set to `null`. This means that later on it won't be exported with export_util, but then I think that we don't have to even bother with null types? (since we are expecting that the user imports only file from the export module, as said in the documentation).
- I added a test for the dictionary as a part of the property, and that should also work fine.
- Temporal types are currently an issue and I will continue working on them next week.

python/import_util.py
jmatak: Yes, thanks for clearing that one

e2e/export_import_deprecated/test_online_export_import/test.yml
jmatak: Just drop a message after investigation on temporal types :smile:

python/import_util.py
```diff
@@ -0,0 +1,120 @@
...
+                microseconds=parsed_time.microsecond,
+            )
+        if str.startswith(property, Parameter.LOCALTIME.value):
+            return time.fromisoformat(property.split("(")[-1].split(")")[0])
+        if str.startswith(property, Parameter.LOCALDATETIME.value):
+            return datetime.fromisoformat(property.split("(")[-1].split(")")[0])
+        if str.startswith(property, Parameter.DATE.value):
+            return date.fromisoformat(property.split("(")[-1].split(")")[0])
+
+    return property
```
antoniofilipovic, line 26: I have a suggestion to style this differently, since you are actually doing some logic on an indent of 3 or more, it makes sense to first return property if it is not a type of string and then do other things if it is a string.
```suggestion
    if not isinstance(property, str):
      return property
      
    if str.startswith(property, Parameter.DURATION.value):
            parsed_time = datetime.strptime(
                property.split("(")[-1].split(")")[0], "%H:%M:%S.%f"
            )
            return timedelta(
                hours=parsed_time.hour,
                minutes=parsed_time.minute,
                seconds=parsed_time.second,
                microseconds=parsed_time.microsecond,
            )
      if str.startswith(property, Parameter.LOCALTIME.value):
          return time.fromisoformat(property.split("(")[-1].split(")")[0])
      if str.startswith(property, Parameter.LOCALDATETIME.value):
          return datetime.fromisoformat(property.split("(")[-1].split(")")[0])
      if str.startswith(property, Parameter.DATE.value):
          return date.fromisoformat(property.split("(")[-1].split(")")[0])

```

antoniofilipovic: It is complicated to style this with Github actions but you see a point I hope 😄

antoniofilipovic: Also, I think you can isolate this part in variable:
`smth = property.split("(")[-1].split(")")[0]`

katarinasupe: Yep, totally makes sense. I added this part with string check later, so I missed this. Thank you! 
Also, isolation of `property.split("(")[-1].split(")")[0]` is great too. Thanks!

python/import_util.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def create_edge(ctx, properties, start_node_id, end_node_id, type, vertex_ids):
+    vertex_from = ctx.graph.get_vertex_by_id(vertex_ids[start_node_id])
+    vertex_to = ctx.graph.get_vertex_by_id(vertex_ids[end_node_id])
+    edge = ctx.graph.create_edge(vertex_from, vertex_to, mgp.EdgeType(type))
+    edge_properties = edge.properties
+
+    for key, value in properties.items():
+        edge_properties[key] = convert_from_isoformat(value)
```
antoniofilipovic, line 49: @jmatak I don't know, but do we want to have types also here on all functions? Or is it more module-dependent, some modules will have types, some won't?

g-despot (approved):

python/import_util.py
katarinasupe: Is the point of `property.split("(")[-1].split(")")[0]` isolation just readability?

antoniofilipovic: Yes, just for readability 😄

katarinasupe: I changed those two, so feel free to check it and let me know if I did it okay :D

katarinasupe:
Thank you @g-despot, totally forgot about that!

README.md
```diff
@@ -208,25 +208,27 @@ CALL pagerank.get()
...
-| [json_util](python/json_util.py)                                                                   | Python | A module for loading JSON from a local file or remote address.
-| [max_flow](python/max_flow.py)                                                                     | Python | An algorithm for calculating maximum flow through a graph using capacity scaling                                                                             |
+| [graph_coloring](python/graph_coloring.py)                                                         | Python | An algorithm for assigning labels to the graph elements subject to certain constraints. In this form, it is a way of coloring the graph vertices such that no two adjacent vertices are of the same color.                        |
+| [import_util](python/import_util.py)                                                               | Python | A module for importing data from different formats (JSON).                                                                                                                                                                        |
```
jmatak, line 229: Please update this here so we can differentiate between import_module and json_module, there is a clear difference in expected format

e2e/export_import_deprecated/test_online_export_import/input.cyp
```diff
@@ -0,0 +1,23 @@
...
+    - |-
+        MATCH (n) RETURN count(n);
+    - |-
+        MATCH (n) RETURN count(n);
+    - |-
+        MATCH (n) RETURN count(n);
+    - |-
+        MATCH (n) RETURN count(n);
+    - |-
+        MATCH (n) RETURN count(n);
```
jmatak, line 21: Nice workaround :smile:

e2e/export_import_deprecated/test_online_export_import/test.yml
```diff
@@ -0,0 +1,44 @@
...
+    - year: 1999
+      month: 10
+      day: 5
+      hour: 14
+      minute: 15
+- query: >-
+    MATCH (n:User {name: "Jimmy"}) RETURN n.clock.hour AS hour, n.clock.minute AS minute;
+  output:
+    - hour: 9
+      minute: 15
```
jmatak, line 44: Append new line

python/import_util.py
jmatak: If we are talking about type checking in functions, yes!

jmatak (commented):
Address these few small small comments, and add Typing to the functions where it is missing, and Voila, done

python/export_util.py
```diff
@@ -0,0 +1,111 @@
...
+            )
+
+        graph = nodes + relationships
+
+    try:
+        with open(path, "w") as outfile:
+            js.dump(graph, outfile, indent=4, default=str)
+    except PermissionError:
+        raise PermissionError(
+            "You don't have permissions to write into that file. Make sure to give the user memgraph the necessary permissions."
```
antepusic, line 106: Wording: „...give the necessary permissions to user memgraph"

python/mage/export_import_util/parameters.py
```diff
@@ -0,0 +1,17 @@
...
+    TYPE = "type"
+    END = "end"
+    LABEL = "label"
+    START = "start"
+    NODE = "node"
+    RELATIONSHIP = "relationship"
+    DURATION = "duration("
+    LOCALTIME = "localTime("
+    LOCALDATETIME = "localDateTime("
+    DATE = "date("
```
antepusic, line 17: Are the trailing `)`s required? If they're not, consider removing them.

python/export_util.py
```diff
@@ -0,0 +1,111 @@
...
+        return Parameter.DURATION.value + str(property) + ")"
+
+    elif isinstance(property, time):
+        return Parameter.LOCALTIME.value + property.isoformat() + ")"
+
+    elif isinstance(property, datetime):
+        return Parameter.LOCALDATETIME.value + property.isoformat() + ")"
+
+    elif isinstance(property, date):
+        return Parameter.DATE.value + property.isoformat() + ")"
```
antepusic, line 54: I left a comment about removing the `(` from property names below. If you're doing so, perhaps replace the relevant part with formatted strings: `f({property.isoformat()})"`.

python/export_util.py
```diff
@@ -0,0 +1,111 @@
...
+                    properties,
+                    edge.from_vertex.id,
+                ).get_dict()
+            )
+
+        graph = nodes + relationships
+
+    try:
+        with open(path, "w") as outfile:
+            js.dump(graph, outfile, indent=4, default=str)
```
antepusic, line 103: This is really minor but we tend to avoid bare numbers like the `4` here; consider defining `STANDARD_INDENT = 4`.

antepusic (commented):
Solid (and really useful) work!

python/import_util.py
antoniofilipovic: @katarinasupe okay, so just add type checks in functions

python/import_util.py
```diff
@@ -0,0 +1,124 @@
+from datetime import datetime, date, time, timedelta
+import json as js
+from mage.export_import_util.parameters import Parameter
+import mgp
+
+
+def convert_from_isoformat(property):
```
antoniofilipovic, line 7: Don't forget to add types here

python/import_util.py
```diff
@@ -0,0 +1,124 @@
...
+        local_time_iso = property.split("(")[-1].split(")")[0]
+        return time.fromisoformat(local_time_iso)
+    elif str.startswith(property, Parameter.LOCALDATETIME.value):
+        local_datetime_iso = property.split("(")[-1].split(")")[0]
+        return datetime.fromisoformat(local_datetime_iso)
+    elif str.startswith(property, Parameter.DATE.value):
+        date_iso = property.split("(")[-1].split(")")[0]
+        return date.fromisoformat(date_iso)
+    else:
+        return property
```
antoniofilipovic, line 30: Is this leftover? I mean if it is nothing of `Parameter.*.value`, do you need to raise Exception here or return property?

python/import_util.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+def create_vertex(ctx, properties, labels):
```
antoniofilipovic, line 33: Also, add here typings.

python/export_util.py
katarinasupe: I have the `(` on the properties since for example, I can have a property with `date` value and then treat it like a temporal, while it's just a string with value date.

`node Meeting, properties: { type: "date" }`

Let me know if this makes sense to you.

python/mage/export_import_util/parameters.py
katarinasupe: Did you mean `(`? I explained my reasoning in the comment below, but feel free to comment more.

e2e/export_import_deprecated/test_online_export_import/input.cyp
katarinasupe: Desperate times call for desperate measures 😂

python/export_util.py
antepusic: Thanks! Your rationale makes total sense to me, I think I can resolve this point now. :slightly_smiling_face:

python/mage/export_import_util/parameters.py
antepusic: Thanks! Marking this as resolved.

antepusic (approved):

jmatak (approved):

PR #101, opened 2021-12-02
Add import util

katarinasupe:


katarinasupe:
Same thoughts about the test, as in export_module. Seems that file permissions are not a problem here, since I am only reading the file.

python/import_util.py
```diff
@@ -0,0 +1,36 @@
...
+    Parameters
+    ----------
+    path : str
+        Path to the JSON that is being imported.
+    """
+    file = open(path)
+    list = js.load(file)
+    vertex_ids = dict()
+
+    for el in list:
```
jmatak, line 20: When using variables please keep the names easier to understand: `el` -> `element`

python/import_util.py
```diff
@@ -0,0 +1,36 @@
...
+        if el["type"] == "node":
+            vertex = ctx.graph.create_vertex()
+            vertex_properties = vertex.properties
+            for key, value in el["properties"].items():
```
jmatak, line 24: Check whether the key 'properties' exist in the dictionary, if not, throw an exception

jmatak: It applies to other keys here

python/import_util.py
```diff
@@ -0,0 +1,70 @@
...
+    path : str
+        Path to the JSON that is being imported.
+    """
+    file = open(path)
+    graph_objects = js.load(file)
+    vertex_ids = dict()
+
+    for object in graph_objects:
+
+        if all(key in object for key in ("type", "properties", "id")):
```
jmatak, line 21: Constants as the comment in export util

python/import_util.py
```diff
@@ -0,0 +1,70 @@
...
+            vertex = ctx.graph.create_vertex()
+            vertex_properties = vertex.properties
+
+            for key, value in properties_value.items():
+                vertex_properties[key] = value
+
+            for label in labels_value:
+                vertex.add_label(label)
+
+            vertex_ids[id_value] = vertex.id
```
jmatak, line 46: This might be exported to another function since it seems overwhelming here

jmatak (commented):
Overall, after fixing clean code 'Pythonic' mistakes, it should be ready :smile:

PR #102, opened 2021-12-06
Initial MAGE PR template

jmatak:


.github/pull_request_template.md
```diff
@@ -0,0 +1,17 @@
+<!--- TITLE -->
+
+#### Module/Algorithm
+- [ ] Core algorithm/module implementation
+- [ ] [Query module](https://memgraph.com/docs/memgraph/reference-guide/query-modules) implementation
+- [ ] Unit tests
+- [ ] End-to-end tests
+- [ ] Code documentation
+- [ ] README short description
+- [ ] Documentation on [memgraph/docs](https://github.com/memgraph/docs)
```
antoniofilipovic, line 10: Maybe the following parts are not obligatory:
unit tests, end-to-end, code documentation, and documentation on docs.

From discussion we had in August with Mislav I think we just want for now others to contribute.

.github/pull_request_template.md
```diff
@@ -0,0 +1,17 @@
...
+
+#### Bug/Enhacement
+
+<!--- Write the problem description and propose how to solve it -->
+
+- [ ] Code documentation
```
antoniofilipovic, line 16: I also don't know if this is necessary. But on the other hand, we can put to update README.

antoniofilipovic (commented):
I left a few discussion points. It is okay for me to leave it this way.

.github/pull_request_template.md
```diff
@@ -0,0 +1,17 @@
...
+- [ ] Working E2E tests
```
antepusic, line 17: Since E2E tests only apply to query modules, do we need to require them for potential contributions on, say, the API?

.github/pull_request_template.md
antepusic: We want to sync with the contributors, and documentation would definitely help us understand their PRs quicker.
I’m not sure if requiring documentation would demotivate potential contributors as it’s a good practice to document your code anyway.

Agree with the rest of your comment! 👍

jmatak: Maybe we can include these as optional. My reasoning for including this stuff was to never again forget them once I merge the module. I've done it several times now. Do you have an idea of how should we have both clean structures without big requirements and with reminders of important stuff (tests, documentation).

.github/pull_request_template.md
jmatak: That's true since the code is checked by Github Action on each PR push.

jmatak: Therefore, merging is blocked before actions pass.

.github/pull_request_template.md
antepusic: Would it make sense to have two templates?
Let’s give the contributors a basic template called something along the lines of “First PR”, and we’ll continue to use the “big” one.

jmatak: It's common to use only one file, feel free to organize it and propose a new solution for two templates within it

antoniofilipovic: I think this now is fine how you structured it. We want others to contribute, when they make PR we can ask them to add tests or we can add them. So I think I will approve this now.

antoniofilipovic (approved):

.github/pull_request_template.md
```diff
@@ -0,0 +1,32 @@
+### Description
+
+Please explain the changes you made here.
```
antepusic, line 3: Maybe “please briefly explain”?

.github/pull_request_template.md
```diff
@@ -0,0 +1,32 @@
...
+- [ ] Documentation content changes
+- [ ] Other (please describe):
+
+### Related issues
+
+Delete if this PR doesn't resolve any issues. Link the issue if it does.
+
+######################################
+
+### Reviewer checklist (this part checks the reviewer)
```
antepusic, line 22: The reviewer checks this part?

antepusic (changes requested):
Looks good; at one point wording is unclear so I left a brief comment.

.github/pull_request_template.md
jmatak: Yes, I thought of it as a list that shows to the reviewer whether the code is ready to be merged with `main` or not. On the other side, a developer does not need to take care of it (not applies to employees :sweat_smile: )

antepusic: Sorry for not being clear, I was trying to say that the word order was the wrong way around

antepusic (commented):
Thanks for the clarification!

jmatak:
> Thanks for the clarification!

Do I get the Approve then :smile:?

.github/pull_request_template.md
```diff
@@ -0,0 +1,32 @@
...
```
antepusic, line 22:
```suggestion
### Reviewer checklist (the reviewer checks this part)
```

antepusic (changes requested):
We should fix this little bit and then we’re ready to go

antepusic (approved):

PR #112, opened 2021-12-17
Update UUID with stduuid. Work in progress

jmatak:
### Description

Changed default uuid implementation with https://github.com/mariusbancila/stduuid.

### Pull request type

- [x] Build related changes

### DIFF
--- a/cpp/uuid_module/uuid_module.cpp
+++ b/cpp/uuid_module/uuid_module.cpp
@@ -1,4 +1,4 @@
-#include <uuid/uuid.h>
+#include <uuid.h>
 
 #include <mg_exceptions.hpp>
 #include <mg_utils.hpp>
@@ -11,14 +11,13 @@ constexpr char const *kFieldUuid = "uuid";
 
 void Generate(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result, mgp_memory *memory) {
   try {
-    uuid_t id;
-    uuid_generate(id);
-    char *string = new char[100];
-    uuid_unparse(id, string);
+    std::random_device dev;
+    std::mt19937 rng(dev());
+    auto id = uuids::uuid_random_generator{rng}();
+    auto id_str = uuids::to_string(id);
 
     auto *record = mgp::result_new_record(result);
-
-    mg_utility::InsertStringValueResult(record, kFieldUuid, string, memory);
+    mg_utility::InsertStringValueResult(record, kFieldUuid, id_str.c_str(), memory);
   } catch (const std::exception &e) {
     // We must not let any exceptions out of our module.
     mgp::result_set_error_msg(result, e.what());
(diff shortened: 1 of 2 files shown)


jmatak:
@gitbuda Introducing [stduuid](https://github.com/mariusbancila/stduuid) - a third-party software that enables cross-platform work and no need for `uuid-dev` package dependency installation on each device. Do you think this is a good idea to have?

antoniofilipovic:
@gitbuda once again, ping whether we can merge this or close it.

gitbuda:
> @gitbuda once again, ping whether we can merge this or close it.

Sorry for hanging here and thanks for the ping! That repo seems not very robust, could you please explore the https://www.boost.org/doc/libs/1_80_0/libs/uuid/doc/uuid.html (maybe @jmatak already did that)? I think UUID functionality should be there, but let's just invest a bit more time to figure out what would be the best solution and then decide/implement it.

gitbuda:
> > @gitbuda once again, ping whether we can merge this or close it.
> 
> Sorry for hanging here and thanks for the ping! That repo seems not very robust, could you please explore the https://www.boost.org/doc/libs/1_80_0/libs/uuid/doc/uuid.html (maybe @jmatak already did that)? I think UUID functionality should be there, but let's just invest a bit more time to figure out what would be the best solution and then decide/implement it.

Took a bit deeper look at `stduuid` -> doesn't look bad!

antoniofilipovic:
> > > @gitbuda once again, ping whether we can merge this or close it.
> > 
> > 
> > Sorry for hanging here and thanks for the ping! That repo seems not very robust, could you please explore the https://www.boost.org/doc/libs/1_80_0/libs/uuid/doc/uuid.html (maybe @jmatak already did that)? I think UUID functionality should be there, but let's just invest a bit more time to figure out what would be the best solution and then decide/implement it.
> 
> Took a bit deeper look at `stduuid` -> doesn't look bad!

Okay! Let me know of your final decision, if it works we can merge this, or I can investigate the suggested repo further!

cpp/uuid_module/uuid_module.cpp
```diff
@@ -11,14 +11,13 @@ constexpr char const *kFieldUuid = "uuid";
...
-    uuid_unparse(id, string);
+    std::random_device dev;
+    std::mt19937 rng(dev());
+    auto id = uuids::uuid_random_generator{rng}();
+    auto id_str = uuids::to_string(id);
 
     auto *record = mgp::result_new_record(result);
-
-    mg_utility::InsertStringValueResult(record, kFieldUuid, string, memory);
+    mg_utility::InsertStringValueResult(record, kFieldUuid, id_str.c_str(), memory);
```
antoniofilipovic, line 20: I think this is a bug here. 
auto id_str will be a local string variable. 
And 
`mg_utility::InsertStringValueResult(record, kFieldUuid, id_str.c_str(), memory);`
will return a pointer to the local variable which will yield undefined behavior, right? 

@BorisTasevski @Josipmrden

antoniofilipovic (changes requested):

cpp/uuid_module/uuid_module.cpp
antoniofilipovic: Actually no error problem here, since underneath we use mgp::value_make_string function which creates new string from this object which is alive still, so all good.

antoniofilipovic (approved):

PR #117, opened 2022-01-10
Dynamic & static Katz centrality implementation

jmatak:
### Description

Implementation of Katz centrality algorithm which calculates the centrality corresponding to the numbers of paths flowing through the node, and their importance based on the walk length. Both static and dynamic algorithms are approximations but have the property that saves the centrality rankings in the final result. The algorithm is based on: [**Scalable Katz Ranking Computation in Large Static and Dynamic Graphs**](https://arxiv.org/pdf/1807.03847.pdf).

### Pull request type

- [x] Algorithm/Module

#### Module/Algorithm
- [x] Unit tests
- [x] End-to-end tests
- [x] Code documentation


cpp/katz_centrality_module/algorithm/katz.cpp
```diff
@@ -0,0 +1,405 @@
...
+
+  return KatzCentralityLoop(katz_alg::context.active_nodes, graph, katz_alg::alpha, katz_alg::k, katz_alg::epsilon,
+                            gamma);
+}
+
+std::vector<std::pair<std::uint64_t, double>> UpdateKatz(
+    const mg_graph::GraphView<> &graph, const std::vector<std::uint64_t> &new_vertices,
+    const std::vector<std::pair<std::uint64_t, uint64_t>> &new_edges, const std::vector<std::uint64_t> &new_edge_ids,
+    const std::vector<std::uint64_t> &deleted_vertices,
+    const std::vector<std::pair<std::uint64_t, uint64_t>> &deleted_edges) {
```
antepusic, line 317: According to the paper, after a dynamic update one needs to check for convergence and if the algorithm doesn't converge anymore, run iterations of the static algorithm until it does. This step appears to be missing, did you find an alternative?

cpp/katz_centrality_module/algorithm/katz.hpp
```diff
@@ -0,0 +1,131 @@
...
+///
+///@param active_nodes Active nodes - list of nodes in convergence process
+///@param graph Graph for convergence calculation
+///@param alpha Attenuation factor
+///@param k Parameter for keeping the first k node centralities sorted
+///@param epsilon Tolerance factor
+///@param gamma Convergence factor
+///@return std::vector<std::pair<std::uint64_t, double>>
+///
+std::vector<std::pair<std::uint64_t, double>> KatzCentralityLoop(std::set<std::uint64_t> &active_nodes,
```
antepusic, line 45: In the rest of your code, the `alpha`, `k` and `epsilon` parameters of `KatzCentralityLoop` are always `katz_alg::alpha` and so on. With this in mind, is it possible to remove these parameters from the signature as `katz_alg::alpha` etc. are visible from inside the function?

cpp/katz_centrality_module/algorithm/katz.cpp
```diff
@@ -0,0 +1,405 @@
...
+  }
+}
+
+std::vector<std::pair<std::uint64_t, double>> WrapResults() {
+  return std::vector<std::pair<std::uint64_t, double>>(
+      katz_alg::context.centralities[katz_alg::context.iteration].begin(),
+      katz_alg::context.centralities[katz_alg::context.iteration].end());
+}
+
+bool IsIncosistent(const mg_graph::GraphView<> &graph) {
```
antepusic, line 263:
```suggestion
bool IsInconsistent(const mg_graph::GraphView<> &graph) {
```

cpp/katz_centrality_module/algorithm/katz.cpp
```diff
@@ -0,0 +1,405 @@
...
+  // Create context and calculate values if not initialized
+  if (!katz_alg::context.IsInitialized()) {
+    return SetKatz(graph);
+  }
+
+  if (IsIncosistent(graph)) {
+    throw std::runtime_error(
+        "Graph has been modified, therefore is incosistent with cached results, please update the Katz centrality by "
+        "calling set/reset!");
+  }
```
antepusic, line 285:
```suggestion
  if (IsInconsistent(graph)) {
    throw std::runtime_error(
        "Graph has been modified and is thus inconsistent with cached Katz centrality scores; to update them, "
        "please call set/reset!");
  }
```

e2e/katz_online_test/test_online_call_get_and_set/input.cyp
```diff
@@ -0,0 +1,38 @@
+setup: |-
+    CREATE TRIGGER katz_online BEFORE COMMIT EXECUTE CALL katz_centrality_online.update(createdVertices, createdEdges, deletedVertices, deletedEdges) YIELD node, rank SET node.katz = rank;
```
antepusic, line 2: This test only appears to call `.update`; consider adding a different test where KC values aren't saved as node attributes, but retrieved by `.set` and `.get`.

cpp/katz_centrality_module/katz_centrality_online_module.cpp
```diff
@@ -0,0 +1,185 @@
...
+// Each module needs to define mgp_init_module function.
+// Here you can register multiple procedures your module supports.
+extern "C" int mgp_init_module(mgp_module *module, mgp_memory *memory) {
+  try {
+    // Dynamic Katz centrality
+    {
+      auto default_alpha = mgp::value_make_double(0.2, memory);
+      auto default_epsilon = mgp::value_make_double(1e-2, memory);
+
+      auto *proc = mgp::module_add_read_procedure(module, kProcedureSet, GetKatzCentrality);
```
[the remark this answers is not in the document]
antepusic: The current state of this code means that `.set` calls `GetKatzCentrality` instead of `SetKatzCentrality`. Consequently, any non-default `alpha` and `epsilon` values are ignored; also consider adding a test with all parameters custom-set.

antepusic (changes requested):
Code looks solid! :+1:

Quick comment - are the separate empty files in `/algorithm_online` necessary? The algorithm as implemented has both offline and online functionality.

cpp/mg_utility/mg_utils.hpp
```diff
@@ -319,4 +322,40 @@ std::vector<std::pair<std::uint64_t, std::uint64_t>> GetEdgeEndpointIDs(mgp_list
...
+}
+
+std::vector<std::uint64_t> GetVerticesFromList(mgp_list *list) {
+  auto size = mgp::list_size(list);
+  auto vertices = std::vector<std::uint64_t>(size);
+  for (std::size_t i = 0; i < size; i++) {
+    vertices[i] = mgp::vertex_get_id(mgp::value_get_vertex(mgp::list_at(list, i))).as_int;
+  }
+  return vertices;
+}
```
antepusic, line 360: I've implemented the `GetNodeIDs` and `GetEdgeEndpointIDs` utilities before; these may be redundant.

cpp/katz_centrality_module/algorithm/katz.cpp
```diff
@@ -0,0 +1,405 @@
...
+                 const std::vector<std::pair<std::uint64_t, uint64_t>> &new_edges,
+                 const std::vector<std::pair<std::uint64_t, uint64_t>> &deleted_edges,
+                 const std::set<std::uint64_t> &new_edge_ids, const mg_graph::GraphView<> &graph) {
+  auto i = context_new.iteration;
+
+  // Create a queue of nodes for update
+  std::queue<std::uint64_t> queue;
+  for (auto v : updated_nodes) {
+    queue.push(v);
+  }
```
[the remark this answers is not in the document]
jmatak: Wouldn't this try to construct the `queue` with 2 values, one from the `updated_nodes` start, and the other with `updated_nodes` end?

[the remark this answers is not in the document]
jmatak: Reference: https://stackoverflow.com/questions/8168127/how-can-i-copy-an-entire-vector-into-a-queue

cpp/katz_centrality_module/algorithm/katz.cpp
jmatak: Static algorithm convergence is done in the method called `KatzCentralityLoop`

jmatak: It's applied after initial handling of dynamic case.

cpp/katz_centrality_module/algorithm/katz.hpp
jmatak: I tried to hide them in the private methods of this module. We can discuss what is the best way to do it, but it certainly is a valuable proposal :+1:

cpp/mg_utility/mg_utils.hpp
jmatak: My bad. It was here before, just didn't see it. Tnx

cpp/katz_centrality_module/algorithm/katz.cpp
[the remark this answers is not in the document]
antepusic: You're right! I must've been thinking of other data structures that can be initialized this way.

cpp/katz_centrality_module/algorithm/katz.cpp
antepusic: I see, thank you!

cpp/katz_centrality_module/algorithm/katz.hpp
antepusic: Either way looks good to me :ok_hand:

e2e/katz_online_test/test_online_call_get_on_inconsistent/test.yml
```diff
@@ -0,0 +1,29 @@
...
+      - node_id: 4
+      - node_id: 0
+      - node_id: 3
+      - node_id: 5
+      - node_id: 2
+
+- query: >-
+    CALL katz_centrality_online.get() YIELD *;
+  exception: >-
+    Graph should be inconsistent at this point!
```
antepusic, line 29: Nothing major, but it would be good if us across MAGE were consistent in using the same error message for this case.
So far, I've been using `"Graph has been modified and is thus inconsistent with cached {NAME} scores; to update them, please call set/reset!"`

jmatak: I agree, but the emphasis on consistency part should be in algorithm's code rather than in tests. Testing module is implemented just to check whether there is an exception. The comment below it does nothing, but it's useful to check what is really going on.

antepusic: Thanks!

PR #118, opened 2022-01-19
Add example how to fix iostream/cout/cerr/stringstream locale issue

gitbuda:


gitbuda:
All locales should work out of the box. Worst case we implement library functions (inside `mg_procedures.h` and `mg_utils.hpp`).

gitbuda:
Yep, thanks for the notification!

PR #120, opened 2022-02-04
Fix build for ARM and newer toolchain

antonio2368:


Dockerfile.arm64
```diff
@@ -0,0 +1,78 @@
+ARG PY_VERSION_DEFAULT=3.9
+
+FROM dokken/centos-stream-9 as base
+
+USER root
+
+ARG PY_VERSION_DEFAULT
+ARG MEMGRAPH_RPM
```
jmatak, line 8: Why is this an argument, and not the image downloaded from the Memgraph download page?

Dockerfile.arm64
```diff
@@ -0,0 +1,78 @@
...
+    libcurl        `memgraph` \
+    openssl         `memgraph` \
+    make `mage-memgraph` \
+    cmake           `mage-memgraph` \
+    curl            `mage-memgraph` \
+    g++             `mage-memgraph` \
+    python3         `mage-memgraph` \
+    python3-pip     `mage-memgraph` \
+    python3-setuptools     `mage-memgraph` \
+    python3-devel     `mage-memgraph` \
```
jmatak, line 25: Where is the guarantee that the Python version matches the above-mentioned PY_VERSION_DEFAULT? 

@antoniofilipovic

Dockerfile.arm64
antonio2368: This should be changed, this PR for now is something that works only locally and with manually building the RPM package on an ARM machine. 
This isn't something that should be merged as is, I only wanted to make and show changes that are necessary to make it work on Centos 9/ARM64.

PR #125, opened 2022-02-24
Max flow implementation

brunos252:
### Description

Algorithm implementation for finding max flow in a graph. Ford-Fulkerson algorithm was used along with capacity scaling.
Also implemented is a BFS search algorithm for finding min and max values of an edge weight, used for the capacity scaling.

### Pull request type

- [x] Algorithm/Module

### Related issues

#### Module/Algorithm
- [x] Unit tests
- [x] End-to-end tests
- [x] Code documentation


python/mage/max_flow/bfs_weight_min_max.py
```diff
@@ -0,0 +1,33 @@
...
+
+    while next_queue:
+        current_v = next_queue.pop(0)
+        visited.add(current_v)
+
+        for e in current_v.out_edges:
+            if e.properties[edge_property] > max_weight:
+                max_weight = e.properties[edge_property]
+            elif e.properties[edge_property] < min_weight:
+                min_weight = e.properties[edge_property]
```
jmatak, line 28:
```suggestion
            max_weight = max(max_weight, e.properties.get(edge_property, 1)
            min_weight = min(min_weight, e.properties.get(edge_property, 1)
```
How max and min are calculated isn't so important, but the above would not fail if a user has sent not existing property to an algorithm.

python/max_flow.py
```diff
@@ -0,0 +1,172 @@
...
+    return: number value of graph max flow
+
+    The procedure can be invoked in openCypher using the following call:
+    MATCH (source {id: 0}), (sink {id: 5})
+    CALL max_flow.get_flow(source, sink, "weight")
+    YIELD max_flow
+    RETURN max_flow
+    """
+
+    records = get_paths(context, start_v, end_v, edge_property)
```
jmatak, line 31: I would rather make another method with different return types, because this one is highly coupled with Memgraph Record

python/max_flow.py
```diff
@@ -0,0 +1,172 @@
...
+
+    :return: flow_bottleneck, smallest remaining capacity on the path,
+             -1 if no path to end_node is found
+    """
+
+    # instead of using residual edges, we check for in_edges with flow
+    for edge in chain(current_v.out_edges, current_v.in_edges):
+        if edge.from_vertex == current_v:
+            to_v = edge.to_vertex
+            remaining_capacity = edge.properties[edge_property] - edge_flows.get(
```
jmatak, line 143: Again, watch for the `edge_property` that might not exist

python/max_flow.py
```diff
@@ -0,0 +1,172 @@
...
+    edge_flows: Dict,
+) -> mgp.Number:
+    """
+    Finds augmenting path for max_flow algorithm using recursive DFS
+    with minimum edge weight delta, as defined by capacity scaling.
+
+    :param path: list for storing path, elements are
+                 alternating mgp.VertexId and mgp.Edge
+    :param delta: lower bound for path flow
+    param edge_flows: dict containing existing flows of edges
```
jmatak, line 133:
```suggestion
    :param edge_flows: dict containing existing flows of edges
```

python/max_flow.py
```diff
@@ -0,0 +1,172 @@
...
+                if isinstance(elem, mgp.Edge):
+                    path.expand(elem)
+
+        return_paths_and_flows.append((path, flow_bottleneck))
+
+    return [mgp.Record(path=path, flow=flow) for path, flow in return_paths_and_flows]
+
+
+def DFS_path_finding(
+    context: mgp.ProcCtx,
```
jmatak, line 118: You don't need context here since it is a user-function, not a Memgraph read procedure.

python/max_flow.py
```diff
@@ -0,0 +1,172 @@
...
+
+            flow_bottleneck = DFS_path_finding(
+                context, path, to_v, end_v, edge_property, delta, edge_flows
+            )
+            if flow_bottleneck != -1:
+                # function call found path, propagate back
+                return min(remaining_capacity, flow_bottleneck)
+
+    # no path found with this vertex, remove it and its edge
+    del path[-2:]
```
jmatak, line 170: Is there a better way to remove last two elements in Python?

e2e/max_flow_test/test_paths/test.yml
```diff
@@ -0,0 +1,8 @@
+query: >
+  MATCH (source {id: 0}), (sink {id: 5})
+  CALL max_flow.get_paths(source, sink)
+  YIELD path, flow
+  RETURN count(path) AS count
+
+output:
+    - count: 4
```
jmatak, line 8: Create another empty test case, and test case with multiple components, since that might be the edge-case

jmatak (changes requested):

python/max_flow.py
brunos252: path = path[:-2] ? why is it better?

python/max_flow.py
brunos252: Not sure what you mean, my idea is that get_paths is also a query module, used for retrieving paths taken by max flow, this could be useful e.g. for min-cut. get_flow is used to just get the max flow.

jmatak: Memgraph's `mgp.Record` isn't something that should be traversed through, module should be an independent unit, and no one except for Memgraph's internal code should use Record fields in their program

python/mage/max_flow/bfs_weight_min_max.py
```diff
@@ -0,0 +1,36 @@
...
+    Breadth-first search for finding the largest and smallest edge weight,
+    largest being used for capacity scaling, and smallest for lower bound
+
+    :param start_v: starting vertex
+    :param edge_propery: str denoting the edge property used as weight
+
+    :return: Number, the largest value of edge_property in graph
+    """
+
+    next_queue = [start_v]
```
jmatak, line 15: Consider using `collections.dequeu()` here since it should be more optimized for popping elements from start

python/mage/max_flow/bfs_weight_min_max.py
```diff
@@ -0,0 +1,36 @@
...
+    min_weight = float("Inf")
+
+    while next_queue:
+        current_v = next_queue.pop(0)
+        visited.add(current_v)
+
+        for e in current_v.out_edges:
+            # if there are edges without the given property, we ignore them,
+            # in order to support heterogeneous graphs
+            if edge_property not in e.properties:
```
jmatak, line 27: Don't forget to mention this in the documentation because traversing through the graph where we do not have a valid weight property would result in having `max_flow=0`

jmatak (commented):
I have one comment related to using `deque`, and what's left is writing the documentation on memgraph/docs:

https://github.com/memgraph/docs/tree/master/mage/algorithms/traditional-graph-analytics
https://github.com/memgraph/docs/tree/master/mage/query-modules/python

Enjoy :smile:

e2e/max_flow_test/test_edge_without_property/input.cyp
```diff
@@ -0,0 +1,9 @@
+MERGE (a:Node {id: 0}) MERGE (b:Node {id: 1}) CREATE (a)-[:RELATION {weight: 6}]->(b);
+MERGE (a:Node {id: 0}) MERGE (b:Node {id: 2}) CREATE (a)-[:RELATION {weight: 14}]->(b);
+MERGE (a:Node {id: 1}) MERGE (b:Node {id: 2}) CREATE (a)-[:RELATION {weight: 1}]->(b);
+MERGE (a:Node {id: 1}) MERGE (b:Node {id: 3}) CREATE (a)-[:RELATION {weight: 5}]->(b);
+MERGE (a:Node {id: 2}) MERGE (b:Node {id: 3}) CREATE (a)-[:RELATION {weight: 7}]->(b);
+MERGE (a:Node {id: 2}) MERGE (b:Node {id: 4}) CREATE (a)-[:RELATION {weight: 10}]->(b);
+MERGE (a:Node {id: 3}) MERGE (b:Node {id: 4}) CREATE (a)-[:RELATION {weight: 1}]->(b);
+MERGE (a:Node {id: 3}) MERGE (b:Node {id: 5}) CREATE (a)-[:RELATION {wght: 11}]->(b);
+MERGE (a:Node {id: 4}) MERGE (b:Node {id: 5}) CREATE (a)-[:RELATION {weight: 12}]->(b);
```
antoniofilipovic, line 9: @jmatak are we adding a new line at the end of the file or not, I forgot what was the rule?

python/max_flow.py
```diff
@@ -0,0 +1,205 @@
...
+    :param augmenting_path: List of interchangeable VertexId and EdgeId
+
+    :return: mgp.Path structure
+    """
+    for i, elem in enumerate(augmenting_path):
+        if i == 0:
+            path = mgp.Path(context.graph.get_vertex_by_id(augmenting_path[0]))
+        else:
+            if isinstance(elem, mgp.Edge):
+                path.expand(elem)
```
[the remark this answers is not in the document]
brunos252: yeah, will do!

Josipmrden (commented):
Looks pretty good to me, minor comment only

Josipmrden (approved):

e2e/max_flow_test/test_edge_without_property/input.cyp
jmatak: The rule is to add it

e2e/max_flow_test/test_empty/input.cyp
```diff
@@ -0,0 +1 @@
+CREATE ({id: 0}), ({id: 5})
```
jmatak, line 1: @brunos252 Set your IDE to add a new line whenever you save a file.

jmatak (approved):
Fix new line and it's finished :tada:

jmatak (approved):

PR #135, opened 2022-04-11
Fix IsSubset checking for unordered sets

jmatak:
### Description

std::includes works for ordered containers. Change `IsSubset` method to work with unordered set in `O(b)` time.

### Pull request type

- [x] Bugfix


cpp/community_detection_module/algorithm_online/community_detection.cpp
```diff
@@ -18,7 +18,7 @@ namespace LabelRankT {
 ///@return -- whether a is subset of b
 template <class T>
 bool IsSubset(const std::unordered_set<T> &a, const std::unordered_set<T> &b) {
-  return std::includes(a.begin(), a.end(), b.begin(), b.end());
+  return std::all_of(b.begin(), b.end(), [&](const T &value) { return a.find(value) != a.end(); });
```
antepusic, line 21: Looks like you switched `a` and `b` (the function checks whether `a` is a subset of `b`).

antepusic, line 21:
```suggestion
  return std::all_of(a.begin(), a.end(), [&](const T &value) { return b.find(value) != b.end(); });
```

antepusic (changes requested):

antepusic (approved):

PR #150, opened 2022-06-21
Igraph support

niko4299:
### Description

 Igraph algorithms (link to their github https://github.com/igraph/python-igraph) 

### Pull request type

- [X] Algorithm/Module

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation


Josipmrden (commented):
Some minor comments. Looks good overall, I have just one concern about creating the igraph, otherwise it's totally fine.

python/igraphalg.py
```diff
@@ -0,0 +1,283 @@
...
+
+@mgp.read_proc
+def get_flow(
+    ctx: mgp.ProcCtx,
+    start_v: mgp.Vertex,
+    end_v: mgp.Vertex,
+    edge_property: str = "weight",
+) -> mgp.Record(max_flow=mgp.Number):
+
+    graph, id_mapping, _ = create_igraph_from_ctx(ctx, directed=True)
```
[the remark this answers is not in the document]
niko4299: I think max_flow needs to be directed.

python/igraphalg.py
```diff
@@ -0,0 +1,290 @@
...
+) -> mgp.Record(length=float):
+    graph, id_mapping, _ = create_igraph_from_ctx(ctx, directed=directed)
+
+    return mgp.Record(
+        length=graph.shortest_paths(
+            source=id_mapping[source.id],
+            target=id_mapping[target.id],
+            weights=edge_property,
+        )[0][0]
+    )
```
antepusic, line 162: Why are we taking only the first element here?

python/igraphalg.py
```diff
@@ -0,0 +1,290 @@
...
+    target: mgp.Vertex,
+    edge_property: mgp.Nullable[str] = None,
+    directed: bool = True,
+) -> mgp.Record(path=List[mgp.Vertex]):
+    graph, id_mapping, invert_id_mapping = create_igraph_from_ctx(
+        ctx, directed=directed
+    )
+    path = graph.get_shortest_paths(
+        v=id_mapping[source.id], to=id_mapping[target.id], weights=edge_property
+    )[0]
```
antepusic, line 193: How is the output of `get_shortest_paths` structured, why do we take only the first element?

antepusic (changes requested):
I’d recommend aligning procedure and argument names with those used by igraph to ensure consistent user experience.

python/igraphalg.py
niko4299: We take only first element because they use same function for finding shortest path between source and one destination and for finding shortest path between source and multiple destinations. Function returns list that contains shortest path between each source node and destination node.

python/igraphalg.py
niko4299: Output of `shortest_paths` is 2D list that contains shortest path lengths between source nodes and destination nodes. In our case we specified source node and destination node so 2D list will have only one member but if didnt specify source and destination we would get 2D list that will contain shortest path length between each pair of nodes.

python/igraphalg.py
```diff
@@ -0,0 +1,283 @@
...
+                (id_mapping[edge.from_vertex.id], id_mapping[edge.to_vertex.id])
+            )
+
+    graph = igraph.Graph(
+        directed=directed,
+        n=len(ctx.graph.vertices),
+        edges=edge_list,
+        edge_attrs=edge_attrs,
+        vertex_attrs=vertex_attrs,
+    )
```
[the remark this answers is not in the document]
niko4299: I think this is the fastest way to create Igraph.Graph but if you think there is nicer way to do it I will look into it.

python/igraphalg.py
jmatak: It would be beneficial to return the Path object, not only the ids of path nodes

python/igraphalg.py
```diff
@@ -0,0 +1,290 @@
...
+
+def convert_vertex_ids_to_mgp_vertices(
+    vertex_ids: List[int], ctx: mgp.ProcCtx, inverted_id_mapping: Dict[int, int]
+) -> List[mgp.Vertex]:
+
+    vertices = []
+    for id in vertex_ids:
+        vertices.append(ctx.graph.get_vertex_by_id(inverted_id_mapping[id]))
+
+    return vertices
```
jmatak, line 208: This can be abstracted in graph class

python/igraphalg.py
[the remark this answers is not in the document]
jmatak: Personally, I would try to abstract this in a class with all the necessary information stored behind the class abstraction. The algorithms would then return this abstraction for example

python/igraphalg.py
niko4299: Currently I'm returning list of mgp.Vertex not just vertex ids. I can convert it to mgp.Path with only vertices without edges, if you think there is a difference.

e2e/igraphalg_test/test_shortest_path_length/test.yml
```diff
@@ -0,0 +1,8 @@
+query: >
+  MATCH (source {id: 0}), (sink {id: 5})
+  CALL igraphalg.shortest_path_length(source, sink,"weight")
+  YIELD length RETURN length;
+
+output:
+    - length: 22.0
+
```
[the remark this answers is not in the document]
niko4299: Every other test.yml from all other testing directories has an emtpy line at the end of file.  So should I change all others or?

python/igraphalg.py
niko4299: It returns empty list.

e2e/igraphalg_test/test_shortest_path_length/test.yml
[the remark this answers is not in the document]
antoniofilipovic: @jmatak I think we agreed that it should have a newline character at the file. But here it seems like there is an extra newline (line 8). On PR there should not be this red circle sign.

e2e/igraphalg_test/test_shortest_path_unweighted/test.yml
```diff
@@ -0,0 +1,9 @@
+query: >
+  MATCH (source {id: 0}), (sink {id: 5})
+  CALL igraphalg.get_shortest_path(source, sink)
+  YIELD path
+  RETURN extract(node IN path | node.id) AS path;
+
+output:
+    - path: [0, 1, 3, 5]
+
```
antoniofilipovic, line 9: One line empty to many it seems.

python/igraphalg.py
```diff
@@ -0,0 +1,168 @@
...
+        value=cut.value,
+    )
+
+
+@mgp.read_proc
+def topological_sort(
+    ctx: mgp.ProcCtx, directed: bool = True, mode: str = "out"
+) -> mgp.Record(nodes=mgp.Nullable[mgp.List[mgp.Vertex]]):
+    graph = MemgraphIgraph(ctx=ctx, directed=directed)
+    sorted_nodes = graph.topological_sort(mode=mode)
```
antoniofilipovic, line 78: Do we want to check here if the user passed the correct mode?

python/igraphalg.py
```diff
@@ -1,23 +1,20 @@
...
 
 @mgp.read_proc
-def get_flow(
+def maxflow(
     ctx: mgp.ProcCtx,
-    start_v: mgp.Vertex,
-    end_v: mgp.Vertex,
-    edge_property: str = "weight",
+    source: mgp.Vertex,
+    destination: mgp.Vertex,
```
antepusic, line 12: This parameter is called `target` in igraph ([reference](https://igraph.org/python/api/latest/igraph.Graph.html#maxflow)).

python/igraphalg.py
```diff
@@ -26,154 +23,133 @@ def get_flow(
...
+    graph = MemgraphIgraph(ctx=ctx, directed=True)
+
+    cut = graph.mincut(source=source, target=target, capacity=capacity)
 
     return mgp.Record(
-        partition_vertex_ids=[
-            convert_vertex_ids_to_mgp_vertices(partition, ctx, inverted_id_mapping)
-            for partition in mincut.partition
+        partition_vertices=[
+            graph.convert_vertex_ids_to_mgp_vertices(vertex_ids=partition)
```
antepusic, line 66: Try to handle the ID conversion inside the `MemgraphIgraph` class.

python/igraphalg.py
```diff
@@ -26,154 +23,133 @@ def get_flow(
...
     ctx: mgp.ProcCtx,
-    edge_property: mgp.Nullable[str] = None,
     resolution_parameter: float = 0.6,
-    number_of_iterations: int = -1,
+    n_iterations: int = -1,
+    beta: float = 0.01,
+    weights: mgp.Nullable[str] = None,
+    objective_function="modularity",
+    directed: bool = False,
 ) -> mgp.Record(community_id=int, community_members=List[mgp.Vertex]):
```
antepusic, line 94: Make the arguments consistent with [igraph](https://igraph.org/python/api/latest/igraph.Graph.html#community_leiden).

python/igraphalg.py
```diff
@@ -26,154 +23,133 @@ def get_flow(
...
     return [
         mgp.Record(
             community_id=i,
-            community_members=convert_vertex_ids_to_mgp_vertices(
-                members, ctx, inverted_id_mapping
-            ),
+            community_members=members,
         )
         for i, members in enumerate(communities)
     ]
```
antepusic, line 111: Same comment as with `min_cut`.

antepusic (changes requested):
The code is much cleaner now :+1:

e2e/igraphalg_test/test_all_shortest_path_length/test.yml
```diff
@@ -0,0 +1,10 @@
+query: >
+  MATCH (source {id: 0}), (sink {id: 5})
+  CALL igraphalg.all_shortest_path_lengths("weight")
+  YIELD lengths
+  RETURN lengths;
+
+output:
+    - lengths: [[0.0, 6.0, 7.0, 11.0, 12.0, 22.0], [6.0, 0.0, 1.0, 5.0, 6.0, 16.0], [7.0, 1.0, 0.0, 6.0, 7.0, 17.0],
+                [11.0, 5.0, 6.0, 0.0, 1.0, 11.0], [12.0, 6.0, 7.0, 1.0, 0.0, 12.0],[22.0, 16.0, 17.0, 11.0, 12.0, 0.0]]
```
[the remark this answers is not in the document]
niko4299: I changed it to return nodes and lengths and I think user can see how the nodes are ordered in length matrix but if you think there is a better way I will look into it a little bit more.

python/igraphalg.py
```diff
@@ -26,154 +23,133 @@ def get_flow(
...
 
 
 @mgp.read_proc
 def spanning_tree(
-    ctx: mgp.ProcCtx, edge_property: mgp.Nullable[str] = None, directed: bool = True
+    ctx: mgp.ProcCtx, weights: mgp.Nullable[str] = None, directed: bool = True
 ) -> mgp.Record(tree=List[List[mgp.Vertex]]):
```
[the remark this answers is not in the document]
niko4299: I will remove `directed` argument but for now I will leave out `return_tree` if that is okay. With `return_tree` true or false we have same functionality only result is in different format also with `return_tree = False` algorithm is much faster and it is easier to convert final result to format that I need. In future if we are going to need this argument I will implement it.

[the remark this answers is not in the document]
antepusic: Your reasoning makes sense, resolving this conversation!

antepusic (commented):
Thanks for addressing my questions regarding the module’s interface!

antepusic (approved):

e2e/igraphalg_test/test_leiden/test.yml
```diff
@@ -0,0 +1,19 @@
...
+      community_members: [1]
+    - community_id: 2
+      community_members: [2]
+    - community_id: 3
+      community_members: [3]
+    - community_id: 4
+      community_members: [4]
+    - community_id: 5
+      community_members: [5]
+
```
jmatak, line 19: Try downloading the "new line" formatter in VSCode or whatever you are using. There won't be flags like these in the code then.

e2e/igraphalg_test/test_leiden/test.yml
```diff
@@ -0,0 +1,19 @@
...
+    - community_id: 1
+      community_members: [1]
+    - community_id: 2
+      community_members: [2]
+    - community_id: 3
+      community_members: [3]
+    - community_id: 4
+      community_members: [4]
+    - community_id: 5
+      community_members: [5]
```
jmatak, line 18: How come each and every node is in its own community. I hope this is generated with some baseline implementation like `networkx`.

jmatak (changes requested):
This is much better than in the first iteration. Most of my questions are related to the API format. I think we should talk about procedure signatures.

e2e/igraphalg_test/test_leiden/test.yml
niko4299: It is because `resoultion_parameter` has value 1 which means that communities will be smaller, more connected. If you want I can call function with different value so there we will be bigger communities or I can maybe think of better test graph.

e2e/igraphalg_test/test_min_cut_unweighted/test.yml
```diff
@@ -0,0 +1,9 @@
+query: >
+  MATCH (source {id: 0}), (sink {id: 5})
+  CALL igraphalg.mincut(source, sink)
+  YIELD partition_vertices, value
+  RETURN extract(partition IN partition_vertices | extract (node in partition | node.id)) as partition_vertex_ids,value;
+
+output:
+    - partition_vertex_ids: [[0,1,2,3,4],[5]]
+      value: 2.0
```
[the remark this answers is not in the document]
niko4299: Value is sum of edges weights we needed to remove to do a min cut.

python/mgp_igraph.py
```diff
@@ -0,0 +1,232 @@
...
+        )
+
+    def maxflow(
+        self, source: mgp.Vertex, target: mgp.Vertex, capacity: str
+    ) -> igraph.Flow:
+        return super().maxflow(
+            self.id_mappings[source.id],
+            self.id_mappings[target.id],
+            capacity=capacity,
+        )
```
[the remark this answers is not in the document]
niko4299: Okay, I will change that.

python/mgp_igraph.py
```diff
@@ -0,0 +1,232 @@
...
+        if weights:
+            weights = self.es[weights]
+
+        min_spanning_tree_edges = super().spanning_tree(
+            weights=weights, return_tree=False
+        )
+
+        return self._get_min_span_tree_vertex_pairs(
+            min_spanning_tree_edges=self.es[min_spanning_tree_edges]
+        )
```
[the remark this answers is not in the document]
niko4299: Output of this function will be list of tuples of connected nodes in minimum spanning tree.

python/igraphalg.py
```diff
@@ -0,0 +1,201 @@
...
+    directed: bool = True,
+) -> mgp.Record(length=float):
+    graph = MemgraphIgraph(ctx, directed=directed)
+
+    return mgp.Record(
+        length=graph.shortest_path_length(
+            source=source,
+            target=target,
+            weights=weights,
+        )
```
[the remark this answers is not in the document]
niko4299: Output is float value which is length between source and target node.

python/igraphalg.py
```diff
@@ -174,13 +181,19 @@ def all_shortest_path_lengths(
...
-        nodes=[vertex for vertex in ctx.graph.vertices],
-        lengths=graph.all_shortest_path_lengths(weights=weights),
-    )
+    return [
+        mgp.Record(
+            src_node=graph.get_vertex_by_id(i),
+            dest_node=graph.get_vertex_by_id(j),
+            length=lengths[i][j],
+        )
+        for j in range(len(lengths[0]))
```
antepusic, line 194: Is `lengths[i]` the same size for all `i`s?

antepusic (approved):

python/igraphalg.py
jmatak: @antoniofilipovic  This looks fixed

python/igraphalg.py
```diff
@@ -0,0 +1,211 @@
...
+        node_weights=node_weights,
+    )
+
+    return [
+        mgp.Record(
+            node=member[0],
+            community_id=member[1],
+        )
+        for member in communities
+    ]
```
jmatak, line 147:
```suggestion
    return [
        mgp.Record(
            node=node,
            community_id=community_id,
        )
        for node, community_id in communities
    ]
```

jmatak (changes requested):
Two more things need to be done. First one is testing on edge cases. Try testing on cases where there is no graph, and cases when the nodes are faulty or there is no path between them for example. The second thing is documentation. Good work  :muscle:, keep on going!

antepusic (approved):

antoniofilipovic (approved):

PR #152, opened 2022-06-27
Allow restricting community detection to subgraphs

### LINKED ISSUE #149: Add ability to execute CALL on subgraph
Add ability to execute logic on a subgraph or projection.

For example, on the sample European roads network, be able to get communities for only the subgraph selected by the below MATCH:

MATCH (a:City{name:'Rome'})-[road:Road*bfs]->(b:City)
 
CALL community_detection.get(
        "length", false,  100000,  1e-06,  0.01
        )
        YIELD community_id, node
        set node.community = community_id
        ;

One possibility is to add nodeProjection and relationshipProjection as CALL parameters:
CALL gds.louvain.write({ nodeProjection: ...,relationshipProjection: , writeProperty: ... {})

antepusic:
### Description

Extend the `community_detection` module to allow restricting community detection to subgraphs.

### Pull request type

- [x] Feature

### Related issues

#149 

#### Module/Algorithm
- [x] Unit tests
- [x] End-to-end tests
- [x] Code documentation


cpp/mg_utility/mg_utils.hpp
```diff
@@ -202,67 +220,159 @@ std::unique_ptr<mg_graph::Graph<TSize>> GetWeightedGraphView(mgp_graph *memgraph
...
+         out_edge = mgp::edges_iterator_next(edges_it)) {
+      if (subgraph_edge_ids.find(mgp::edge_get_id(out_edge).as_int) ==
+          subgraph_edge_ids.end()) {
+        continue;
+      }
+      auto vertex_to = mgp::edge_get_to(out_edge);
+      auto weight = mg_utility::GetNumericProperty(out_edge, weight_property,
+                                                   memory, default_weight);
+      mg_graph::CreateWeightedGraphEdge(graph.get(), vertex_from, vertex_to,
+                                        weight, graph_type);
```
jmatak, line 315: Two comments here. Firstly, let's make these methods re-usable rather than copy/pasting from another place. Second thing, the subgraph can be defined only by using a list of edges, no need to send the list of nodes because that can be extracted from the list of edges.

cpp/community_detection_module/community_detection_module.cpp
```diff
@@ -6,46 +6,86 @@
 namespace {
 
 const char *kProcedureGet = "get";
+const char *kProcedureGetSubgraph = "get_subgraph";
 
 const char *kFieldNode = "node";
 const char *kFieldCommunity = "community_id";
 
-const char *kArgumentWeightProperty = "weight";
+// const char *kArgumentWeightProperty = "weight";
```
jmatak, line 14: Why is this not used?

jmatak (changes requested):
I have added some comments, mostly pointing to the changes related to the repetition of code.

cpp/mg_utility/mg_utils.hpp
antepusic: Fully agree about your first point. Regarding the second, using only the edges doesn’t work when the subgraph contains isolated nodes, and makes the function inconsistent with existing ones like `color_subgraph`.

cpp/mg_utility/mg_utils.hpp
```diff
@@ -202,67 +220,159 @@ std::unique_ptr<mg_graph::Graph<TSize>> GetWeightedGraphView(mgp_graph *memgraph
...
     }
   }
 
   return graph;
 }
 
 namespace {
-void InsertRecord(mgp_result_record *record, const char *field_name, mgp_value *value) {
+void InsertRecord(mgp_result_record *record, const char *field_name,
+                  mgp_value *value) {
```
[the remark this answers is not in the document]
antepusic: I am using Google-style formatting, going to check if line length is configured correctly or if it’s screen size-dependent in VSCode.

cpp/community_detection_module/community_detection_module.cpp
antepusic: Good catch! I commented it out because of it being unused, but now I noticed it has mistakenly been left out from line 71 in the main branch code.

antepusic:
> I have added some comments, mostly pointing to the changes related to the repetition of code.

Thanks for the review! The current state of the branch is a minimum viable solution, and I’m now improving the code.

cpp/mg_utility/mg_utils.hpp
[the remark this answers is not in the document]
antepusic: Just checked, mine is 80 cols/chars and in line with the [style guide](https://google.github.io/styleguide/cppguide.html#Line_Length).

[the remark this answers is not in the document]
jbajic: Commenting here since I was invited :smiley: . I am for introducing the clang-check in workflow since there already is a clang-format defined for mage. so the style won't depend on individual configuration.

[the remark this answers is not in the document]
antepusic: Switched to the settings in mage/.clang-format & agree with Jure :+1:

cpp/mg_utility/mg_utils.hpp
antepusic: Added a reusable method for all currently used graph views.

jmatak (approved):
This is much better, it's much clear to distinguish between weighted graph and subgraph.

antepusic:
@threedliteguy: Hi Dan! The PR adding community detection on subgraphs has been approved - come and try it out :slightly_smiling_face:

cpp/mg_utility/mg_utils.hpp
```diff
@@ -101,34 +105,55 @@ class OnScopeExit {
...
+    mgp_graph *memgraph_graph, mgp_result *result, mgp_memory *memory, const mg_graph::GraphType graph_type,
+    const bool weighted = false, const char *weight_property = "weight", double default_weight = 1.0,
+    const bool subgraph = false, mgp_list *subgraph_nodes = nullptr, mgp_list *subgraph_relationships = nullptr) {
   auto graph = std::make_unique<mg_graph::Graph<TSize>>();
 
+  std::unordered_set<std::uint64_t> subgraph_node_ids, subgraph_edge_ids;
+  if (subgraph) {
+    for (std::size_t i = 0; i < mgp::list_size(subgraph_nodes); i++) {
+      subgraph_node_ids.insert(mgp::vertex_get_id(mgp::value_get_vertex(mgp::list_at(subgraph_nodes, i))).as_int);
+    }
```
antoniofilipovic, line 137: Why don't we save here reference to `vertex`, and later use that reference, instead of checking inside `vertices_it` whether there exists a node with such `id`?

cpp/mg_utility/mg_utils.hpp
```diff
@@ -101,34 +105,55 @@ class OnScopeExit {
...
-
   auto *vertices_it = mgp::graph_iter_vertices(memgraph_graph, memory);
   mg_utility::OnScopeExit delete_vertices_it([&vertices_it] { mgp::vertices_iterator_destroy(vertices_it); });
 
   // Iterate through Memgraph vertices and map them to GraphView
   for (auto *vertex = mgp::vertices_iterator_get(vertices_it); vertex;
        vertex = mgp::vertices_iterator_next(vertices_it)) {
+    if (subgraph && subgraph_node_ids.find(mgp::vertex_get_id(vertex).as_int) == subgraph_node_ids.end()) {
+      continue;
+    }
```
antoniofilipovic, line 156: Maybe we don't need this here, this seems like a costly operation and if we only save a reference to vertex instead of `vertix_id`, everything would be fine, right?

cpp/mg_utility/mg_utils.hpp
```diff
@@ -143,75 +168,55 @@ std::unique_ptr<mg_graph::Graph<TSize>> GetGraphView(mgp_graph *memgraph_graph,
...
-  return graph;
+template <typename TSize = std::uint64_t>
+std::unique_ptr<mg_graph::Graph<TSize>> GetWeightedSubgraphView(mgp_graph *memgraph_graph, mgp_result *result,
+                                                                mgp_memory *memory, mgp_list *subgraph_nodes,
+                                                                mgp_list *subgraph_relationships,
+                                                                const mg_graph::GraphType graph_type,
+                                                                const char *weight_property, double default_weight) {
+  return GraphViewGetter(memgraph_graph, result, memory, graph_type, true, weight_property, default_weight, true,
+                         subgraph_nodes, subgraph_relationships);
 }
```
antoniofilipovic, line 220: If we want to organize it like this, isn't then `GetSubgraphView` function missing (unweighted + subgraph), since we have
`GetWeightedSubgraphView` (weight + sub),
`GetWeightedGraphView` (weight + graph),
and `GetGraphView` (unweight + graph)?

antoniofilipovic (changes requested):
There are two comments to discuss, so I will put `request changes` since it seems like we could use both of them.

cpp/mg_utility/mg_utils.hpp
antepusic: As `vertex_get_id` is already called in `CreateGraphNode`, would it make sense to move the vertex existence check inside that function?

cpp/mg_utility/mg_utils.hpp
```diff
@@ -53,7 +56,8 @@ void CreateGraphEdge(mg_graph::Graph<TSize> *graph, mgp_vertex *vertex_from, mgp
   graph->CreateEdge(memgraph_id_from, memgraph_id_to, memgraph_edge_id, graph_type);
 }
 
-///@brief Creates a weighted edge within GraphView. Requires Memgraph’s UIDs for its endpoints.
+///@brief Creates a weighted edge within GraphView. Requires Memgraph’s UIDs for
+/// its endpoints.
```
jmatak, line 60: Is this file correctly formatted?

cpp/mg_utility/mg_utils.hpp
jmatak: I think you should only traverse the nodes provided by the function input. However, this check of inclusiveness might still be relevant when traversing edges too since both source and destination nodes must be in the subgraph.

cpp/mg_utility/mg_utils.hpp
```diff
@@ -143,75 +168,55 @@ std::unique_ptr<mg_graph::Graph<TSize>> GetGraphView(mgp_graph *memgraph_graph,
 
   for (auto *vertex_from = mgp::vertices_iterator_get(vertices_it); vertex_from;
        vertex_from = mgp::vertices_iterator_next(vertices_it)) {
+    if (subgraph && subgraph_node_ids.find(mgp::vertex_get_id(vertex_from).as_int) == subgraph_node_ids.end()) {
+      continue;
+    }
```
jmatak, line 173: Again, traverse only subgraph nodes, or if there is no subgraph traverse all nodes, create a lambda iterator.

cpp/mg_utility/mg_utils.hpp
jmatak: Agreed here :arrow_up:

jmatak (changes requested):
@antoniofilipovic  Left some good notes, shouldn't be much of a hassle to refactor that.

cpp/mg_utility/mg_utils.hpp
antepusic: Understanding your viewpoint, I would like to raise the following concern: when the user calls sth. like `MATCH (a:Node)-[e]->(b:Node) WHERE {criterion} WITH collect(a) AS nodes, collect(e) AS edges ...`,  `nodes` will contain duplicates, e.g.
* if edges `3->5` and `4->5` are `MATCH`ed, node `5` is duplicated
* undirected `MATCH`ing like `MATCH (a:Node)-[e]-(b:Node)` doubles *everything*
As for storing `mgp` types in sets, `set<mgp_vertex>` and `set<mgp_edge>` cause errors as `mgp_vertex` and `mgp_edge` are incomplete types. Extending the types calls for changing the MGP API; I’d already consulted Benjo about that (with iteration over mgp containers) and he was against because the API should be as C-like as possible
Doing `set<mgp_vertex*>` & `set<mgp_edge*>` doesn’t work either because duplicates have different hashes

With this in mind, I‘m keeping this part as-is for now.

antepusic: I’ve succeeded in resolving this by adding duplicate checks to the `CreateNode` and `CreateEdge` methods of graphview.

jmatak (changes requested):
Unify CreateEdge in mg_graph.hpp and fix creating node return value when node is already created.

cpp/mg_utility/mg_graph.hpp
```diff
@@ -152,14 +152,16 @@ class Graph : public GraphView<TSize> {
...
-    return !weights_.empty();
-  }
+  bool IsWeighted() const override { return !weights_.empty(); }
 
   /// Creates a node.
   ///
   /// @return     Created node id
   TSize CreateNode(std::uint64_t memgraph_id) {
+    if (memgraph_to_inner_id_.find(memgraph_id) != memgraph_to_inner_id_.end()) {
+      return 0;
```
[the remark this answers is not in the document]
antepusic: Fixed.

cpp/mg_utility/mg_utils.hpp
antepusic: Fixed.

cpp/mg_utility/mg_utils.hpp
antepusic: Fixed formatting (the snippet in question has been removed in the meantime btw).

jmatak:
@antepusic Comment out here the link to `memgraph/docs` PR

antepusic:
> @antepusic Comment out here the link to `memgraph/docs` PR

Here you are: https://github.com/memgraph/docs/pull/450

jmatak (approved):
Finished :tada:

antoniofilipovic (approved):

PR #158, opened 2022-07-25
New C++ query module API

antepusic:
### Description

Add and extend C++ API from #68, and update query modules to use it.

### Pull request type

- [ ] Bugfix
- [ ] Algorithm/Module
- [x] Feature
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Build related changes

### Related issues

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation


Josipmrden (commented):
I still have to review mage.hpp, here are the initial comments

cpp/mg_utility/mage.hpp
```diff
@@ -0,0 +1,2757 @@
...
+  static_assert(std::is_arithmetic<TSrc>::value, "MemcpyCast expects source to be an arithmetic type");
+  static_assert(std::is_arithmetic<TDest>::value, "MemcpyCast expects destination to be an arithmetic type");
+  std::memcpy(&dest, &src, sizeof(src));
+  return dest;
+}
+}  // namespace util
+
+// Forward declarations
+class Nodes;
+using GraphNodes = Nodes;
```
gitbuda, line 57: Why not just use `Nodes`? 🤔

cpp/mg_utility/mage.hpp
```diff
@@ -0,0 +1,2757 @@
...
+class GraphRelationships;
+class Relationships;
```
gitbuda, line 59: What's the difference between `GraphRelationships` and `Relationships`?

cpp/mg_utility/mage.hpp
```diff
@@ -0,0 +1,2757 @@
...
+  explicit Value(const std::string_view value) : ptr_(mgp::value_make_string(value.data(), memory)) {}
+
+  // Container constructors:
+
+  /// @brief Constructs a List value from the copy of the given `list`.
+  explicit Value(const List &list) : ptr_(mgp::value_make_list(list.ptr_)) {}
+  /// @note The behavior of accessing `list` after performing this operation is undefined.
+  explicit Value(List &&list) {
+    ptr_ = mgp::value_make_list(list.ptr_);
+    delete &list;
```
gitbuda, line 951: This looks scary 😂 Haven't seen this pattern before? @antaljanosbenjamin can maybe leave an opinion/guide here?

gitbuda (commented):
A couple of small comments + I think the style guide stuff should be a bit adjusted, but overall a bunch of work that will converge soon 💪

cpp/api_test_module/api_test_module.cpp
```diff
@@ -0,0 +1,589 @@
...
+  for (const auto element : list) {
+    if (!first) std::cout << ", ";
+    std::cout << element.ValueString();
+    first = false;
+  }
+  std::cout << "]\n";
+}
+
+void PrintMap(mage::Map &map) {
+  bool first = true;
```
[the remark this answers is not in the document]
antepusic: This is only a helper function for my "tests"; this entire part will be replaced with proper tests after moving it to core :hand:

cpp/api_test_module/api_test_module.cpp
```diff
@@ -0,0 +1,589 @@
...
+      return "LocalDateTime";
+    case mage::Type::Duration:
+      return "Duration";
+    default:
+      return "Type unknown";
+  }
+}
+
+void PrintList(mage::List &list) {
+  bool first = true;
```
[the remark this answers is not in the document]
antepusic: Ditto

cpp/api_test_module/api_test_module.cpp
```diff
@@ -0,0 +1,589 @@
...
+  std::cout << "[";
+  for (const auto element : list) {
```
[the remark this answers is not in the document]
antepusic: Done

cpp/mg_utility/mage.hpp
antepusic: In the existing C API one cannot iterate over all relationships in the graph, only over those adjacent to a node (see `vertex_iter_in_edges` and `vertex_iter_out_edges`).
`GraphRelationships` implements iterating over all relationships in the graph, whereas `Relationship` is just a wrapper over `mgp_edges_iterator`.

cpp/mg_utility/mage.hpp
antepusic: I’ve added the `GraphNodes` alias to make it clearer that this class can be used for iterating over all graph nodes (for iterating over relationships, two classes were necessary).

cpp/mg_utility/mage.hpp
```diff
@@ -0,0 +1,2757 @@
...
+
+  /// @brief Returns the graph order (number of nodes).
+  int64_t order() const;
+  /// @brief Returns the graph size (number of relationships).
+  int64_t size() const;
+
+  /// @brief Returns an iterable structure of the graph’s nodes.
+  GraphNodes nodes() const;
+  /// @brief Returns an iterable structure of the graph’s relationships.
+  GraphRelationships relationships() const;
```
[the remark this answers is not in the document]
antepusic: Thanks for the heads-up! This is something I have [discussed](https://memgraph.slack.com/archives/CSR01RS80/p1660824077606649) on Slack, I’ll make them uppercase.

cpp/mg_utility/mage.hpp
```diff
@@ -0,0 +1,2757 @@
...
+  List,
+  Map,
+  Node,
+  Relationship,
+  Path,
+  Date,
+  LocalTime,
+  LocalDateTime,
+  Duration,
+  UNDEF,  // for indicating an undefined value
```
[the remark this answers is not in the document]
antepusic: `UNDEF` is only used in `Parameter` objects under `list_item_type`, where it means that the parameter is not a list and thus list item type is left out.
Anyway, I’ll try to avoid this and make a `ListParameter` struct that inherits from `Parameter`, eliminating the need for one type of objects to handle both lists and non-lists.

cpp/mg_utility/mage.hpp
antepusic: This part is inherited from [Matak’s API concept](#68); it looked fine to me, but Benjo’s opinion is always appreciated! :slightly_smiling_face:

cpp/mg_procedure/mg_procedure.h
```diff
@@ -167,6 +167,8 @@ enum mgp_value_type {
   MGP_VALUE_TYPE_DURATION,
 };
 
+enum mgp_error mgp_value_copy(struct mgp_value *val, struct mgp_memory *memory, struct mgp_value **result);
+
```
antaljanosbenjamin, line 171: ~Why these copy functions needs to be part of the C API? Cannot they be done on the "user side"? Currently we only have copy functions for objects that are otherwise couldn't be copied. With these functions we add 3 copy functions, but not to others. My concern here is we make our API inconsistent with these, but I don't see a proper reason why are they important.~

`mgp_path` already has a copy function, so these copy functions are actually a good fit.

cpp/mg_utility/mgp.hpp
```diff
@@ -98,29 +101,40 @@ struct mgp_value *value_make_path(struct mgp_path *val) {
   return MgInvoke<mgp_value *>(mgp_value_make_path, val);
 }
 
-enum mgp_value_type value_get_type(struct mgp_value *val) {
-  return MgInvoke<enum mgp_value_type>(mgp_value_get_type, val);
+struct mgp_value *value_make_date(struct mgp_date *val) {
```
antaljanosbenjamin, line 104: As this file is a C++ file, it is unnecessary to write out all of the `struct` and `enum` prefixes for the parameter types.

cpp/mg_utility/mage.hpp
```diff
@@ -0,0 +1,2757 @@
...
+class Id {
+ public:
+  Id() = default;
+
+  /// Construct Id from uint64_t
+  static Id FromUint(uint64_t id) { return Id(util::MemcpyCast<int64_t>(id)); }
+  /// Construct Id from int64_t
+  static Id FromInt(int64_t id) { return Id(id); }
+
+  int64_t AsInt() const { return id_; }
```
jbajic, line 80: This can be noexcept
```suggestion
  int64_t AsInt() const noexcept { return id_; }
```

cpp/mg_utility/mage.hpp
```diff
@@ -0,0 +1,2757 @@
...
+class Nodes {
+ public:
+  explicit Nodes(mgp_vertices_iterator *nodes_iterator) : nodes_iterator_(nodes_iterator) {}
+
+  class Iterator {
+   public:
+    friend class Nodes;
+
+    explicit Iterator(mgp_vertices_iterator *nodes_iterator);
+    ~Iterator();
```
jbajic, line 150: Do you need a destructor? If you are releasing memory, then yes I presume, but then also define copy ctor and move ctor and copy assignment operator and move assignment operator

cpp/mg_utility/mage.hpp
```diff
@@ -0,0 +1,2757 @@
...
+    Iterator &operator++();
+    Iterator operator++(int);
+    bool operator==(Iterator other) const;
+    bool operator!=(Iterator other) const { return !(*this == other); }
+    Node operator*();
```
jbajic, line 155: Also, define traits above methods declarations, plus I think you need to return the reference to the underlying element since you might want it to be mutable from iterator point of view
```suggestion
    reference operator*();
```

cpp/mg_utility/mage.hpp
```diff
@@ -0,0 +1,2757 @@
...
+    // Traits
+    using difference_type = Node;
```
jbajic, line 157: This should be the result of subtracting two pointers (https://en.cppreference.com/w/cpp/iterator/iterator_traits)
```suggestion
    using difference_type = std::ptrdiff_t;
```

cpp/mg_utility/mage.hpp
```diff
@@ -0,0 +1,2757 @@
...
+class GraphRelationships {
+ public:
+  explicit GraphRelationships(mgp_graph *graph) : graph_(graph) {}
+
+  class Iterator {
+   public:
+    friend class GraphRelationships;
+
+    explicit Iterator(mgp_vertices_iterator *nodes_iterator);
+    ~Iterator();
```
jbajic, line 187: Same as above, the rule of 5

cpp/mg_utility/mage.hpp
```diff
@@ -0,0 +1,2757 @@
...
+    Iterator &operator++();
+    bool operator==(Iterator other) const;
+    bool operator!=(Iterator other) const { return !(*this == other); }
+    Relationship operator*();
+    // Traits
+    using difference_type = Relationship;
```
jbajic, line 193:
```suggestion
    using difference_type = std::ptrdiff_t;
```

cpp/mg_utility/mage.hpp
```diff
@@ -0,0 +1,2757 @@
...
+    using iterator_category = std::forward_iterator_tag;
+
+   private:
+    mgp_vertices_iterator *nodes_iterator_ = nullptr;
+    mgp_edges_iterator *out_relationships_iterator_ = nullptr;
+    size_t index_ = 0;
+  };
+
+  Iterator begin();
+  Iterator end();
```
jbajic, line 206:
```suggestion
  Iterator begin() const;
  Iterator end() const;
```

cpp/mg_utility/mage.hpp
```diff
@@ -0,0 +1,2757 @@
...
+    using reference = const Node &;
+    using iterator_category = std::forward_iterator_tag;
+
+   private:
+    mgp_vertices_iterator *nodes_iterator_ = nullptr;
+    size_t index_ = 0;
+  };
+
+  Iterator begin();
+  Iterator end();
```
jbajic, line 169:
```suggestion
  Iterator begin() const;
  Iterator end() const;
```

cpp/mg_utility/mage.hpp
jbajic: Also I think you should implement `operator->()` to support pointers as well

cpp/mg_utility/mage.hpp
jbajic: You might also want to consider implementing const style iterators as well. I would say this article explains implementation iterators well https://www.internalpointers.com/post/writing-custom-iterators-modern-cpp

cpp/mg_utility/mage.hpp
```diff
@@ -0,0 +1,2757 @@
...
+  size_t size() const { return property_map_.size(); }
+  /// @brief Returns whether the properties map is empty.
+  bool empty() const { return size() == 0; }
+
+  /// @brief Returns the value associated with the given `key`. If there’s no such value, the behavior is undefined.
+  /// @note Each key-value pair needs to be checked, ensuing O(n) time complexity.
+  Value operator[](const std::string_view key) const;
+
+  std::map<std::string_view, Value>::const_iterator begin() const { return property_map_.begin(); }
+  std::map<std::string_view, Value>::const_iterator end() const { return property_map_.end(); }
```
jbajic, line 263:
```suggestion
  std::map<std::string_view, Value>::const_iterator cbegin() const { return property_map_.cbegin(); }
  std::map<std::string_view, Value>::const_iterator cend() const { return property_map_.cend(); }
```

cpp/mg_utility/mage.hpp
```diff
@@ -0,0 +1,2757 @@
...
+  /// @brief Creates a List from the given vector.
+  explicit List(const std::vector<Value> &values);
+  /// @brief Creates a List from the given vector.
+  explicit List(std::vector<Value> &&values);
+
+  /// @brief Creates a List from the given initializer_list.
+  explicit List(const std::initializer_list<Value> list);
+
+  List(const List &other) : List(other.ptr_) {}
+  List(List &&other) : ptr_(other.ptr_) { other.ptr_ = nullptr; }
```
jbajic, line 356:
```suggestion
  List(List &&other) noexcept : ptr_(other.ptr_) { other.ptr_ = nullptr; }
```

cpp/mg_utility/mage.hpp
```diff
@@ -0,0 +1,2757 @@
...
+
+  List &operator=(const List &other) = delete;
+  List &operator=(List &&other) = delete;
```
jbajic, line 359:
```suggestion
  List &operator=(List &&other) noexcept = delete;
```

cpp/mg_utility/mage.hpp
```diff
@@ -0,0 +1,2757 @@
...
+    ptr_ = mgp::value_make_duration(duration.ptr_);
+    delete &duration;
+    duration.ptr_ = nullptr;
+  }
+
+  Value(const Value &other) : Value(other.ptr_) {}
+  Value(Value &&other) : ptr_(other.ptr_) { other.ptr_ = nullptr; }
+
+  // TODO check:
+  // Value &operator=(const Value &other) { return *this; }
```
jbajic, line 1044: Can this TODO be done?

cpp/mg_utility/mage.hpp
```diff
@@ -0,0 +1,2757 @@
...
+  }
+  return i;
+}
+
+int64_t Graph::size() const {
+  int64_t i = 0;
+  for (const auto &_ : relationships()) {
+    i++;
+  }
+  return i;
```
jbajic, line 1600: Can we use `std::distance` https://en.cppreference.com/w/cpp/iterator/distance?

cpp/mg_utility/mage.hpp
```diff
@@ -0,0 +1,2757 @@
...
+      return *this;
+    }
+    index_++;
+  }
+  return *this;
+}
+
+inline Nodes::Iterator Nodes::Iterator::operator++(int) {
+  Nodes::Iterator retval = *this;
+  ++(*this);
```
jbajic, line 1679:
```suggestion
  ++*this;
```

cpp/mg_utility/mage.hpp
```diff
@@ -0,0 +1,2757 @@
...
+    if (node == nullptr) {
+      mgp::vertices_iterator_destroy(nodes_iterator_);
+      nodes_iterator_ = nullptr;
+      return;
+    }
+
+    // Check if node has out-relationships
+    out_relationships_iterator_ = mgp::vertex_iter_out_edges(node, memory);
+    auto relationship = mgp::edges_iterator_get(out_relationships_iterator_);
+    if (relationship != nullptr) return;
```
jbajic, line 1728:
```suggestion
    if (relationship != nullptr) {
      return;
    }
```

cpp/mg_utility/mage.hpp
```diff
@@ -0,0 +1,2757 @@
...
+
+      // Check if node has out-relationships
+      out_relationships_iterator_ = mgp::vertex_iter_out_edges(node, memory);
+      auto relationship = mgp::edges_iterator_get(out_relationships_iterator_);
+      if (relationship != nullptr) {
+        return *this;
+      } else {
+        mgp::edges_iterator_destroy(out_relationships_iterator_);
+        out_relationships_iterator_ = nullptr;
+      }
```
jbajic, line 1778: No need for else if return
```suggestion
      if (relationship != nullptr) {
        return *this;
      }
      mgp::edges_iterator_destroy(out_relationships_iterator_);
      out_relationships_iterator_ = nullptr;
```

cpp/mg_utility/mage.hpp
```diff
@@ -0,0 +1,2757 @@
...
+  }
+}
+
+inline Map::~Map() {
+  if (ptr_ != nullptr) {
+    mgp::map_destroy(ptr_);
+  }
+}
+
+inline const Value Map::operator[](std::string_view key) const { return Value(mgp::map_at(ptr_, key.data())); }
```
jbajic, line 1981: hmm this usually cannot be const because `operator[]` serves as both for setting values and viewing, if it is const it loses the purpose of setting value

jbajic (commented):
I would like to see unit tests for the whole interface, since it seems complex and there are a lot of stuff written here

cpp/mg_utility/mgp.hpp
antepusic: Pushed :heavy_check_mark:

cpp/mg_utility/mage.hpp
```diff
@@ -0,0 +1,2757 @@
...
+
+  /// @brief Returns the graph node with the given ID.
+  Node GetNodeById(const Id node_id) const;
+
+  /// @brief Returns whether the graph contains a node with the given ID.
+  bool Contains(const Id node_id) const;
+  /// @brief Returns whether the graph contains the given node.
+  bool Contains(const Node &node) const;
+  /// @brief Returns whether the graph contains the given relationship.
+  bool Contains(const Relationship &relationship) const;
```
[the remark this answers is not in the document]
antepusic: I chose to rename the methods like you suggested, with the caveat that there’s no `ContainsRelationship(const Id relationship_id)` because of the corresponding function (`mgp_graph_get_edge_by_id`) not existing in the C API.

cpp/mg_utility/mage.hpp
antepusic: Done

cpp/mg_utility/mage.hpp
antepusic: Thanks! I’ve actually studied this article while working on this API :slightly_frowning_face:

cpp/mg_utility/mage.hpp
antepusic: NB: This goes for both `List` and `Map`.

As setting map values is highly non-trivial in the existing C API, I’ve chosen to mark this operator as const and thus prohibit value-setting. This does make it equivalent to `at`, but I believe this API is better off with it because people are used to accessing maps with `[]`.

cpp/mg_utility/mage.hpp
antepusic: Done (for all iterators where this is defined)

cpp/mg_utility/mage.hpp
antepusic: Should there be `noexcept` on the above method (`List &operator=(const List &other) = delete;`) too?

cpp/mg_utility/mage.hpp
antepusic: Done

cpp/mg_utility/mage.hpp
antepusic: Done

cpp/mg_utility/mage.hpp
antepusic: Done (and elsewhere)

cpp/mg_utility/mage.hpp
antepusic: Why mark a deleted operator as `noexcept`? If it can’t be executed, it shouldn’t be able to throw exceptions either.

cpp/mg_utility/mage.hpp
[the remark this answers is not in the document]
antepusic: Removed

cpp/mg_utility/mage.hpp
antepusic: With iterators, `operator->` returns the private pointer used by the iterator. 
In this case, that pointer is a MGP type `mgp_vertices_iterator *`, and returning it  would defeat the purpose of this API, which is abstracting MGP types away from the user. Therefore, I’ve left it unimplemented.

For the same reason, `operator*` returns `Node` instead of the reference type. `mgp_vertices_iterator *` contains MGP vertices which need to be converted into new API types.

I understand this is not conventional usage, but it is what had to be implemented to allow iteration with new API types.

cpp/mg_utility/mage.hpp
antepusic: Done

cpp/mg_utility/mage.hpp
antepusic: Done

cpp/mg_utility/mage.hpp
antepusic: Done

cpp/mg_utility/mage.hpp
antepusic: Done

cpp/mg_utility/mage.hpp
antepusic: `std::distance` keeps incrementing the iterator until it reaches the `.end()`; with some iterators as this one this causes an error because incrementing them doesn’t eventually reach it (`.end()` is defined as returning `nullptr`)

cpp/mg_utility/mage.hpp
```diff
@@ -0,0 +1,2771 @@
...
+
+  /// @brief Creates a List from the given vector.
+  explicit List(const std::vector<Value> &values);
+  /// @brief Creates a List from the given vector.
+  explicit List(std::vector<Value> &&values);
+
+  /// @brief Creates a List from the given initializer_list.
+  explicit List(const std::initializer_list<Value> list);
+
+  List(const List &other) : List(other.ptr_) {}
```
antaljanosbenjamin, line 571: If you copy owning pointers, then you will end up in double free situations. I haven't spotted this on the mage repo, only as I checked a memory leak that @Josipmrden just fixed. This is true for all copy constructors. In my opinion best solution is to call `mgp_xxx_copy` to copy the underlying object.

jbajic (approved):

gitbuda:
@Josipmrden and @antaljanosbenjamin I think this should be closed because we move this part into the memgraph repo -> https://github.com/memgraph/memgraph/pull/546?

antoniofilipovic:
Closing this PR as I opened a new one for integration, and this was part of the development process which was moved to Memgraph.

PR #162, opened 2022-08-19
Degree centrality algorithm

jmatak:
### Description

Implementation of degree centrality algorithm.  More information about algorithm: https://www.sci.unich.it/~francesc/teaching/network/degree.html

### Pull request type

- [x] Algorithm/Module

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [x] Code documentation


antepusic (approved):

cpp/degree_centrality_module/degree_centrality_module.cpp
```diff
@@ -0,0 +1,59 @@
...
+  return 0;
+}
+
+// This is an optional function if you need to release any resources before the
+// module is unloaded. You will probably need this if you acquired some
+// resources in mgp_init_module.
+extern "C" int mgp_shutdown_module() {
+  // Return 0 to indicate success.
+  return 0;
+}
```
antoniofilipovic, line 59: Empty line here

cpp/degree_centrality_module/degree_centrality_module.cpp
```diff
@@ -0,0 +1,59 @@
...
+const char *kFieldNode = "node";
+const char *kFieldDegree = "degree";
+
+void InsertDegreeCentralityRecord(mgp_graph *graph, mgp_result *result, mgp_memory *memory, std::uint64_t node,
+                                  double degree) {
+  auto *record = mgp::result_new_record(result);
+
+  mg_utility::InsertNodeValueResult(graph, record, kFieldNode, node, memory);
+  mg_utility::InsertDoubleValueResult(record, kFieldDegree, degree, memory);
+}
```
antoniofilipovic, line 20: Why don't we insert node here instead of node_id?

Because this function `InsertNodeValueResult` needs to get node for every id, which means CAPI call to DbAccessor to get node with given ID which is fast with indexing but slow otherwise (but still NlogN), right?

antoniofilipovic: We can make it faster by sending here pointer to node, if I am not wrong?

antoniofilipovic (commented):
Check comments, I was just thinking about how to make it faster (maybe) and a few empty lines are missing.

cpp/degree_centrality_module/degree_centrality_module.cpp
jmatak: You are not wrong about the time complexity, however, we do not have a node there, we are storing the reindexed nodes in our GraphView, and only work with `int` type. This is the step where we fetch the concrete node from storage.

jmatak: Also true, but the same logic could be applied to any other algorithm out there. There are two reasons to have GraphView API:
1. It's easier to use. It's much more comfortable to use this API than directly using the C API that Memgraph offers.
2. It can be tested and developed independently. You don't need a Memgraph instance to test the GraphView. Tests can be generated without having the Memgraph instance running.

antoniofilipovic: I think the same logic cannot be applied since in other algorithms we are doing more complex traversals and it is easier to test and implement algorithms that way, but here we have directly what we need through API. We can make it faster by avoiding copying stuff to GraphView and getting every vertex from the database (mgp_graph_get_vertex_by_id). And instead of the whole algorithm and switching to GraphView, we can just use API. But, if testing is still an issue, I am fine with this.

antoniofilipovic: The only issue could be the subgraph, but we can skip that part of the implementation, right? Since that feature should be on master by 31.8.

antoniofilipovic: I think this is resolved now.

antoniofilipovic (approved):
:ship: :it:

PR #170, opened 2022-10-04
Query module for connecting to ElasticSearch

as51340:
### Description

Query module which enables developers to serialize Memgraph into Elastic search instance using basic authentication. It can also scan based on the query given. It indexes documents using bulking_scan. 

### Pull request type

- [ ] Bugfix
- [ ] Algorithm/Module
- [x] Feature
- [ ] Build related changes

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation


python/elastic_search_serialization.py
```diff
@@ -0,0 +1,505 @@
...
+        elastic_password (str): User's password for connecting to the Elasticsearch.
+    Returns:
+        Elasticsearch: Client reference.
+    """
+    client = Elasticsearch(
+        elastic_url,
+        ca_certs=ca_certs,
+        basic_auth=(elastic_user, elastic_password),
+    )
+    print(f"Client info: {client.info()}")
```
antoniofilipovic, line 260: Consider using mgp.Logger here or completely removing the print.

python/elastic_search_serialization.py
```diff
@@ -0,0 +1,505 @@
...
+        number_of_replicas=number_of_replicas,
+    )
+
+    # Created objects can be vertices and edges
+    created_vertices, created_edges = get_created_updated_objects(createdObjects)
+
+    # Now create iterable of documents that need to be indexed
+    nodes, edges = generate_documents(created_vertices, created_edges)
+    print(f"Nodes: {nodes}")
+    print(f"Edges: {edges}")
```
antoniofilipovic, line 400: Consider using mgp.Logger here or completely remove the print.

antoniofilipovic: Same for all the other prints.

python/elastic_search_serialization.py
```diff
@@ -0,0 +1,505 @@
...
+        scroll=scroll,
+        raise_on_error=raise_on_error,
+        preserve_order=preserve_order,
+        size=size,
+        request_timeout=request_timeout,
+        clear_scroll=clear_scroll,
+    )
+    items = []
+    for item in response:
+        items.append(item)
```
antoniofilipovic, line 504: items = [item for item in items]

antoniofilipovic: even better
return [item for item in items]

python/elastic_search_serialization.py
```diff
@@ -0,0 +1,505 @@
...
+    Args:
+        edge (mgp.Edge): Reference to the edge in Memgraph DB.
+    Returns:
+        Dict[str, Any]: ElasticSearch object representation.
+    """
+    rel: Dict[str, Any] = {}
+    rel[ACTION] = {INDEX: {ID: str(edge.id)}}
+    source = serialize_properties(edge.properties.items())
+    source[LK_TYPE] = edge.type.name
+    rel[SOURCE] = source
```
antoniofilipovic, line 220: source = serialize_properties(edge.properties.items())
return {ACTION:INDEX: {ID: str(edge.id)},
        LK_TYPE : edge.type.name,
        SOURCE: source }

python/elastic_search_serialization.py
```diff
@@ -0,0 +1,505 @@
...
+    return rel
```
antoniofilipovic, line 221:
```suggestion
    source = serialize_properties(edge.properties.items())
    return {ACTION: {INDEX: {ID: str(edge.id)}},
                SOURCE: source,
                LK_TYPE: edge.type.name}
```

python/elastic_search_serialization.py
```diff
@@ -0,0 +1,505 @@
...
+    Args:
+        elastic_url (str): URL for connecting to the Elasticsearch instance.
+        ca_certs (str): Path to the certificate file.
+        elastic_user (str): The user trying to connect to the Elasticsearch.
+        elastic_password (str): User's password for connecting to the Elasticsearch.
+    Returns:
+        Elasticsearch: Client reference.
+    """
+    client = Elasticsearch(
+        elastic_url,
```
antoniofilipovic, line 256: For consistency, consider setting elastic_url as key=value pair.

python/elastic_search_serialization.py
```diff
@@ -0,0 +1,505 @@
...
+        context_edges (List[mgp.Edge]]): Edges in Memgraph that were created/updated.
+    Returns:
+        Tuple[List[Dict[str, Any]], List[Dict[str, Any]]]
+    """
+    nodes, edges = [], []
+    for vertex in context_vertices:
+        nodes.append(serialize_vertex(vertex))
+    for edge in context_edges:
+        edges.append(serialize_edge(edge))
+    return nodes, edges
```
antoniofilipovic, line 326: Can we parallelize this? Does it make sense?

python/elastic_search_serialization.py
```diff
@@ -0,0 +1,505 @@
...
+        vertex (mgp.Vertex): Reference to the vertex in Memgraph DB
+    Returns:
+        Dict[str, Any]: ElasticSearch object representation.
+    """
+    node: Dict[str, Any] = {}  # whole part
+    node[ACTION] = {INDEX: {ID: str(vertex.id)}}
+    source = serialize_properties(vertex.properties.items())
+    source[LK_CATEGORIES] = [label.name for label in vertex.labels]
+    node[SOURCE] = source
+    return node
```
antoniofilipovic, line 205: You have different namings for the same thing, consider using vertex or node, not both. And same suggestion as below may be applied.

antoniofilipovic (changes requested):

python/elastic_search_serialization.py
as51340: Hmm not sure I think yes. Multiprocessing in Python. But this requires rewriting because I don't think setting up more than a multiprocessing pool makes sense.

python/elastic_search_serialization.py
```diff
@@ -0,0 +1,514 @@
...
+        Dict[str, Any]: Object that conforms ElasticSearch's schema.
+    """
+    source: Dict[str, Any] = {}
+    for prop_key, prop_value in properties:
+        if type(prop_value) == datetime:
+            # Convert datetime to str, replace microsecond and add Z suffix(Zulu or zero offset) manually because Python doesn't support it out of the box
+            prop_value = prop_value.replace(microsecond=0).isoformat() + "Z"
+            source[prop_key + LKE_DATE] = prop_value
+        elif type(prop_value) in lke_mapping:
+            source[prop_key + lke_mapping[type(prop_value)]] = prop_value
```
BorisTasevski, line 235: Use formatted string where possible instead of string concatenation

python/elastic_search_serialization.py
```diff
@@ -0,0 +1,514 @@
...
+        properties (Dict[str, Any]]): Properties of nodes and relationships.
+    Returns:
+        Dict[str, Any]: Object that conforms ElasticSearch's schema.
+    """
+    source: Dict[str, Any] = {}
+    for prop_key, prop_value in properties:
+        if type(prop_value) == datetime:
+            # Convert datetime to str, replace microsecond and add Z suffix(Zulu or zero offset) manually because Python doesn't support it out of the box
+            prop_value = prop_value.replace(microsecond=0).isoformat() + "Z"
+            source[prop_key + LKE_DATE] = prop_value
```
BorisTasevski, line 233: Use formatted string where possible instead of string concatenation

python/elastic_search_serialization.py
```diff
@@ -0,0 +1,514 @@
...
+    """
+    Generates nodes and edges documents for indexing and returns them as lists.
+    Args:
+        context_vertices (List[mgp.Vertex]]): Vertices in Memgraph that were created/updated.
+        context_edges (List[mgp.Edge]]): Edges in Memgraph that were created/updated.
+    Returns:
+        Tuple[List[Dict[str, Any]], List[Dict[str, Any]]]
+    """
+    nodes, edges = [], []
+    for vertex in context_vertices:
```
BorisTasevski, line 327: use list comprehension here

python/elastic_search_serialization.py
```diff
@@ -0,0 +1,514 @@
...
+        nodes.append(serialize_vertex(vertex))
+    for edge in context_edges:
+        edges.append(serialize_edge(edge))
```
BorisTasevski, line 330: use list comprehension here

python/elastic_search_serialization.py
```diff
@@ -0,0 +1,514 @@
...
+    Generates nodes and edges documents for indexing and returns them as lists.
+    Args:
+        context_vertices (List[mgp.Vertex]]): Vertices in Memgraph that were created/updated.
+        context_edges (List[mgp.Edge]]): Edges in Memgraph that were created/updated.
+    Returns:
+        Tuple[List[Dict[str, Any]], List[Dict[str, Any]]]
+    """
+    nodes, edges = [], []
+    for vertex in context_vertices:
+        nodes.append(serialize_vertex(vertex))
```
BorisTasevski, line 328:
```suggestion
        nodes = [serialize_vertex(vertex) for vertex in context_vertices]
```

python/elastic_search_serialization.py
```diff
@@ -0,0 +1,514 @@
...
+    elastic_password: str,
+    source_index: mgp.Any,
+    target_index: str,
+    query: str,
+    chunk_size: int = 500,
+    scroll: str = "5m",
+    op_type: mgp.Nullable[str] = None,
+) -> mgp.Record(response=str):
+    """
+    Reindex all documents from one index that satisfy a given query to another, potentia    lly (if target_client is specified) on a different cluster. If you don’t specify the    query you will reindex all the documents.
```
BorisTasevski, line 435: Remove extra spaces from string

python/elastic_search_serialization.py
```diff
@@ -0,0 +1,514 @@
...
+        chunk_size (int): number of docs in one chunk sent to es (default: 500).
+        scroll (str): Specify how long a consistent view of the index should be maintained for scrolled search.
+        op_type (Optional[str]): Explicit operation type. Defaults to ‘_index’. Data streams must be set to ‘create’. If not specified, will auto-detect if target_index is a data stream.
+    Returns:
+        response (str): Response of the query.
+    """
+    client = connect_to_elasticsearch(
+        elastic_url, ca_certs, elastic_user, elastic_password
+    )
+    query_obj = json.loads(query)
```
BorisTasevski, line 455: is this local object needed?

python/elastic_search_serialization.py
```diff
@@ -0,0 +1,514 @@
...
+        elastic_password (str): User's password for connecting to the Elasticsearch.
+        index_name (str): A name of the index.
+        query (str): Query written as JSON.
+    Returns:
+         mgp.Record(items=mgp.List[mgp.Map]): List of all items matched by the specific query.
+    """
+    client = connect_to_elasticsearch(
+        elastic_url, ca_certs, elastic_user, elastic_password
+    )
+    query_obj = json.loads(query)
```
BorisTasevski, line 501: is this local object needed?

python/elastic_search_serialization.py
```diff
@@ -0,0 +1,514 @@
...
+        query=query_obj,
+        index=index_name,
+        scroll=scroll,
+        raise_on_error=raise_on_error,
+        preserve_order=preserve_order,
+        size=size,
+        request_timeout=request_timeout,
+        clear_scroll=clear_scroll,
+    )
+    items = [item for item in response]
```
BorisTasevski, line 513: is this local object needed?

BorisTasevski (changes requested):

python/elastic_search_serialization.py
```diff
@@ -0,0 +1,514 @@
...
+lke_mapping[int] = LKE_NUMBER
+lke_mapping[float] = LKE_NUMBER
+lke_mapping[bool] = LKE_BOOLEAN
+lke_mapping[datetime] = LKE_DATE
+
+# Create global logger object
+logger = mgp.Logger()
+
+# Create node index
+node_index_body: Dict[str, Any] = {
```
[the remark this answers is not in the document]
as51340: Yes, the new version decouples this in a way that the user just specifies the path to the schema. Because also logically doesn't make sense to check if indexes exist every time he/she wants to index some document.

python/elastic_search_serialization.py
as51340: The new version modified for loops in order to optimize the code and possibly introduce multiprocessing so no need for list comprehension anymore.

python/elastic_search_serialization.py
```diff
@@ -0,0 +1,514 @@
...
+            },
+            {"date": {"match": "*_lke_date", "mapping": {"type": "date"}}},
+            {"number": {"match": "*_lke_number", "mapping": {"type": "double"}}},
+            {"boolean": {"match": "*_lke_boolean", "mapping": {"type": "keyword"}}},
+        ]
+    },
+}
+
+# Create node index
+edge_index_body: Dict[str, Any] = {
```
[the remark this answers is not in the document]
as51340: Same, the user will be able to specify a path to the schema.

python/elastic_search_serialization.py
```diff
@@ -0,0 +1,514 @@
...
+    """
+    Serializes edge to specified ElasticSearch schema.
+    Args:
+        edge (mgp.Edge): Reference to the edge in Memgraph DB.
+    Returns:
+        Dict[str, Any]: ElasticSearch object representation.
+    """
+    source = serialize_properties(edge.properties.items())
+    source[LK_TYPE] = edge.type.name
+    return {ACTION: {INDEX: {ID: str(edge.id)}}, SOURCE: source}
```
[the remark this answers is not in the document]
as51340: In my opinion not, because we need that to get JSON representation so dictionaries are very useful for this and also such a class would never be reused but we can discuss.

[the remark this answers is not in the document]
antoniofilipovic: @BorisTasevski ping here for discussion, I am not sure what is better also. But it seems this is the only comment left.

python/elastic_search_serialization.py
```diff
@@ -0,0 +1,474 @@
...
+        index_name (str): Name of the index that needs to be created.
+        schema_path (str): Path to the schema from where it will be loaded.
+        schema_parameters: Dict[str, Any]
+            number_of_shards (int): Number of shards index will use.
+            number_of_replicas (int): Number of replicas index will use.
+        analyzer (str): Custom analyzer, can be set to any legal Elasticsearch analyzer.
+    Returns:
+       mgp.Map: Response message from Elasticsearch service.
+    """
+    global client
```
antoniofilipovic, line 220: Maybe one idea here. Instead of using a `global client`, we could create a `Singleton` class and here get a `client` from that class. And in the method connect, you can initialize that class. This way we could provide unit tests only, but it would be enough to check that the implementation works, right?

antoniofilipovic (commented):
Check the comment for the singleton class, we can discuss how to proceed but I think we need at least unit tests for this. When you implement a singleton class in unit tests we could `mock` elasticsearch. How does this sound to you?

python/elastic_search_serialization.py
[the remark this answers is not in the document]
BorisTasevski: I've expanded my initial comment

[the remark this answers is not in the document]
BorisTasevski: Please address this

python/elastic_search_serialization.py
```diff
@@ -0,0 +1,565 @@
...
+        if isinstance(prop_value, datetime):
+            # Convert datetime to str, replace microsecond and add Z suffix(Zulu or zero offset) manually because Python doesn't support it out of the box
+            prop_value = f"{prop_value.replace(microsecond=0).isoformat()}Z"
+            source[f"{prop_key}{LKE_DATE}"] = prop_value
+        elif type(prop_value) in lke_mapping:
+            source[f"{prop_key}{lke_mapping[type(prop_value)]}"] = prop_value
+    return source
+
+
+def generate_document(context_object):
```
BorisTasevski, line 100: Typing omitted

BorisTasevski (changes requested):

BorisTasevski (approved):

python/elastic_search_serialization.py
```diff
@@ -0,0 +1,622 @@
...
+CREATED_VERTEX = "created_vertex"
+CREATED_EDGE = "created_edge"
+VERTEX = "vertex"
+EDGE = "edge"
+INDEX_TYPE = "index_type"
+AGGREGATIONS = "aggregations"
+HITS = "hits"
+TOTAL = "total"
+
+# Linkurious constants
```
antoniofilipovic, line 35: I think we should remove anything that has connection with clients

antoniofilipovic (approved):
Remove the connection to Linkurious and we are ready to ship it.

python/elastic_search_serialization.py
as51340: Yes, this was left accidently, I removed other things.

PR #187, opened 2022-12-19
Add meta_util.schema procedure that generates graph schema as a graph result

tonilastre:
### Description

Module `meta_util` can be used to generate a graph schema as a graph result.  It can be used like this:

```
// Get schema without properties count
CALL meta_util.schema() YIELD nodes, relationships RETURN nodes, relationships;

// Get schema, but with properties count
CALL meta_util.schema(true) YIELD nodes, relationships RETURN nodes, relationships;
```

It will show graph view only in Lab version >= 2.4.0. For earlier versions of the Lab, `UNWIND` needs to be called on returned object properties `nodes` and `edges`.

### Pull request type

- [x] Algorithm/Module

#### Module/Algorithm

- [ ] Unit tests
- [x] End-to-end tests
- [x] Code documentation
- [x] Documentation on [memgraph/docs](https://github.com/memgraph/docs) - https://github.com/memgraph/docs/pull/646


python/meta.py
```diff
@@ -0,0 +1,148 @@
...
+class Counter:
+    def __init__(self, initial_value: int = 0):
+        self.total_count = initial_value
+        self.count_by_property_name = defaultdict(int)
+
+    def count(self) -> None:
+        self.total_count += 1
+
+    def count_property(self, property_name: str) -> None:
+        self.count_by_property_name[property_name] += 1
```
antepusic, line 20: I believe `increment`/`increment_property` are better names.

antepusic (commented):
All looks OK, just check out the suggestions I left :smile:

python/meta.py
katarinasupe: Are you referring to the procedures' names or variables' names?

antepusic: The procedures

katarinasupe: Sure, makes sense and I agree. I will fix it now :D

antepusic (commented):
Good job!

python/meta.py
```diff
@@ -0,0 +1,151 @@
...
+        Get graph schema without properties count:
+            `CALL meta.schema() YIELD nodes, relationships RETURN nodes, relationships;`
+        Get graph schema with properties count:
+            `CALL meta.schema(true) YIELD nodes, relationships RETURN nodes, relationships;`
+    """
+
+    if len(context.graph.vertices) == 0:
+        raise Exception(
+            "Can't generate a graph schema since there is no data in the database."
+        )
```
antoniofilipovic, line 57: This might be a problem if the graph is huge. Because, what will happen here, is that the iterator needs to go through each node, and that means making calls to the Memgraph database each time to get a vertex, creating one as Python, and destroying afterward. You can do the following here. Instead of this chunk of code, you have for loop already going over nodes. Add some kind of counter inside that loop. Or use enumerate function. If at the end "i" is 0, throw an error then. I know this is better practice, but I am here talking about optimization.

e2e/meta_test/test_schema_nodes_properties/test.yml
```diff
@@ -0,0 +1,19 @@
...
+    name: 2
+    year: 1
+  - count: 2
+    age: 
+    name: 2
+    year:
+  - count: 1
+    age: 
+    name: 1
+    year:
```
antoniofilipovic, line 19: I am not sure if it is possible for this to break, meaning we have "no guarantee" of order, right? Is it possible to somehow guarantee it?

e2e/meta_test/test_schema_nodes_labels/test.yml
```diff
@@ -0,0 +1,8 @@
+query: >
+    CALL meta.schema() YIELD nodes, relationships UNWIND nodes AS node UNWIND node.labels AS label RETURN label;
```
antoniofilipovic, line 2: Also, for this part we don't have guarantee of order, maybe we can add here ORDER BY label DESC and then be sure.

antoniofilipovic (changes requested):
Looks good, but from the optimization side, I think we should make this change of order when checking if the graph is empty, and make sure tests don't break.

python/meta.py
katarinasupe: Cool, makes sense, I will add this :)

e2e/meta_test/test_schema_nodes_properties/test.yml
katarinasupe: I am also not sure, but I also thought about that before, so I think it makes sense to add order to be sure. Gonna do that.

antoniofilipovic (approved):

antepusic (approved):

PR #199, opened 2023-03-02
Add CSV export query results module

katarinasupe:
### Description

I created `csv_query` procedure inside `export_util` module. Find arguments description in docstring. 

TODO / discuss with @Josipmrden:

- [ ] Potentially use `pandas` instead of `csv` - what are the pros?
- [ ] Leave config as a map or add stream as an argument. Another option is to return the stream and remove it from the arguments altogether. I vote for this. 
- [ ] Add tests
- [ ] Add examples in the docstring
- [ ] Write docs 
- [ ] Probably edit the path inside the code to be `/usr/lib/memgraph/query_modules` instead of `/var/lib/memgraph/internal_modules` used for development. In `json()` procedure in this module, I left it to the user to provide the whole path and advised them to provide `/usr/lib/memgraph/query_modules/something.json`. This may be a better way to go.

### Pull request type

- [x] Algorithm/Module

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation

### DIFF
--- a/python/export_util.py
+++ b/python/export_util.py
@@ -1,9 +1,14 @@
+import csv
+import io
+import json as js
+import mgp
+
 from dataclasses import dataclass
 from datetime import datetime, date, time, timedelta
-import json as js
+from gqlalchemy import Memgraph
+from typing import Any, Dict, List, Union
+
 from mage.export_import_util.parameters import Parameter
-import mgp
-from typing import Union, List, Dict, Any
 
 
 @dataclass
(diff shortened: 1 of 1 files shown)


Josipmrden:
As for Pandas, I think it makes it more easier to work with CSV files, but we can check what are the implications of adding another dependency into MAGE. 

As for the second one, I'm fine with both.

3. ok, makes sense

katarinasupe:
I did not benchmark it @antoniofilipovic. `csv` can be used out of the box, while `pandas` must be added as a new dependency to MAGE.

PR #203, opened 2023-03-28
Conditional execution procedures

antepusic:
### Description

This PR adds a query module that allows conditional execution of read-only procedures.

### Pull request type

- [x] Algorithm/Module

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation


python/do.py
```diff
@@ -0,0 +1,199 @@
...
+
+    return [
+        mgp.Record(
+            value={
+                field_name: _gqlalchemy_type_to_mgp(ctx.graph, field_value)
+                for field_name, field_value in result.items()
+            }
+        )
+        for result in results
+    ]
```
as51340, line 113: And the private method can then be reused here

python/do.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+def _get_edge_by_id(
```
as51340, line 116: Maybe a bit better naming like get edge from starting node or something similar because my first impression when I see such name is that everything will be searched.

python/do.py
```diff
@@ -0,0 +1,199 @@
...
+            graph, edge_id=int(variable._id), from_id=int(variable._start_node_id)
+        )
+
+    elif isinstance(variable, gqlalchemy.models.Path):
+        start_id = int(variable._nodes[0]._id)
+        path = mgp.Path(graph.get_vertex_by_id(start_id))
+
+        for relationship, node in zip(variable._relationships, variable._nodes[1:]):
+            next_edge_id = int(relationship._id)
+            next_node_id = int(node._id)
```
as51340, line 161: I don't think there is need for a local variable since it is used at only one place.

python/do.py
```diff
@@ -0,0 +1,199 @@
...
+
+            edge_to_add = _get_edge_by_id(graph, edge_id=next_edge_id, from_id=start_id)
+            if edge_to_add is None:
+                # This should happen if and only if the query that returned the path treated graph edges as undirected.
+                # For example, such a query might return an (a)-[b]->(c) path from a graph that only contains (c)->[b].
+                # In that case, flipping the path direction to retrieve the linking edge is valid.
+                edge_to_add = _get_edge_by_id(
+                    graph, edge_id=next_edge_id, from_id=next_node_id
+                )
```
as51340, line 170:
```suggestion
                edge_to_add = _get_edge_by_id(
                    graph, edge_id=next_edge_id, from_id=int(node_.id)
                )
```

python/do.py
```diff
@@ -0,0 +1,199 @@
...
+    elif isinstance(variable, gqlalchemy.models.Path):
+        start_id = int(variable._nodes[0]._id)
+        path = mgp.Path(graph.get_vertex_by_id(start_id))
+
+        for relationship, node in zip(variable._relationships, variable._nodes[1:]):
+            next_edge_id = int(relationship._id)
+            next_node_id = int(node._id)
+
+            edge_to_add = _get_edge_by_id(graph, edge_id=next_edge_id, from_id=start_id)
+            if edge_to_add is None:
```
as51340, line 164: I am not really sure I understand when can this happen. If we are always exploring path from starting node and searching out edges why would we ever want to search reverse edges? Can you explain this a bit, I am sure you had a good reason for this.

python/do.py
```diff
@@ -0,0 +1,199 @@
...
+
+# Conditional queries are run via gqlalchemy. The module supports parameterized
+# queries, which are only supported from gqlalchemy 1.4.0. Since that version’s
+# dgl requirement causes arm64 checks for MAGE to fail, this function serves to
+# support running parameterized queries with older gqlalchemy versions.
+#
+# Relevant code:
+# https://github.com/memgraph/gqlalchemy/blob/main/gqlalchemy/vendors/database_client.py#L54-L59
+# https://github.com/memgraph/gqlalchemy/blob/main/gqlalchemy/connection.py#L87-L100
+def _execute_and_fetch_parametrized(
```
as51340, line 187: This is nice that you implemented this since you needed for the debugging but I think we cannot use this since this gqlalchemy code. Did you check if python3.8 supports DGL on arm? Maybe that's not even an issue, it was just an idea that popped into my mind. Did you also check how is DGL built? If I am not wrong we added some part to be built from source. Because I really think is conceptually wrong to merge gqlalchemy functionalities with mage modules

as51340: Also, please mention this on the daily so we can discuss it. I think this gives us enough evidence that we need a unified Python version on all Memgraph related things -> gqla, mage and Memgraph, otherwise there will be whole mess of dependencies in the future.

as51340 (changes requested):
Nice work, very clean solution I really don't see many things that could be improved. However, some e2e tests are failing so you chekc this + please mention the comment about built, Python and gqla on the daily, we should discuss that.

python/do.py
```diff
@@ -0,0 +1,199 @@
...
+        (
+            edge
+            for edge in graph.get_vertex_by_id(from_id).out_edges
+            if edge.id == edge_id
+        ),
+        None,
+    )
+
+
+def _gqlalchemy_type_to_mgp(graph: mgp.Graph, variable: Any) -> mgp.Any:
```
as51340, line 129: Just a comment to double check these are all mgp supported types, just to be sure that something isn't forgotten accidentally.

e2e/do_test/do_case_group/test_params/test.yml
```diff
@@ -0,0 +1,10 @@
+query: >
+    CALL do.case([TRUE, "MATCH (n {id: $id}) RETURN n;"],
```
as51340, line 2: Is it possible to have multiple CALLs within one query? Like you do CALL do.case bla bla you get some output values and can those values be send further into other query modules? Not sure, just a question.

as51340:
I think it would be good if you can provide some real queries for conditions not just true/false so we have "real" e2e example, e.g match on some person check some condition and then do something with it.

python/do.py
```diff
@@ -0,0 +1,199 @@
...
+from typing import Any, Dict, Iterator
+
+import mgp
+import gqlalchemy
+from gqlalchemy.connection import _convert_memgraph_value
+
+WRONG_STRUCTURE_MSG = "The `do_case` argument named `conditionals` must be structured as follows: [BOOLEAN, STRING, BOOLEAN, STRING, …​]."
+
+
+@mgp.read_proc
```
as51340, line 11: I think we have a problem with @mgp.read_proc because that means we don't support write queries. Did you maybe intentionally left it?

as51340:
Also, I think it would be good to test the behavior when there are no output values from do.case and do.when, e.g you user enter do.case and provides some query which doesn't yield any values so those empty values are sent into the other module/query and shouldn't do anything because there are none of them.

python/do.py
as51340:
```cypher
MATCH (node:Node {id:12345})
FOREACH (i in CASE WHEN node.needsUpdate THEN [1] ELSE [] END |
  SET node.newProperty = 5678
  REMOVE node.needsUpdate
  SET node:Updated)
...
```
something like this

as51340:
Also, can you the behavior of checking the string property value, when it has quotes?

Like this:
```cypher
MATCH (n) WHERE n.id = "hello" RETURN n;
```

python/do.py
antepusic: The queries passed to the `case`/`when` functions can be either read or write queries because gqlalchemy opens a new connection independent of the procedure (the `memgraph = gqlalchemy.Memgraph()` line).

python/do.py
antepusic: They are the Python equivalents of all mgp-supported types ([reference](https://github.com/memgraph/memgraph/blob/master/include/mgp.py#L1377-L1395)).

python/do.py
antepusic: Cypher lets you match paths without taking edge direction into account ([example](https://neo4j.com/docs/cypher-manual/current/introduction/uniqueness/#_homomorphism)) even though database edges are all directed by design.
In that case, this piece of code searches for the edge in both directions.

python/do.py
antepusic: The docs don’t mention anything about ARM; I’ve only found that people have run dgl on ARM machines. Where do I find how we build dgl?

antepusic: I agree with you about it being conceptually wrong to merge gqlalchemy features with mage, but this workaround is supposed to be there only until https://github.com/memgraph/gqlalchemy/issues/227 is solved.

antepusic:
@as51340:

> https://github.com/memgraph/mage/pull/203#pullrequestreview-1363429927
Will do. The check only failed because igraph e2e tests don’t make sure the database is empty - I’ll update the e2e script to do that before each test.

> https://github.com/memgraph/mage/pull/203#issuecomment-1488871668
No problem - I only omitted such tests because evaluating the condition to true/false is the query engine’s job.

> https://github.com/memgraph/mage/pull/203#issuecomment-1488876204
Good idea, I’ll add a test :+1: 

> https://github.com/memgraph/mage/pull/203#issuecomment-1488882971
Do you mean strings that contain quote marks inside them?

python/do.py
as51340: You can check yml files in workflow directory and Dockerfile

as51340: Do you know how this issue should be resolved? How can we make optional dependencies

antepusic: The GQLAlchemy team is already aware of the optional dependency issue :heavy_check_mark:

python/do.py
antepusic: It’s used twice (see `start_id = next_node_id`)

python/do.py
```diff
@@ -0,0 +1,199 @@
...
+
+    return [
+        mgp.Record(
+            value={
+                field_name: _gqlalchemy_type_to_mgp(ctx.graph, field_value)
+                for field_name, field_value in result.items()
+            }
+        )
+        for result in results
+    ]
```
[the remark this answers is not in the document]
antepusic: Done

python/do.py
antepusic: :heavy_check_mark:

python/do.py
antepusic: Done

python/do.py
antepusic: See above (variable used more than once)

python/do.py
antepusic: Since we concluded at the daily that this isn’t a blocker, can this be resolved?
When GQLAlchemy with optional dependencies is shipped, I can make a quick patch to remove this workaround.

e2e/do_test/do_case_group/test_params/test.yml
antepusic: That’s possible in general :slightly_smiling_face:

python/utils/subquery.py
```diff
@@ -0,0 +1,13 @@
...
+    "CREATE CONSTRAINT ON",
+    "DROP CONSTRAINT ON",
+    "SET GLOBAL TRANSACTION ISOLATION LEVEL",
+    "STORAGE MODE IN_MEMORY_TRANSACTIONAL",
+    "STORAGE MODE IN_MEMORY_ANALYTICAL",
+]
+
+
+def is_global_operation(subquery: str) -> bool:
+    return any(subquery.startswith(pattern) for pattern in QUERY_PATTERNS)
```
as51340, line 13: I think we should support also lower case

python/do.py
as51340: in my opinion yes, when gqla is fixed, patch here should suffice

as51340:
I would like to add one more thing if you think it makes sense and that is testing whole vertex and relationship objects return result, not just size of the returned results so we are sure vertex and relationships are correctly returned from gqla to mage.

python/utils/subquery.py
antepusic: Added

antepusic:
@as51340:
> https://github.com/memgraph/mage/pull/203#issuecomment-1490599098

This makes sense, added one test each for nodes, relationships and paths that make use of their peculiarities (node properties, `type()` for relationships and `nodes()` for paths).

as51340 (approved):

antoniofilipovic (approved):

damianbenente:
[BUG] Failed do.when() when using database with authentication
https://github.com/memgraph/mage/issues/214

PR #221, opened 2023-06-12
[master < T221] Periodic iterate module

Josipmrden:
### Description

In the pull request, periodic.iterate module has been added to support periodic transactions over a query. It helps when having to allocate a lot of deltas that need to be read all over again during the each passing of the operator tree.

### Pull request type

- [X] Algorithm/Module

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation


cpp/periodic_module/periodic_iterate.cpp
```diff
@@ -0,0 +1,241 @@
...
+  mgp::memory = memory;
+  const auto arguments = mgp::List(args);
+  const auto record_factory = mgp::RecordFactory(result);
+  auto record = record_factory.NewRecord();
+
+  const auto input_query = std::string(arguments[0].ValueString());
+  const auto running_query = std::string(arguments[1].ValueString());
+  const auto config = arguments[2].ValueMap();
+
+  const auto batch_size_value = config.At(kConfigKeyBatchSize);
```
antoniofilipovic, line 163: `At` can throw and in this case it is not covered by any `try`. I would rather check with `contains` if `kConfigKeyBatchSize` is inside and throw something more with graceful message

cpp/periodic_module/periodic_iterate.cpp
```diff
@@ -0,0 +1,241 @@
...
+
+    const auto batch_size = batch_size_value.ValueInt();
+
+    if (batch_size <= 0) {
+      throw std::runtime_error("Batch size must be a non-negative number!");
+    }
+
+    mg::Client::Init();
+
+    mg::Client::Params params{.host = "localhost", .port = 7687};
```
antoniofilipovic, line 181: Can this be configurable?

cpp/periodic_module/periodic_iterate.cpp
```diff
@@ -0,0 +1,241 @@
...
+    batch.reserve(batch_size);
+    int rows = 0;
+    while (const auto maybe_result = client->FetchOne()) {
+      if ((*maybe_result).size() == 0) {
+        break;
+      }
+
+      auto result = *maybe_result;
+
+      if (rows == batch_size) {
```
antoniofilipovic, line 205: When will `rows` be increased?

cpp/periodic_module/periodic_iterate.cpp
```diff
@@ -0,0 +1,241 @@
...
+    } else {
+      primitive_names.push_back(columns[i]);
+    }
+  }
+
+  return ParamNames{
+      .node_names = node_names, .relationship_names = relationship_names, .primitive_names = primitive_names};
+}
+
+std::vector<std::string> ConstructColumns(std::vector<std::string> columns, std::vector<mg::Value> &batch_row) {
```
antoniofilipovic, line 45: you can pass columns by `const &`

Dockerfile
```diff
@@ -26,6 +26,7 @@ RUN apt-get update && apt-get install -y \
     python3-dev     `mage-memgraph` \
     clang           `mage-memgraph` \
     git             `mage-memgraph` \
+    libboost-all-dev `mage-memgraph` \
```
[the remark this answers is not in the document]
antoniofilipovic: and dockerfile cugraph

antoniofilipovic (changes requested):
Good work, I left few comments and one optimization point to possibly benchmark

python/tests/requirements.txt
```diff
@@ -9,4 +9,3 @@ pytest-cov==2.12.1
 pytest-benchmark==3.4.1
 pytest-flake8==1.0.7
 pytest-black==0.3.12
-gqlalchemy==1.3.2
```
antoniofilipovic, line 12: I think we need this requirement for tests to run

antoniofilipovic: For e2e tests: 
- name: Run End-to-end tests
        if: matrix.architecture != 'arm64'
        env:
          PYTHONPATH: "$PWD/e2e"
        run: |
          cd e2e
          python -m pytest . -k 'not cugraph'

cpp/periodic_module/periodic_iterate.cpp
```diff
@@ -0,0 +1,241 @@
...
+  std::vector<std::string> with_entity_vector;
+  for (const auto &node_name : names.node_names) {
+    with_entity_vector.push_back(fmt::format("{}.{} AS __{}_id", kBatchRowInternalName, node_name, node_name));
+  }
+  for (const auto &rel_name : names.relationship_names) {
+    with_entity_vector.push_back(fmt::format("{}.{} AS __{}_id", kBatchRowInternalName, rel_name, rel_name));
+  }
+  for (const auto &prim_name : names.primitive_names) {
+    with_entity_vector.push_back(fmt::format("{}.{} AS {}", kBatchRowInternalName, prim_name, prim_name));
+  }
```
as51340, line 75: I would extract this into separate function since it does too many things. Also, each one of these for loops does the same thing so you can avoid duplication again

cpp/periodic_module/periodic_iterate.cpp
```diff
@@ -0,0 +1,241 @@
...
+  auto with_variables = fmt::format("WITH {}", boost::algorithm::join(with_entity_vector, ", "));
+
+  std::string match_string = "";
+  std::vector<std::string> match_by_id_vector;
+  for (const auto &node_name : names.node_names) {
+    match_by_id_vector.push_back(fmt::format("MATCH ({}) WHERE ID({}) = __{}_id", node_name, node_name, node_name));
+  }
+  for (const auto &rel_name : names.relationship_names) {
+    match_by_id_vector.push_back(fmt::format("MATCH ()-[{}]->() WHERE ID({}) = __{}_id", rel_name, rel_name, rel_name));
+  }
```
as51340, line 86: Again, I would opt for another smaller function

cpp/periodic_module/periodic_iterate.cpp
```diff
@@ -0,0 +1,241 @@
...
+      new_columns.push_back(fmt::format("__{}_id", columns[i]));
+    } else {
+      new_columns.push_back(columns[i]);
+    }
+  }
+
+  return new_columns;
+}
+
+std::string ConstructQueryPreffix(ParamNames names) {
```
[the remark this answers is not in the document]
as51340: There is utils::Join in Memgraph also if the performance is problem

cpp/periodic_module/periodic_iterate.cpp
```diff
@@ -0,0 +1,241 @@
...
+      }
+
+      batch.push_back(std::move(result));
+    }
+
+    ExecuteRunningQuery(running_query, columns, batch);
+
+    mg::Client::Finalize();
+
+    record.Insert(kReturnSuccess, true);
```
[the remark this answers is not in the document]
as51340: From my perspective, one success flag is better because even if some batch failed, what will user do with it? He will see false and then what

cpp/periodic_module/periodic_iterate.cpp
```diff
@@ -0,0 +1,241 @@
...
+  }
+
+  if (match_by_id_vector.size()) {
+    match_string = boost::algorithm::join(match_by_id_vector, " ");
+  }
+
+  return fmt::format("{} {} {}", unwind_batch, with_variables, match_string);
+}
+
+mg::Map ConstructParams(std::vector<std::string> columns, std::vector<std::vector<mg::Value>> &batch) {
```
as51340, line 95: Since this function is only used for constructing query params, I would say the better naming is ConstructQueryParams, what do you think?

cpp/periodic_module/periodic_iterate.cpp
```diff
@@ -0,0 +1,241 @@
...
+    }
+    if (!batch_size_value.IsInt()) {
+      throw std::runtime_error("Batch size not provided as an integer in the periodic iterate configuration!");
+    }
+
+    const auto batch_size = batch_size_value.ValueInt();
+
+    if (batch_size <= 0) {
+      throw std::runtime_error("Batch size must be a non-negative number!");
+    }
```
as51340, line 177: This smells to me like a single function: ValidateBatchSize

e2e/periodic_iterate_test/test_periodic_iterate_invalid_batch_argument_negative/test.yml
```diff
@@ -0,0 +1,5 @@
+query: >
+    CALL periodic.iterate("MATCH (n) RETURN n", "SET n.prop = 1", {batch_size: -1}) YIELD success RETURN success;
+
+exception: >
+    Batch size not provided as an integer in the periodic iterate configuration!
```
as51340, line 5: I think here the message should be Batch size must be a non-negative number!

cpp/periodic_module/periodic_iterate.cpp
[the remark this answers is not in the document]
as51340: But maybe returning number of batches would be good. The user would explicitly see in how many batches, the query was executed so he/she could maybe tune it next time + you can test that the code is working as expected.

as51340 (changes requested):
Nice job but I think there are things you can improve in your code. I would suggest adding test when batch is not 1000 or 1 because that would prove a bug that rows variable isn't incremented and test which works with relationships also not just vertices.

.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "cpp/memgraph"]
 	path = cpp/memgraph
 	url = https://github.com/memgraph/memgraph.git
-	branch = release/2.8
+	branch = fix-creating-mgp-value-from-nullptr
```
Josipmrden, line 4: This will be rewritten to release/v2.9 once it's up

cpp/periodic_module/periodic_iterate.cpp
```diff
@@ -0,0 +1,241 @@
...
+      record.Insert(kReturnSuccess, false);
+      return;
+    }
+
+    auto columns = client->GetColumns();
+
+    std::vector<std::vector<mg::Value>> batch;
+    batch.reserve(batch_size);
+    int rows = 0;
+    while (const auto maybe_result = client->FetchOne()) {
```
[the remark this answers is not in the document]
Josipmrden: but it should exist since `const auto maybe_result = client->FetchOne()` will evaluate to bool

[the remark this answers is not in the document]
Josipmrden: that's why it's in the while condition

cpp/periodic_module/periodic_iterate.cpp
```diff
@@ -0,0 +1,241 @@
...
+
+  auto param_names = ExtractParamNames(columns, batch[0]);
+  auto query_columns = ConstructColumns(columns, batch[0]);
+
+  auto preffix_query = ConstructQueryPreffix(param_names);
+  auto final_query = ConstructFinalQuery(running_query, preffix_query);
+
+  auto query_params = ConstructParams(columns, batch);
+
+  mg::Client::Params session_params{.host = "localhost", .port = 7687};
```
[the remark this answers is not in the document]
as51340: I am quite sure that both ways are correct, what I wrote made more sense to me but do as you want :smile:

cpp/periodic_module/periodic_iterate.cpp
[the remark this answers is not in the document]
as51340: shouldn't then the condition be:
`while (const auto maybe_result = client->FetchOne(); maybe_result)`

cpp/periodic_module/periodic_iterate.cpp
```diff
@@ -0,0 +1,306 @@
...
+      res.relationship_names.push_back(columns[i]);
+    } else {
+      res.primitive_names.push_back(columns[i]);
+    }
+  }
+
+  return res;
+}
+
+auto Join(const std::vector<std::string> &strings, const std::string &delimiter) {
```
as51340, line 45: std::string instead of auto

cpp/periodic_module/periodic_iterate.cpp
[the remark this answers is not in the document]
as51340: Both ways are correct, do as you want :smile:

cpp/periodic_module/periodic_iterate.cpp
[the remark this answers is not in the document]
as51340: But which type is then maybe_result?

cpp/periodic_module/periodic_iterate.cpp
```diff
@@ -0,0 +1,306 @@
...
+  if (maybe_username) {
+    username = std::move(maybe_username);
+  }
+
+  const auto *maybe_password = std::getenv(kMgPassword);
+  if (maybe_password) {
+    password = std::move(maybe_password);
+  }
+
+  return mg::Client::Params{.host = std::move(host),
```
as51340, line 217: just curious, what happens if some of these env variables aren't set, how will std::move behave?

cpp/periodic_module/periodic_iterate.cpp
```diff
@@ -0,0 +1,306 @@
...
+  }
+
+  auto unwind_batch = fmt::format("UNWIND ${} AS {}", kBatchInternalName, kBatchRowInternalName);
+  auto with_variables = ConstructWithStatement(names);
+  auto match_string = ConstructMatchGraphEntitiesById(names);
+
+  return fmt::format("{} {} {}", unwind_batch, with_variables, match_string);
+}
+
+auto ConstructQueryParams(const std::vector<std::string> &columns, const std::vector<std::vector<mg::Value>> &batch) {
```
as51340, line 121: why not explicit return value mg::Map?

cpp/periodic_module/periodic_iterate.cpp
```diff
@@ -0,0 +1,306 @@
...
+  }
+
+  if (match_by_id_vector.size()) {
+    match_string = Join(match_by_id_vector, " ");
+  }
+
+  return match_string;
+}
+
+auto ConstructQueryPrefix(const ParamNames &names) {
```
as51340, line 109: std::string instead of auto

cpp/periodic_module/periodic_iterate.cpp
```diff
@@ -0,0 +1,306 @@
...
+
+auto ConstructMatchingNodeById(const std::string &node_name) {
+  return fmt::format("MATCH ({}) WHERE ID({}) = __{}_id", node_name, node_name, node_name);
+}
+
+auto ConstructMatchingRelationshipById(const std::string &rel_name) {
+  return fmt::format("MATCH ()-[{}]->() WHERE ID({}) = __{}_id", rel_name, rel_name, rel_name);
+}
+
+auto ConstructMatchGraphEntitiesById(const ParamNames &names) {
```
as51340, line 92: std::string

cpp/periodic_module/periodic_iterate.cpp
```diff
@@ -0,0 +1,306 @@
...
+  std::string match_string = "";
+  std::vector<std::string> match_by_id_vector;
+  for (const auto &node_name : names.node_names) {
+    match_by_id_vector.push_back(ConstructMatchingNodeById(node_name));
```
as51340, line 96: emplace_back

cpp/periodic_module/periodic_iterate.cpp
```diff
@@ -0,0 +1,306 @@
...
+  }
+  for (const auto &rel_name : names.relationship_names) {
+    match_by_id_vector.push_back(ConstructMatchingRelationshipById(rel_name));
```
as51340, line 99: emplace_back

cpp/periodic_module/periodic_iterate.cpp
```diff
@@ -0,0 +1,306 @@
...
+}
+
+auto GetPrimitiveEntityAlias(const std::string &internal_name, const std::string &primitive_name) {
+  return fmt::format("{}.{} AS {}", internal_name, primitive_name, primitive_name);
+}
+
+auto ConstructWithStatement(const ParamNames &names) {
+  std::vector<std::string> with_entity_vector;
+  for (const auto &node_name : names.node_names) {
+    with_entity_vector.push_back(GetGraphFirstClassEntityAlias(kBatchRowInternalName, node_name));
```
as51340, line 70: emplace

cpp/periodic_module/periodic_iterate.cpp
```diff
@@ -0,0 +1,306 @@
...
+  }
+  for (const auto &rel_name : names.relationship_names) {
+    with_entity_vector.push_back(GetGraphFirstClassEntityAlias(kBatchRowInternalName, rel_name));
```
as51340, line 73: emplace

cpp/periodic_module/periodic_iterate.cpp
```diff
@@ -0,0 +1,306 @@
...
+  }
+  for (const auto &prim_name : names.primitive_names) {
+    with_entity_vector.push_back(GetPrimitiveEntityAlias(kBatchRowInternalName, prim_name));
```
as51340, line 76: emplace

cpp/periodic_module/periodic_iterate.cpp
```diff
@@ -0,0 +1,306 @@
...
+  }
+
+  auto with_variables = fmt::format("WITH {}", Join(with_entity_vector, ", "));
```
as51340, line 79: return immediately

cpp/periodic_module/periodic_iterate.cpp
as51340, line 45: more efficient version of this would be reserve the total size of the string upfront but :+1:

as51340 (changes requested):

python/tests/requirements.txt
antoniofilipovic: Ah, you are right, I didn't see that we are installing `requirements.txt`:

```
- name: Install Python test dependencies
        run: |
          python -m pip install --upgrade pip
          pip install -r ./python/tests/requirements.txt
```

cpp/periodic_module/periodic_iterate.cpp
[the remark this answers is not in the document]
antoniofilipovic: Maybe we can copy-paste it here?

.gitmodules
antoniofilipovic: Yes, we can bump this up then.

cpp/periodic_module/periodic_iterate.cpp
antoniofilipovic: Yes, check how is this implemented in memgraph. in `src/utils/string.hpp`

antoniofilipovic: Yes, it makes sense to do that here as we are not just returning output from other function

cpp/periodic_module/periodic_iterate.cpp
```diff
@@ -0,0 +1,306 @@
...
+const char *kDefaultHost = "localhost";
+const uint16_t kDefaultPort = 7687;
+
+struct ParamNames {
+  std::vector<std::string> node_names;
+  std::vector<std::string> relationship_names;
+  std::vector<std::string> primitive_names;
+};
+
+auto ExtractParamNames(const std::vector<std::string> &columns, const std::vector<mg::Value> &batch_row) {
```
antoniofilipovic, line 30: I would also put ParamNames as the return type instead of auto as we are returning type from within the function not just returning the output of other functions.

cpp/periodic_module/periodic_iterate.cpp
antoniofilipovic: And without `GetGraphFirstClassEntityAlias` as we want to not use ctor and move ctor then

cpp/periodic_module/periodic_iterate.cpp
antoniofilipovic: I think @Josipmrden does the correct check, gets env and if doesn't exist it should be predefined.

cpp/periodic_module/periodic_iterate.cpp
Josipmrden: will do

cpp/periodic_module/periodic_iterate.cpp
Josipmrden: I initialized it all to default values in the beginning, so I guess it's safe to assume that it will move the default variables since it did not go to if

cpp/periodic_module/periodic_iterate.cpp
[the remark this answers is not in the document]
antoniofilipovic: I think both `while (const auto maybe_result = client->FetchOne(); maybe_result)` and  while(const auto maybe_result = client->FetchOne()) are correct. @as51340  `std::optional<std::vector<Value>>` is type and when stream is empty it returns std::nullopt, that's why this will work correctly.

antoniofilipovic (approved):

antoniofilipovic:
@Josipmrden fix comments which @as51340 suggested, but otherwise looks good to me after that is fixed

cpp/periodic_module/periodic_iterate.cpp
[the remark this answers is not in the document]
Josipmrden: it's optional, and then with the assignment in the condition it is evaluated like true or false.

as51340 (approved):

Josipmrden:
@vpavicic  I still need docs for this, will do tomorrow

vpavicic:
> @vpavicic I still need docs for this, will do tomorrow

@Josipmrden release note as well please thank you you're awesome

vpavicic:
@Josipmrden ok?
When executing complex queries, the `period` module allows batching results
  from one query into another to improve execution time.

PR #222, opened 2023-06-12
[main < T222] Rewrite the do module in C++

Josipmrden:
### Description

This pull request concerns rewriting the `do` module and its corresponding methods (`do.case` and `do.when`) in C++ for increase of performance

### Pull request type

- [X] Refactoring (no functional changes, no api changes)

#### Module/Algorithm
- [X] Core algorithm/module implementation
- [X] [Query module](https://memgraph.com/docs/memgraph/reference-guide/query-modules) implementation
- [X] Unit tests -> not needed
- [x] End-to-end tests
- [X] Code documentation
- [X] README short description
- [X] Documentation on [memgraph/docs](https://github.com/memgraph/docs)


python/tests/requirements.txt
```diff
@@ -9,4 +9,3 @@ pytest-cov==2.12.1
 pytest-benchmark==3.4.1
 pytest-flake8==1.0.7
 pytest-black==0.3.12
-gqlalchemy==1.3.2
```
antoniofilipovic, line 12: I think we need this one for e2e tests. You can rewrite `.gitmodules` to run tests from inside docker and that way we can remove this dependency similar as it is now in case of Python module unit test workflow:

```
 - name: Python modules unit tests
        if: matrix.target == 'dev'
        env:
          PYTHONPATH: "$PWD/python"
        run: |
          docker exec -i -u root ${{ env.MAGE_CONTAINER }} bash -c "cd /mage/python/ && python3 -m pytest ."
```

e2e/do_test/type_conversion_group/test_return_node_relationship/test.yml
```diff
@@ -8,6 +8,9 @@ query: >
...
 
-output:
-    - n_nodes: 2
-      n_relationships: 2
+exception: >
+    Returning nodes in do procedures not yet 
+
+# output:
+#     - n_nodes: 2
+#       n_relationships: 2
```
antoniofilipovic, line 16: Comment left

e2e/do_test/test_strings_containing_quotes/test.yml
```diff
@@ -4,5 +4,8 @@ query: >
     YIELD value
     RETURN value.n.id AS nice_id;
 
-output:
-    - nice_id: 0
+exception: >
+    Returning nodes in do procedures not yet supported.
+
+# output:
+#     - nice_id: 0
```
antoniofilipovic, line 11: Comment

e2e/do_test/graph_types_group/test_return_node/test.yml
```diff
@@ -5,14 +5,17 @@ query: >
...
+#     - node:
+#           labels:
+#               - Node
+#           properties:
+#               id: 0
+#     - node:
+#           labels:
+#               - Node
+#           properties:
+#               id: 1
```
antoniofilipovic, line 21: Comment left, same in other files

cpp/do_module/do_module.cpp
```diff
@@ -0,0 +1,313 @@
...
+      node_names.push_back(std::string(map_item.key));
+    } else if (map_item.value.IsRelationship()) {
+      relationship_names.push_back(std::string(map_item.key));
+    } else {
+      primitive_names.push_back(std::string(map_item.key));
+    }
+  }
+
+  return ParamNames{
+      .node_names = node_names, .relationship_names = relationship_names, .primitive_names = primitive_names};
```
antoniofilipovic, line 54: I think `std::move` on `node_names` and others is valid to avoid copies

cpp/do_module/do_module.cpp
```diff
@@ -0,0 +1,313 @@
...
+  }
+
+  if (match_by_id_vector.size()) {
+    match_string = boost::algorithm::join(match_by_id_vector, " ");
+  }
+
+  return fmt::format("{} {}", with_variables, match_string);
+}
+
+std::string ConstructPreffixQuery(const mgp::Map &parameters) {
```
antoniofilipovic, line 91: you can even state that we return `auto` here

cpp/do_module/do_module.cpp
```diff
@@ -0,0 +1,313 @@
...
+
+  return ConstructQueryPreffix(param_names);
+}
+
+std::string ConstructFinalQuery(const std::string &running_query, const std::string &preffix_query) {
+  return fmt::format("{} {}", preffix_query, running_query);
+}
+
+mg::Map ConstructParams(const ParamNames &param_names, const mgp::Map &parameters) {
+  auto new_params = mg::Map(parameters.Size());
```
antoniofilipovic, line 102: I think this is better option as it avoids calling copy ctor:
 `mg::Map new_params{parameters.Size()};`

cpp/do_module/do_module.cpp
```diff
@@ -0,0 +1,313 @@
...
+      throw std::runtime_error("Can't parse some of the arguments!");
+    }
+  }
+
+  return new_params;
+}
+
+QueryResults ExecuteQuery(const std::string &query, const mgp::Map &query_parameters) {
+  mg::Client::Init();
+  mg::Client::Params params{.host = "localhost", .port = 7687};
```
antoniofilipovic, line 128: Does this need to be configurable

cpp/do_module/do_module.cpp
```diff
@@ -0,0 +1,313 @@
...
+
+    auto result = *maybe_result;
+    results.push_back(std::move(result));
+  }
+
+  client->DiscardAll();
+
+  mg::Client::Finalize();
+
+  return QueryResults{.columns = columns, .results = results};
```
antoniofilipovic, line 162: I think std::move on columns is valid to avoid copy, same on results

antoniofilipovic (changes requested):
Few minor suggestions, otherwise seems good

cpp/do_module/do_module.cpp
```diff
@@ -0,0 +1,313 @@
...
+    }
+
+    auto record = record_factory.NewRecord();
+    record.Insert(kReturnValue, result_map);
+  }
+}
+
+bool IsGlobalOperation(const std::string &query) {
+  for (const auto &global_op : kGlobalOperations) {
+    if (boost::algorithm::starts_with(query, global_op)) {
```
Ignition, line 201: https://en.cppreference.com/w/cpp/string/basic_string_view/starts_with

python/tests/requirements.txt
Josipmrden: but we do install it in the main requirements file, I'm not sure why it's again needed? These commands from the workflow don't install anything via pip

.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "cpp/memgraph"]
 	path = cpp/memgraph
 	url = https://github.com/memgraph/memgraph.git
-	branch = release/2.8
+	branch = mgp-add-any-type
```
Josipmrden, line 4: Will be changed to `release/2.9` once it's up

cpp/do_module/do_module.cpp
Josipmrden: Added configuration

cpp/do_module/do_module.cpp
```diff
@@ -0,0 +1,313 @@
...
+    } else {
+      primitive_names.push_back(std::string(map_item.key));
+    }
+  }
+
+  return ParamNames{
+      .node_names = node_names, .relationship_names = relationship_names, .primitive_names = primitive_names};
+}
+
+std::string ConstructQueryPreffix(const ParamNames &names) {
```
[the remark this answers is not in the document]
Josipmrden: where exactly do you think we would gain performance benefit?

cpp/do_module/do_module.cpp
```diff
@@ -0,0 +1,391 @@
...
+
+  auto joined_strings_size = strings[0].size();
+  for (size_t i = 1; i < strings.size(); i++) {
+    joined_strings_size += strings[i].size();
+  }
+
+  std::string joined_strings;
+  joined_strings.reserve(joined_strings_size + delimiter.size() * (strings.size() - 1));
+
+  for (size_t i = 1; i < strings.size(); i++) {
```
Ignition, line 72: missing `strings[0]`

python/tests/requirements.txt
antoniofilipovic: I think this is resolved now

cpp/do_module/do_module.cpp
```diff
@@ -0,0 +1,391 @@
...
+      res.relationship_names.push_back(std::string(map_item.key));
+    } else {
+      res.primitive_names.push_back(std::string(map_item.key));
+    }
+  }
+
+  return res;
+}
+
+std::string Join(std::vector<std::string> &strings, const std::string &delimiter) {
```
Ignition, line 59: `std::string Join(std::vector<std::string> const &strings, std::string_view delimiter)`

cpp/do_module/do_module.cpp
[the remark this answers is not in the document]
antoniofilipovic: On string creation. That's why I suggested that we can use `istream` to send everything in that stream and then produce one string from that stream.

I can't judge if it is a dumb suggestion from me as I think for some queries where batches are huge we are doing bigger concatenations of strings, and it could be an issue, but on the other hand doubt it will make a difference as actually executing of the query will take the most time, so feel free to skip a suggestion and resolve comment.

cpp/do_module/do_module.cpp
```diff
@@ -0,0 +1,391 @@
...
+  for (const auto &rel_name : names.relationship_names) {
+    with_entity_vector.emplace_back(GetGraphFirstClassEntityAlias(rel_name));
+  }
+  for (const auto &prim_name : names.primitive_names) {
+    with_entity_vector.emplace_back(GetPrimitiveEntityAlias(prim_name));
+  }
+
+  return fmt::format("WITH {}", Join(with_entity_vector, ", "));
+}
+
```
antoniofilipovic, line 101: Does it make sense to use here `std::move` on `node_names`

cpp/do_module/do_module.cpp
```diff
@@ -0,0 +1,391 @@
...
+  for (const auto &node_name : names.node_names) {
+    match_by_id_vector.emplace_back(ConstructMatchingNodeById(node_name));
+  }
+  for (const auto &rel_name : names.relationship_names) {
+    match_by_id_vector.emplace_back(ConstructMatchingRelationshipById(rel_name));
+  }
+
+  if (match_by_id_vector.size()) {
+    match_string = Join(match_by_id_vector, " ");
+  }
```
antoniofilipovic, line 122: `std::string match_string{""}`

cpp/do_module/do_module.cpp
```diff
@@ -0,0 +1,391 @@
+#include <fmt/core.h>
+#include <mgp.hpp>
+
+#include "mgclient.hpp"
+
+const char *kProcedureCase = "case";
```
Ignition, line 6: `const char *` is a global that may change.
Prefer `constexpr std::string_view` for better codegen.

cpp/do_module/do_module.cpp
```diff
@@ -0,0 +1,391 @@
...
+      new_params.Insert(map_item.key, mg::Value(map_item.value.ValueBool()));
+    } else if (map_item.value.IsString()) {
+      new_params.Insert(map_item.key, mg::Value(map_item.value.ValueString()));
+    } else if (map_item.value.IsInt()) {
+      new_params.Insert(map_item.key, mg::Value(map_item.value.ValueInt()));
+    } else if (map_item.value.IsDouble()) {
+      new_params.Insert(map_item.key, mg::Value(map_item.value.ValueDouble()));
+    } else {
+      // Temporal types and paths not yet supported
+      throw std::runtime_error("Can't parse some of the arguments!");
```
antoniofilipovic, line 166: Throw error what can't be parsed exactly.

cpp/do_module/do_module.cpp
```diff
@@ -0,0 +1,391 @@
...
+
+  return ConstructQueryPreffix(param_names);
+}
+
+std::string ConstructFinalQuery(const std::string &running_query, const std::string &preffix_query) {
+  return fmt::format("{} {}", preffix_query, running_query);
+}
+
+mg::Map ConstructParams(const ParamNames &param_names, const mgp::Map &parameters) {
+  mg::Map new_params{parameters.Size()};
```
antoniofilipovic, line 149: `mg::Map new_params{parameters.Size()};`
prefer initialization to doing assignment

cpp/do_module/do_module.cpp
```diff
@@ -0,0 +1,391 @@
...
+
+  client->DiscardAll();
+
+  mg::Client::Finalize();
+
+  return QueryResults{.columns = std::move(columns), .results = std::move(results)};
+}
+
+void InsertConditionalResults(const mgp::RecordFactory &record_factory, const QueryResults &query_results) {
+  for (const auto &row : query_results.results) {
```
antoniofilipovic, line 244: `const auto &`
no need to do copy

cpp/do_module/do_module.cpp
```diff
@@ -0,0 +1,391 @@
...
+
+  auto columns = client->GetColumns();
+
+  std::vector<std::vector<mg::Value>> results;
+  while (const auto maybe_result = client->FetchOne()) {
+    if ((*maybe_result).size() == 0) {
+      break;
+    }
+
+    auto result = *maybe_result;
```
antoniofilipovic, line 232: you can just do
```
results.push_back(std::move(*maybe_result));
```

cpp/do_module/do_module.cpp
```diff
@@ -0,0 +1,391 @@
...
+  client->DiscardAll();
+
+  mg::Client::Finalize();
+
+  return QueryResults{.columns = std::move(columns), .results = std::move(results)};
+}
+
+void InsertConditionalResults(const mgp::RecordFactory &record_factory, const QueryResults &query_results) {
+  for (const auto &row : query_results.results) {
+    auto result_map = mgp::Map();
```
antoniofilipovic, line 245: prefer initialization to assignment:

```
mgp::Map result_map;
```

Ignition: FYI that is construction, no assignment is involved. see https://godbolt.org/z/8P78MeWrW

vpavicic:
@Josipmrden release note pliz

vpavicic:
@Josipmrden  ok?

The `conditional_execution` module that allows executing different queries
  based on certain conditions being met has been rewritten from Python to C++ to
  improve performance.

cpp/do_module/do_module.cpp
antoniofilipovic: @Ignition I thought all the time it was an assignment, thanks :+1:

antoniofilipovic (approved):
Looks good to me

Ignition (approved):
Looks good to me 👍

PR #226, opened 2023-06-14
Add build language picker

gitbuda:
Trying to build the whole `mage` under Ubuntu 22.04 -> tests or even running everything inside Memgraph is tricky because Python is expensive to set up. My only need was to build cpp part, so this change will make it easier to build and test parts of the project.

Notes
- Usage: `python setup build --lang Cpp`
- Build related changes (add build option without breaking compatibility)


antoniofilipovic (approved):
Nice! Approve from me

PR #228, opened 2023-06-23
Fix DockerHub production tag

gitbuda:
- [ ] It Seems like this will actually work next time -> at least leave a comment where the version if specified what is going to happen if version specifier structure changes


gitbuda:
all good on this side

PR #238, opened 2023-07-19
[E-matija < T427] Implement unionAll

mpintaric55334:
### Description

Implemented unionAll and tests

### Pull request type

- [ ] Algorithm/Module


cpp/collections_module/algorithms/collections.hpp
```diff
@@ -0,0 +1,23 @@
+#include <mgp.hpp>
+
+
+void unionAll(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result, mgp_memory *memory){
+  mgp::memory=memory;
+  const auto arguments=mgp::List(args);
+  const auto record_factory = mgp::RecordFactory(result);
+  try{
+    mgp::List list1=arguments[0].ValueList();
```
antoniofilipovic, line 9: `const mgp::List &` to avoid copies

cpp/collections_module/algorithms/collections.hpp
```diff
@@ -0,0 +1,23 @@
+#include <mgp.hpp>
+
+
+void unionAll(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result, mgp_memory *memory){
+  mgp::memory=memory;
+  const auto arguments=mgp::List(args);
```
antoniofilipovic, line 6: empty space between arugments and =, and = and mgp::List. I think we need to apply formatting here, ping me on slack to discuss about formatting

cpp/collections_module/algorithms/collections.hpp
```diff
@@ -0,0 +1,23 @@
...
+  const auto record_factory = mgp::RecordFactory(result);
+  try{
+    mgp::List list1=arguments[0].ValueList();
+    mgp::List list2=arguments[1].ValueList();
+    
+    for (size_t i=0; i< list2.Size(); i++){
+        list1.AppendExtend(list2[i]);
```
antoniofilipovic, line 13: you can use std::move on list2[i] I think, as we are moving values from one list to other, so should be fine. check about move semantics here: https://en.cppreference.com/w/cpp/utility/move

cpp/collections_module/algorithms/collections.hpp
```diff
@@ -0,0 +1,23 @@
...
+    }
+    auto record=record_factory.NewRecord();
+    record.Insert("return_list",list1);
```
antoniofilipovic, line 16: add "return_list" as global var

cpp/collections_module/collections.cpp
```diff
@@ -1,9 +1,23 @@
 #include <mgp.hpp>
 
+#include "algorithms/collections.hpp"
+
+const char *kProcedureUnionALl = "unionAll";
+
+const char *kReturnUnionAll="return_list";
+
+
+const char *kArgumentList1= "list1";
```
antoniofilipovic, line 10: use `constexpr std::string:view` instead of `const char*` as it can be changed for better code style

e2e/collections_test/test_unionAll/input.cyp
```diff
@@ -0,0 +1,2 @@
+CREATE (dog:Dog)-[loves:Loves]->(human:Human)
+CREATE(player:Player)-[scores:Scores]->(goal:Goal)
```
antoniofilipovic, line 2: add empty line here too

e2e/collections_test/test_unionAll/test.yml
```diff
@@ -0,0 +1,21 @@
+query: >
+    CALL collections.unionAll([1,2,3,4,5], [3,[3,6,7],5,6,"a"]) YIELD return_list RETURN return_list;
+
+output:
+  - return_list: [1, 2, 3, 4, 5, 3, [3, 6, 7], 5, 6, "a"]
```
antoniofilipovic, line 5: split these tests in different files, as I mentioned on other PR

antoniofilipovic (changes requested):
Looks good, left few comments

cpp/collections_module/algorithms/collections.hpp
mpintaric55334: If I add that how can I append to list1 and use move on list2 when  the are const?

cpp/collections_module/collections.cpp
mpintaric55334: I get an error, same as Idora in Slack

cpp/collections_module/algorithms/collections.hpp
```diff
@@ -0,0 +1,29 @@
...
+  const auto record_factory = mgp::RecordFactory(result);
+  try{
+    mgp::List list1 = arguments[0].ValueList();
+    mgp::List list2 = arguments[1].ValueList();
+    
+    for (size_t i = 0; i < list2.Size(); i++){
+        list1.AppendExtend(std::move(list2[i]));
+    }
+    auto record = record_factory.NewRecord();
+    record.Insert(std::string(kReturnValueUnionAll).c_str(), list1);
```
antoniofilipovic, line 22: I have an idea here to extend CPP API so we can move the list here instead of doing copy underlying. Ping me on Monday so we can sync what to do here.

antoniofilipovic (changes requested):
One small comment still and then we can merge. Considering CPP API, it is not issue for this PR but I think we should extend it

cpp/collections_module/algorithms/collections.hpp
```diff
@@ -0,0 +1,29 @@
...
+
+
+
+void unionAll(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result, mgp_memory *memory){
+  mgp::memory = memory;
+  const auto arguments = mgp::List(args);
+  const auto record_factory = mgp::RecordFactory(result);
+  try{
+    mgp::List list1 = arguments[0].ValueList();
+    mgp::List list2 = arguments[1].ValueList();
```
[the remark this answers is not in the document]
antoniofilipovic: Comment left to fix

antoniofilipovic (approved):
Good to go

PR #240, opened 2023-07-19
[E-matija-add-collections < T421] Add collections.contains

mpintaric55334:
### Description
Implemented coll.contains and wrote tests for it

### Pull request type

- [ ] Algorithm/Module


cpp/collections_module/algorithms/collections.hpp
```diff
@@ -0,0 +1,30 @@
+#include <mgp.hpp>
+
+
+void contains(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result, mgp_memory *memory){
+  mgp::memory=memory;
+  const auto arguments=mgp::List(args);
+  const auto record_factory = mgp::RecordFactory(result);
+  try{
+    mgp::List list=arguments[0].ValueList();
```
antoniofilipovic, line 9: `const mgp::List &` so that you avoid copies

cpp/collections_module/algorithms/collections.hpp
```diff
@@ -0,0 +1,30 @@
...
+    mgp::Value value=arguments[1];
```
antoniofilipovic, line 10: `const mgp::Value &`

cpp/collections_module/algorithms/collections.hpp
```diff
@@ -0,0 +1,30 @@
...
+    bool contains_value=false;
```
antoniofilipovic, line 11:
```suggestion
    bool contains_value{false};
```

cpp/collections_module/algorithms/collections.hpp
```diff
@@ -0,0 +1,30 @@
...
+    if(list.Size()==0 || list.Empty()){
```
antoniofilipovic, line 12: Shouldn't this size() or empty be the same?

cpp/collections_module/algorithms/collections.hpp
```diff
@@ -0,0 +1,30 @@
...
+      contains_value=false;
```
antoniofilipovic, line 13: you can set value on record here and use "early return" concept, you can check it here: https://medium.com/swlh/return-early-pattern-3d18a41bba8

cpp/collections_module/algorithms/collections.hpp
```diff
@@ -0,0 +1,30 @@
...
+    }else{
```
antoniofilipovic, line 14: now you wouldn't need else branch with early return

cpp/collections_module/algorithms/collections.hpp
```diff
@@ -0,0 +1,30 @@
...
+      for(size_t i=0;i<list.Size();i++){
```
antoniofilipovic, line 15: `const size_t `

cpp/collections_module/algorithms/collections.hpp
```diff
@@ -0,0 +1,30 @@
...
+        if(list[i]==value){
+          contains_value=true;
+          break;
+        }
+      }
+    }
+    auto record=record_factory.NewRecord();
+    record.Insert("output",contains_value);
```
antoniofilipovic, line 23: set "output" as global variable. You can check Ivan's PRs

cpp/collections_module/algorithms/collections.hpp
```diff
@@ -0,0 +1,30 @@
...
+    
+  }catch (const std::exception &e) {
+    record_factory.SetErrorMessage(e.what());
+    return;
+  }
+
+}
```
antoniofilipovic, line 30: This red circle means that empty line is missing. Please add empty line to other files too

e2e/collections_test/test_contains/test.yml
```diff
@@ -0,0 +1,37 @@
...
+    CALL collections.contains([1,2,3,4,5], 4) YIELD output RETURN output;
+
+output:
+  - output: true
+
+query: >
+    CALL collections.contains([1,2,3,4,5], [3,4])  YIELD output RETURN output;
+
+output:
+  - output: false
```
antoniofilipovic, line 11: Did you check somehow that this was called? I am not sure if it was by code in "test_module.py" which is calling only one "query". But nevertheless, I would split these in different files as they are effectively different tests, and this was practice until now.

antoniofilipovic (changes requested):
Good work, left few c++ comments and styling guides

cpp/collections_module/algorithms/collections.hpp
mpintaric55334: How would I then increase i, if its constant

antoniofilipovic (changes requested):
All seems good, just run clang-tidy format

e2e/collections_test/test_contains1/test.yml
```diff
@@ -0,0 +1,8 @@
+query: >
+    CALL collections.contains([1,2,3,4,5], 4) YIELD output RETURN output;
+
+output:
+  - output: true
+
+
+
```
antoniofilipovic, line 8: You have too many empty lines here now :joy:

e2e/collections_test/test_contains3/test.yml
```diff
@@ -0,0 +1,11 @@
...
+query: >
+    MATCH ()-[r]->()
+    MATCH (n:Dog)-[k]->(t:Human)
+    WITH collect(DISTINCT r) AS collectedR, k
+    CALL collections.contains(collectedR, k) YIELD output RETURN output;
+
+output:
+  - output: true
+
+
```
antoniofilipovic, line 11: And here one to many I think

e2e/collections_test/test_contains4/test.yml
```diff
@@ -0,0 +1,9 @@
+query: >
+    MATCH (l:Player)-[r]->(g:Goal)
+    MATCH (n:Dog)-[k]->(t:Human)
+    WITH collect(DISTINCT r) AS collectedR, k
+    CALL collections.contains(collectedR, k) YIELD output RETURN output;
+output:
+  - output: false
+
```
antoniofilipovic, line 8: Here also one to many empty line

e2e/collections_test/test_contains5/test.yml
```diff
@@ -0,0 +1,9 @@
+
+query: >
+    MATCH p = ()-[*]->()
+    WITH collect(p) AS paths
+    MATCH q = (:Dog)-[*]->(:Human)
+    CALL collections.contains(paths, q) YIELD output RETURN output;
+output:
+  - output: true
+
```
antoniofilipovic, line 9: This shouldn't also be like this, you can check input.cyp of this test, this is how it is supposed to be :smile:

antoniofilipovic (commented):
Please just refactor to algorithms.cpp and algorithms.hpp and then that should be it

antoniofilipovic (approved):
Good to go

PR #249, opened 2023-07-24
[E-matija-add-collections < T411] Add collections.min

mpintaric55334:
Implemented collections.min and wrote tests


cpp/collections_module/algorithms/collections.hpp
```diff
@@ -0,0 +1,45 @@
+#include <mgp.hpp>
+#include <sstream>
+#include <string>
+
+constexpr std::string_view kReturnValueMin = "min";
+constexpr std::string_view kProcedureMin = "min";
+constexpr std::string_view kArgumentListMin = "list";
+
+void min(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result, mgp_memory *memory) {
```
antoniofilipovic, line 9: Naming, here it should be called `Min`

cpp/collections_module/algorithms/collections.hpp
```diff
@@ -0,0 +1,45 @@
...
+    auto record = record_factory.NewRecord();
+
+    if (type == mgp::Type::Map || type == mgp::Type::Path || type == mgp::Type::List) {
+      std::ostringstream oss;
+      oss << type;
+      std::string s = oss.str();
+      throw mgp::ValueException("Unsuppported type for this operation, receieved type: " + s);
+    }
+
+    mgp::Value min = list[0];
```
antoniofilipovic, line 28: you can do it like this too:
`mgp::Value min{std::move(list[0])};`

cpp/collections_module/algorithms/collections.hpp
```diff
@@ -0,0 +1,45 @@
...
+    for (size_t i = 0; i < list.Size(); i++) {
+      if (list[i].Type() != type && !(list[0].IsNumeric() && list[i].IsNumeric())) {
```
antoniofilipovic, line 30: what are you doing with list[0]? I think you can use move above and cache in some variable, `bool isListNumeric = list[0].isNumeric()`.

But do that before you move value from list

e2e/collections_test/test1/test.yml
```diff
@@ -0,0 +1,6 @@
+query: > 
+  CALL collections.min([1,2,3,4,5]) YIELD min RETURN min;
+
+output:
+  - min: 1
+
```
antoniofilipovic, line 6: You somehow have empty line and then there is this red circle :joy: try deleting one line and then it should be okay, but press enter after `- min: 1` line

antoniofilipovic (changes requested):
Please also refactor as we talked on Slack

cpp/collections_module/algorithms/collections.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+void min(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result, mgp_memory *memory) {
+  mgp::memory = memory;
+  const auto arguments = mgp::List(args);
+  const auto record_factory = mgp::RecordFactory(result);
+  try {
+    const mgp::List &list = arguments[0].ValueList();
+    if (list.Empty()) {
+      throw mgp::ValueException("Empty input list");
+    }
```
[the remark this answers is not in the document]
mpintaric55334: Neo4j returns null, i thought an exception made more sense here, but can change it to null if you'd like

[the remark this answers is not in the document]
antoniofilipovic: I think it makes more sense as it is not breaking behaviour

cpp/collections_module/algorithm/algorithm.cpp
```diff
@@ -0,0 +1,40 @@
+#include "algorithm.hpp"
+
+void Collections::Min(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result, mgp_memory *memory) {
+  mgp::memory = memory;
+  const auto arguments = mgp::List(args);
+  const auto record_factory = mgp::RecordFactory(result);
+  try {
+    const mgp::List &list = arguments[0].ValueList();
+    if (list.Empty()) {
+      throw mgp::ValueException("Empty input list");
```
antoniofilipovic, line 10: Change this to return null

antoniofilipovic: It makes more sense as it is not breaking behaviour

cpp/collections_module/algorithm/algorithm.cpp
```diff
@@ -0,0 +1,40 @@
...
+    }
+    const mgp::Type &type = list[0].Type();
+    auto record = record_factory.NewRecord();
+
+    if (type == mgp::Type::Map || type == mgp::Type::Path || type == mgp::Type::List) {
+      std::ostringstream oss;
+      oss << type;
+      std::string s = oss.str();
+      throw mgp::ValueException("Unsuppported type for this operation, receieved type: " + s);
+    }
```
antoniofilipovic, line 20: Won't this break on our mgp::Value side where we compare?

antoniofilipovic (approved):
Just fix this couple of stuff and it can be merged

PR #250, opened 2023-07-24
[E-Idora < T426] Implement union

ind1xa:
### Description

implemented union procedure and e2e tests for it

### Pull request type

- [ ] Bugfix
- [x ] Algorithm/Module
- [ ] Feature
- [ ] Build related changes

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation


cpp/collections_module/algorithms/union.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+void Union(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result, mgp_memory *memory) {
+  mgp::memory = memory;
+  const auto arguments = mgp::List(args);
+  const auto record_factory = mgp::RecordFactory(result);
+  try {
+    const auto &list1 = arguments[0].ValueList();
+    const auto &list2 = arguments[1].ValueList();
+
+    std::unordered_set<mgp::Value, mgp::Value::Hash> unionSet;
```
antoniofilipovic, line 16: There is one potential issue with this approach. Some values can have hash collisions. So what I would suggest is that you use std::unordered_map with hash as key, and std::vector as values. If it happens that values have same hash value, there are two options:
 - either they are the same values so you don't do anything
 - they are different values so you have a hash collision. In that case, add value to that vector

I think this is correct approach and still fast enough

antoniofilipovic (changes requested):
Please also don't forget to refactor, and if you have some questions about what I was thinking with this comment, ping me on slack

cpp/collections_module/algorithms/union.hpp
imilinovic: I think that isn't necessary because unordered_set resolves collision with chaining (to implement unordered structures you need define hash and == operator).
More info https://stackoverflow.com/questions/21518704/how-does-c-stl-unordered-map-resolve-collisions

antoniofilipovic: Okay, works then, didn't know about that :smile:

antoniofilipovic (approved):
Good to go

PR #251, opened 2023-07-25
[E-ivan < T422] Implement collections.containsAll

imilinovic:
### Description

### Pull request type

- [ ] Bugfix
- [X] Algorithm/Module
- [ ] Feature
- [ ] Build related changes

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation

### DIFF
--- a/cpp/collections_module/algorithm/algorithm.cpp
+++ b/cpp/collections_module/algorithm/algorithm.cpp
@@ -1,7 +1,61 @@
 #include "algorithm.hpp"
-
+#include <sstream>
 #include <unordered_set>
 #include "mgp.hpp"
+
+void Collections::SumLongs(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result, mgp_memory *memory) {
+  mgp::memory = memory;
+  const auto arguments = mgp::List(args);
+  const auto record_factory = mgp::RecordFactory(result);
+
+  try {
+    int64_t sum{0};
+    const auto &list = arguments[0].ValueList();
+
+    for (const auto list_item : list) {
+      if (!list_item.IsNumeric()) {
+        std::ostringstream oss;
+        oss << list_item.Type();
+        throw mgp::ValueException("Unsupported type for this operation, received type: " + oss.str());
+      }
+      sum += static_cast<int64_t>(list_item.ValueNumeric());
+    }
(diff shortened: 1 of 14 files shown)


cpp/collections_module/algorithm/algorithm.cpp
```diff
@@ -0,0 +1,30 @@
...
+#include <unordered_set>
+#include "mgp.hpp"
+
+void Collections::ContainsAll(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result, mgp_memory *memory) {
+  mgp::memory = memory;
+  const auto arguments = mgp::List(args);
+  const auto record_factory = mgp::RecordFactory(result);
+
+  try {
+    const auto &list1{arguments[0].ValueList()};
```
antoniofilipovic, line 12: You don't need reference here as you are creating value list1. References are  like aliases.

cpp/collections_module/algorithm/algorithm.cpp
```diff
@@ -0,0 +1,30 @@
...
+    std::unordered_set<mgp::Value> set{list1.begin(), list1.end()};
+    const auto &list2{arguments[1].ValueList()};
```
antoniofilipovic, line 14: also you don't need reference here

cpp/collections_module/algorithm/algorithm.cpp
```diff
@@ -0,0 +1,30 @@
...
+    std::unordered_set<mgp::Value> values{list2.begin(), list2.end()};
+
+    bool contained{true};
+    for (const auto &value : values)
+      if (set.find(value) == set.end()) {
+        contained = false;
+        break;
+      }
```
antoniofilipovic, line 22: can you use here one of the following: https://en.cppreference.com/w/cpp/container/unordered_set/operator_cmp

or https://en.cppreference.com/w/cpp/algorithm/all_any_none_of, you check if every element is in another set?

antoniofilipovic (changes requested):
Few small changes but good work

cpp/collections_module/algorithm/algorithm.cpp
imilinovic: Changed to all_of

PR #253, opened 2023-07-25
[E-matija-add-collections < T417] implement collections.toSet

mpintaric55334:
### Description

### Pull request type

- [ ] Bugfix
- [ ] Algorithm/Module
- [ ] Feature
- [ ] Build related changes

### Related issues

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation


cpp/collections_module/algorithm/algorithm.hpp
```diff
@@ -0,0 +1,10 @@
+#include <mgp.hpp>
+#include <string>
+
+namespace Collections {
+
+constexpr std::string_view kReturnToSet = "result";
+constexpr std::string_view kProcedureToSet = "toSet";
```
antoniofilipovic, line 7: can we call this "to_set". I think this is the style we have been using for now in Memgraph procedures

cpp/collections_module/algorithm/algorithm.cpp
```diff
@@ -0,0 +1,25 @@
+#include "algorithm.hpp"
+#include <unordered_set>
+
+void Collections::toSet(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result, mgp_memory *memory) {
+  mgp::memory = memory;
+  auto arguments = mgp::List(args);
+  const auto record_factory = mgp::RecordFactory(result);
+  try {
+    mgp::List list = arguments[0].ValueList();
+    std::unordered_set<mgp::Value, std::hash<mgp::Value>> set;
```
antoniofilipovic, line 10: I think @imilinovic somewhere wrote you don't need anymore std::hash part

antoniofilipovic, line 10: You can construct this set from a list directly, like this
std::unordered_set<..> set{list.begin(), list.end())
https://en.cppreference.com/w/cpp/container/unordered_set/unordered_set, constructor (6)

cpp/collections_module/algorithm/algorithm.cpp
```diff
@@ -0,0 +1,25 @@
...
+    mgp::List return_list;
+    for (auto list_elem : list) {
+      if (set.find(list_elem) == set.end()) {
+        set.insert(list_elem);
+        return_list.AppendExtend(std::move(list_elem));
+      }
+    }
```
antoniofilipovic, line 17: And here you just go over the elements of set and move them to list

antoniofilipovic (changes requested):
Few small changes

cpp/collections_module/algorithm/algorithm.cpp
mpintaric55334: FYI: when I do that, I messaged you on SLack, the order is not in reverse after all like I messaged, but its completely random, I guess thats okay because of the hashes.

PR #254, opened 2023-07-25
[E-matija< T414] Implement collections partition

mpintaric55334:
### Description

### Pull request type

- [ ] Bugfix
- [ ] Algorithm/Module
- [ ] Feature
- [ ] Build related changes

### Related issues

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation


cpp/collections_module/algorithm/algorithm.cpp
```diff
@@ -0,0 +1,37 @@
+#include "algorithm.hpp"
+
+void Collections::Partition(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result, mgp_memory *memory) {
+  mgp::memory = memory;
+  auto arguments = mgp::List(args);
+  const auto record_factory = mgp::RecordFactory(result);
+  try {
+    mgp::List input_list = arguments[0].ValueList();
+    const int64_t &partition_size = arguments[1].ValueInt();
```
antoniofilipovic, line 9: I don't think you need reference here, it is plain int and I think this is reference to temporary so it can lead to undefined behaviour

antoniofilipovic, line 9: I think ValueInt returns new object, so no need for reference here. It is I think reference to temporary

cpp/collections_module/algorithm/algorithm.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+    int64_t current_size = 0;
+    mgp::List temp;
+    mgp::List result;
+    for (mgp::Value list_value : input_list) {
```
antoniofilipovic, line 14: you can use `&` here to avoid copy

cpp/collections_module/algorithm/algorithm.cpp
```diff
@@ -0,0 +1,37 @@
...
+      if (current_size == 0) {
```
antoniofilipovic, line 15: can you get the size current with temp.Size() or something like that?

cpp/collections_module/algorithm/algorithm.cpp
```diff
@@ -0,0 +1,37 @@
...
+      current_size++;
+
+      if (current_size == partition_size) {
+        auto record = record_factory.NewRecord();
+        record.Insert(std::string(kReturnValuePartition).c_str(), std::move(temp));
+        current_size = 0;
+      }
+    }
+
+    if (current_size != partition_size && current_size != 0) {
```
antoniofilipovic, line 28: check again if you can use temp.Size to check size of this

antoniofilipovic (changes requested):
small changes to do

antoniofilipovic (approved):
Looks good

PR #256, opened 2023-07-25
[E-Idora < T429] Implement remove all for collections

ind1xa:
### Description

### Pull request type

- [ ] Bugfix
- [x ] Algorithm/Module
- [ ] Feature
- [ ] Build related changes

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation


cpp/collections_module/algorithms/algorithms.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+void Collections::RemoveAll(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result, mgp_memory *memory) {
+  mgp::memory = memory;
+  const auto arguments = mgp::List(args);
+  const auto record_factory = mgp::RecordFactory(result);
+
+  try {
+    const auto &input_list = arguments[0].ValueList();
+    const auto &to_remove_list = arguments[1].ValueList();
+    std::vector<mgp::Value> searchable;
```
antoniofilipovic, line 11: Can you use set here?

cpp/collections_module/algorithms/algorithms.cpp
```diff
@@ -0,0 +1,37 @@
+#include "algorithms.hpp"
+
+void Collections::RemoveAll(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result, mgp_memory *memory) {
+  mgp::memory = memory;
+  const auto arguments = mgp::List(args);
+  const auto record_factory = mgp::RecordFactory(result);
+
+  try {
+    const auto &input_list = arguments[0].ValueList();
+    const auto &to_remove_list = arguments[1].ValueList();
```
antoniofilipovic, line 10: These two don't need references, sorry

cpp/collections_module/algorithms/algorithms.cpp
```diff
@@ -0,0 +1,37 @@
...
+    std::vector<mgp::Value>::iterator itr;
+
+    for (const auto key : to_remove_list) {
+      while (true) {
+        itr = std::find(searchable.begin(), searchable.end(), key);
+        if (itr != searchable.cend()) {
+          searchable.erase(itr);
+        } else {
+          break;
+        }
```
antoniofilipovic, line 26:
```suggestion
        if (itr == searchable.cend()) {
           break;
        }
        searchable.erase(itr);
```

antoniofilipovic (changes requested):
I think set can be utilized instead od vector, otherwise looks good

cpp/collections_module/algorithms/algorithms.cpp
```diff
@@ -0,0 +1,41 @@
...
+  const auto record_factory = mgp::RecordFactory(result);
+
+  try {
+    const auto input_list = arguments[0].ValueList();
+    const auto to_remove_list = arguments[1].ValueList();
+    std::set<mgp::Value> searchable;
+
+    for (const auto value : input_list) {
+      searchable.insert(std::move(value));
+    }
```
antoniofilipovic, line 15: you can join these two into one part:
`std::set<mgp::Value> searchable{input_list.begin(), input_list.end()};`

cpp/collections_module/algorithms/algorithms.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+    for (const auto key : to_remove_list) {
+      while (true) {
+        itr = searchable.find(key);
+        if (itr == searchable.end()) {
+          break;
+        }
+        searchable.erase(itr);
+      }
+    }
```
antoniofilipovic, line 27: I am wondering now will this implementation work in case you have:

[1,2,3,2,4], [4]

because with the set, you lose cardinality. You can use std::multiset in that way so you are safe with cardinality of numbers

antoniofilipovic (changes requested):
Please check if current implementation works properly

cpp/collections_module/algorithms/algorithms.cpp
ind1xa: i tried this but it won't compile and I'm not sure I understand why:
error: no matching function for call to ‘std::_Rb_tree<mgp::Value, mgp::Value, std::_Identity<mgp::Value>, std::less<mgp::Value>, std::allocator<mgp::Value> >::_M_insert_range_equal(mgp::List::Iterator&, mgp::List::Iterator&)’

antoniofilipovic: Leave it as it is

antoniofilipovic (approved):

PR #267, opened 2023-07-28
[E-Ivan < T393] Implement from_values on map

imilinovic:
### Description
Implement map.from_values

### Pull request type

- [ ] Bugfix
- [X] Algorithm/Module
- [ ] Feature
- [ ] Build related changes

### Related issues

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation

### DIFF
--- a/cpp/map_module/algorithm/map.cpp
+++ b/cpp/map_module/algorithm/map.cpp
@@ -1,6 +1,57 @@
 #include "map.hpp"
+#include <string>
 #include "mg_utils.hpp"
 #include "mgp.hpp"
+
+namespace {
+std::string value_to_string(const mgp::Value &value) {
+  switch (value.Type()) {
+    case mgp::Type::Null:
+      return "";
+    case mgp::Type::Any:
+      return "";
+    case mgp::Type::Bool:
+      return (value.ValueBool() ? "true" : "false");
+    case mgp::Type::Int:
+      return std::to_string(value.ValueInt());
+    case mgp::Type::Double:
+      return std::to_string(value.ValueDouble());
+    case mgp::Type::String:
+      return std::string(value.ValueString());
+    case mgp::Type::List:
+      return "";
+    case mgp::Type::Map:
+      return "";
+    case mgp::Type::Node:
+      return "Node[" + std::to_string(value.ValueNode().Id().AsInt()) + "]";
+    case mgp::Type::Relationship:
+      return "Relationship[" + std::to_string(value.ValueRelationship().Id().AsInt()) + "]";
+    case mgp::Type::Path:
+      return "";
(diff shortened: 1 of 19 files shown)


cpp/map_module/algorithm/map.cpp
```diff
@@ -1 +1,36 @@
...
+      throw mgp::ValueException("List needs to have an even number of elements");
+    }
+
+    auto iterator = values.begin();
+    while(iterator != values.end()) {
+      auto &key = *iterator;
+      if (!key.IsString()) {
+        throw mgp::ValueException("Keys need to be strings");
+      }
+      ++ iterator;
```
antoniofilipovic, line 24: `++iterator`

cpp/map_module/algorithm/map.cpp
```diff
@@ -1 +1,36 @@
...
+      map.Update(key.ValueString(), *iterator);
+      ++ iterator;
```
antoniofilipovic, line 26:
```suggestion
      ++iterator;
```

cpp/map_module/algorithm/map.cpp
antoniofilipovic, line 24:
```suggestion
      ++iterator;
```

antoniofilipovic (changes requested):
Small changes, otherwise looks good

cpp/map_module/algorithm/map.cpp
```diff
@@ -1 +1,86 @@
 #include "map.hpp"
+#include <string>
+#include "mg_utils.hpp"
+#include "mgp.hpp"
+
+namespace {
+std::string value_to_string(const mgp::Value &value) {
```
imilinovic, line 7: This function supports key being any type that isn't null, list, map or path. There wasn't a simple way to implement it (except null which can be ignored) but I don't think it's very important at the moment.
This function could perhaps be added inside mgp::Value

cpp/map_module/algorithm/map.cpp
```diff
@@ -1 +1,86 @@
...
+    case mgp::Type::Int:
+      return std::to_string(value.ValueInt());
+    case mgp::Type::Double:
+      return std::to_string(value.ValueDouble());
+    case mgp::Type::String:
+      return std::string(value.ValueString());
+    case mgp::Type::List:
+      return "";
+    case mgp::Type::Map:
+      return "";
```
antoniofilipovic, line 24: Can we support that via recursive calls? I think it makes sense to add that as part of mgp::Value and maybe overload operator << for mgp::Value directly. What do you think?

antoniofilipovic (approved):

PR #273, opened 2023-07-31
[E-idora-create < T461-MAGE] Implement set properties for create functions

ind1xa:
### Description

Implemented create.set_properties and e2e correctness tests for it

### Pull request type

- [ ] Bugfix
- [x ] Algorithm/Module
- [ ] Feature
- [ ] Build related changes

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation


cpp/create_module/algorithm/create.cpp
```diff
@@ -1 +1,50 @@
...
+
+    if (arguments[0].IsList()) {
+      for (const auto element : arguments[0].ValueList()) {
+        if (element.IsNode()) {
+          nodeIds.insert(std::move(element.ValueNode().Id()));
+        } else if (element.IsInt()) {
+          nodeIds.insert(std::move(mgp::Id::FromInt(element.ValueInt())));
+        } else {
+          throw mgp::ValueException("First argument must be a node, node's id or a list of those.");
+        }
```
antoniofilipovic, line 19: I see your comment that you can isolate this part into function but can you explain about type casting? Can you just send mgp::Value and then check what you got there?

cpp/create_module/algorithm/create.cpp
```diff
@@ -1 +1,50 @@
...
+    const auto prop_value_list = arguments[2].ValueList();
+
+    if (prop_key_list.Size() != prop_value_list.Size()) {
+      throw mgp::IndexException("Key and value lists must be the same size.");
+    }
+
+    for (auto nodeId : nodeIds) {
+      auto node = graph.GetNodeById(nodeId);
+      for (size_t i = 0; i < prop_key_list.Size(); i++) {
+        node.SetProperty((std::string)prop_key_list[i].ValueString(), prop_value_list[i]);
```
antoniofilipovic, line 41: why do you have (std::string)? If you need to cast you can do following: `std::string(prop_key_list[i].ValueString())`

antoniofilipovic (changes requested):
Few changes

antoniofilipovic (changes requested):
I think we can avoid repetition in one more case

cpp/create_module/algorithm/create.cpp
```diff
@@ -0,0 +1,52 @@
...
+      throw mgp::IndexException("Key and value lists must be the same size.");
+    }
+
+    if (arguments[0].IsList()) {
+      for (const auto element : arguments[0].ValueList()) {
+        ProcessElement(element, graph, prop_key_list, prop_value_list, record_factory);
+      }
+      return;
+    }
+    ProcessElement(arguments[0], graph, prop_key_list, prop_value_list, record_factory);
```
antoniofilipovic, line 46:
```suggestion
    if (!arguments[0].IsList()) {
      ProcessElement(arguments[0], graph, prop_key_list, prop_value_list, record_factory);
      return;
    }
    for (const auto element : arguments[0].ValueList()) {
        ProcessElement(element, graph, prop_key_list, prop_value_list, record_factory);
    }

```

cpp/create_module/algorithm/create.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+void Create::ProcessElement(const mgp::Value &element, const mgp::Graph graph, const mgp::List &prop_key_list,
+                            const mgp::List &prop_value_list, const mgp::RecordFactory &record_factory) {
+  if (element.IsNode()) {
+    auto node = element.ValueNode();
+    SetElementProp(node, prop_key_list, prop_value_list, record_factory);
+    return;
+  }
+  if (element.IsInt()) {
+    auto node = graph.GetNodeById(mgp::Id::FromInt(element.ValueInt()));
```
antoniofilipovic, line 20: This may throw. You should wrap this ProcessElement in try catch block and in case of error, set Error message in case of error. Check Record::SetErrorMessage method

ind1xa: Isn't try-catch in main function(SetProperties) enough since that is the function which is calling ProcessElement? Right now if I send unexisting node id I get "Query failed: create.set_properties: Node with ID 90 not found!" error that seems informative enough? Or should I put something else?

antoniofilipovic: You are right, didn't see that, sorry

antoniofilipovic:
Wait for checks to pass

PR #276, opened 2023-07-31
[main < fix property comparison bug] Fix property comparison bug

mpintaric55334:
### Description

### Pull request type

- [ ] Bugfix
- [ ] Algorithm/Module
- [ ] Feature
- [ ] Build related changes

### Related issues

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation


e2e_correctness/query_neo_mem.py
```diff
@@ -58,6 +58,14 @@ def __eq__(self, other):
             if v != other._properties[k]:
                 logger.debug(f"Value {v} not equal to {other._properties[k]}")
                 return False
+            
+        for k, v in other._properties.items():
+            if k not in self._properties:
+                logger.debug(f"Property with key {k} not in {self._properties.keys()}")
+                return False
+            if v != self._properties[k]: #this if might be uselles, comment in review if you agree, this was checked in for loop above
+                logger.debug(f"Value {v} not equal to {self._properties[k]}")
```
antoniofilipovic, line 67: Generally I think we can do following:
1. find bigger map
2. iterate over that map

1. and 2. will cover if maps are equal and something is wrong. 

Add 3. check where we just check size of maps. If maps are of different size, no need to check again, just return false. Do you agree?

antoniofilipovic (changes requested):
I think we can make it more optimal, but good work finding a bug! :partying_face:

antoniofilipovic (approved):

PR #278, opened 2023-07-31
[E-idora-create < T458-MAGE] Implement remove labels for create functions

ind1xa:
### Description

Implemented create.remove_labels and e2e tests for it

### Pull request type

- [ ] Bugfix
- [x ] Algorithm/Module
- [ ] Feature
- [ ] Build related changes

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation


cpp/create_module/algorithm/create.cpp
```diff
@@ -0,0 +1,46 @@
...
+          throw mgp::ValueException("First argument must be a node, node's id or a list of those.");
+        }
+      }
+    } else if (arguments[0].IsNode()) {
+      nodeIds.insert(std::move(arguments[0].ValueNode().Id()));
+    } else if (arguments[0].IsInt()) {
+      nodeIds.insert(std::move(mgp::Id::FromInt(arguments[0].ValueInt())));
+    } else {
+      throw mgp::ValueException("First argument must be a node, node's id or a list of those.");
+    }
```
antoniofilipovic, line 28: Can you isolate this in function so it accepts mgp::Value?

antoniofilipovic (commented):
One discussion point

cpp/create_module/algorithm/create.cpp
```diff
@@ -0,0 +1,47 @@
...
+  } else if (argument.IsNode()) {
+    result_set.insert(std::move(argument.ValueNode().Id()));
+  } else if (argument.IsInt()) {
+    result_set.insert(std::move(mgp::Id::FromInt(argument.ValueInt())));
+  } else {
+    throw mgp::ValueException("First argument must be a node, node's id or a list of those.");
+  }
+  return result_set;
+}
```
antoniofilipovic, line 23: You can change this according to other function in PR: https://github.com/memgraph/mage/pull/273

antoniofilipovic (changes requested):
Refactor function and it should be good to go

cpp/create_module/algorithm/create.cpp
```diff
@@ -0,0 +1,47 @@
...
+    auto node = element.ValueNode();
+    RemoveElementLabels(node, labels, record_factory);
+    return;
+  }
+  if (element.IsInt()) {
+    auto node = graph.GetNodeById(mgp::Id::FromInt(element.ValueInt()));
+    RemoveElementLabels(node, labels, record_factory);
+    return;
+  }
+  throw mgp::ValueException("First argument must be type node, relationship, id or a list of those.");
```
antoniofilipovic, line 24: Add this check before other checks, so you can have this order.

pseudocode
```
check if not correct type ():
  throw errro // this is early return part 1
check if type node():
  do something...
  return // early return part 2
do other if int...
```

cpp/create_module/algorithm/create.cpp
```diff
@@ -0,0 +1,47 @@
...
+}
+
+void Create::ProcessElement(const mgp::Value &element, const mgp::Graph graph, const mgp::List &labels,
+                            const mgp::RecordFactory &record_factory) {
+  if (element.IsNode()) {
+    auto node = element.ValueNode();
+    RemoveElementLabels(node, labels, record_factory);
+    return;
+  }
+  if (element.IsInt()) {
```
antoniofilipovic, line 19: Can this int here be relationship id?

cpp/create_module/algorithm/create.cpp
antoniofilipovic, line 24: Where are we processing the relationship?

antoniofilipovic (changes requested):
One suggestion and few discussion points

cpp/create_module/algorithm/create.cpp
ind1xa: No it can't, sorry I copied the exception message from other procedure and forgot to change it

PR #281, opened 2023-08-01
[E-ivan < T463] Implement create.set_rel_properties

imilinovic:
### Description

Implement create.set_rel_properties

### Pull request type

- [ ] Bugfix
- [X] Algorithm/Module
- [ ] Feature
- [ ] Build related changes

### Related issues

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation


cpp/create_module/algorithm/create.cpp
```diff
@@ -0,0 +1,67 @@
...
+          relationship.SetProperty(std::string((*it1).ValueString()), *it2);
+          ++it1;
+          ++it2;
+        }
+        auto record = record_factory.NewRecord();
+        record.Insert(std::string(kResultSetRelProperties).c_str(), relationship);
+      }
+    }
+
+  } catch (const std::exception &e) {
```
antoniofilipovic, line 63: I think this is too big a complexity. I would have a different setup for when users pass a relationship as an object and when users pass an ID directly. In the case of an object, work with the object. In the case of ID, we need to go through all relationships. This is really bad, but if/when someone complains about sending IDs and setting properties we can later extend functionality for users to be able to pass vertex_id which contains the relationship as id, and set for that relationship that property. Then it will be easier to find relationship in graph.

antoniofilipovic (changes requested):
I would refactor this not to go trough all relationships unless really needed

antepusic (approved):
Looks good to go :rocket:

PR #290, opened 2023-08-04
[E-util < T580] Implement util.md5

mpintaric55334:
### Description

### Pull request type

- [ ] Bugfix
- [ ] Algorithm/Module
- [ ] Feature
- [ ] Build related changes

### Related issues

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation

### DIFF
--- a/cpp/util_module/algorithm/util.cpp
+++ b/cpp/util_module/algorithm/util.cpp
@@ -1,1 +1,21 @@
 #include "util.hpp"
+#include "md5.hpp"
+
+void Util::Md5(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result, mgp_memory *memory) {
+  mgp::memory = memory;
+  const auto arguments = mgp::List(args);
+  const auto record_factory = mgp::RecordFactory(result);
+  try {
+    const mgp::List arg_list = arguments[0].ValueList();
+    std::string return_string = "";
+    for (auto value : arg_list) {
+      return_string += value.ToString();
+    }
+    const std::string md5_hash = md5(return_string);
+    auto record = record_factory.NewRecord();
+    record.Insert(std::string(kArgumentResultMd5).c_str(), md5_hash);
+  } catch (const std::exception &e) {
+    record_factory.SetErrorMessage(e.what());
+    return;
+  }
+}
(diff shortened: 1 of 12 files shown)


cpp/util_module/algorithm/util.cpp
```diff
@@ -1 +1,21 @@
...
+#include "util.hpp"
+#include "md5.hpp"
+
+void Util::Md5(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result, mgp_memory *memory) {
+  mgp::memory = memory;
+  const auto arguments = mgp::List(args);
+  const auto record_factory = mgp::RecordFactory(result);
+  try {
+    const mgp::List arg_list = arguments[0].ValueList();
+    std::string return_string = "";
```
antoniofilipovic, line 10:
```suggestion
    std::string return_string{};
```

cpp/util_module/algorithm/util.cpp
```diff
@@ -1 +1,21 @@
...
+    for (auto value : arg_list) {
+      return_string += value.ToString();
+    }
+    const std::string md5_hash = md5(return_string);
+    auto record = record_factory.NewRecord();
+    record.Insert(std::string(kArgumentResultMd5).c_str(), md5_hash);
```
antoniofilipovic, line 16:
```suggestion
    auto record = record_factory.NewRecord();
    record.Insert(std::string(kArgumentResultMd5).c_str(),  md5(return_string));
```

e2e/util_module_test/util_test1/test.yml
```diff
@@ -0,0 +1,5 @@
+query: >
+    CALL util_module.md5(["string",[1,2,3]]) YIELD result RETURN result; 
+
+output:
+  - result: 7c45115a3283a86ecc2ad9b1b678cce0
```
antoniofilipovic, line 5: Can you make use of online tests, to use try multiple times to hash the same input to test we get same output? Check for example: e2e/node2vec_online_test/test_online_streamwalk_updater_missing
test to check how it works, and maybe we can have input as two queries which are empty, and for test call two times following input: 
CALL util_module.md5(["string",[1,2,3]]) YIELD result RETURN result;
to test we get both times same output.

antoniofilipovic (changes requested):
Few changes, but good work overall

antoniofilipovic (approved):
Looks good to me

PR #291, opened 2023-08-04
[main < T585 T586] Implement date functions

ind1xa:
### Description

Implemented date.parse and date.format procedures and e2e tests for them.

### Pull request type

- [ ] Bugfix
- [x] Algorithm/Module
- [ ] Feature
- [ ] Build related changes

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation


python/date_module.py
```diff
@@ -0,0 +1,87 @@
...
+        raise Exception("Unit doesn't exist. Check documentation to see available units.")
+
+    if (timezone in pytz.all_timezones):
+        offset, subtract = getOffset(timezone, new_date)
+        if (subtract):
+            tz_new = new_date - offset
+        else:
+            tz_new = new_date + offset
+    else:
+        raise Exception("Timezone doesn't exist. Check documentation to see available timezones.")
```
antoniofilipovic, line 85:
```suggestion
    if (timezone not in pytz.all_timezones):
        raise Exception("Timezone doesn't exist. Check documentation to see available timezones.")
    offset, subtract = getOffset(timezone, new_date)
    tz_new = new_date - offset if subtract else new_date + offset
```

python/date_module.py
```diff
@@ -0,0 +1,87 @@
...
+    input_date = datetime.datetime.strptime(time, format)
+
+    if (timezone in pytz.all_timezones):
+        offset, add = getOffset(timezone, input_date)
+        if (add):
+            tz_input = input_date + offset
+        else:
+            tz_input = input_date - offset
+    else:
+        raise Exception("Timezone doesn't exist. Check documentation to see available timezones.")
```
antoniofilipovic, line 36: Same comment as below. Check if you can isolate this part into  function

python/date_module.py
```diff
@@ -0,0 +1,87 @@
+import mgp
+import pytz
+import datetime
+
+MINUTES_IN_HOUR = 60
+SECONDS_IN_MINUTE = 60
+MILLISECONDS_IN_SECOND = 1000
+HOURS_IN_DAY = 24
+UNIX_EPOCH = datetime.datetime(1970, 1, 1, 0, 0, 0)
```
antoniofilipovic, line 9: For these constants we often make class Constants into which we store all of them, check tgn.py for example. Please adapt code to follow same style. In general I think this is not mistake, just to follow same code style :)

python/date_module.py
```diff
@@ -0,0 +1,87 @@
...
+
+def getOffset(timezone, date):
```
antoniofilipovic, line 11: Are there any Typings from pytz library we can use here? So that we have suggestion which type is sent here

antoniofilipovic (changes requested):
Few comments, but overall good work.

python/date_module.py
antoniofilipovic: Aha, typings yet. Yes sorry. So on Python, you have typings, https://docs.python.org/3/library/typing.html. Are there any available on `pytz` library which you can use in this case?

PR #296, opened 2023-08-07
[main < T584] Implement meta.stats

imilinovic:
### Description
Implement meta.stats online and offline version.

### Pull request type

- [ ] Bugfix
- [X] Algorithm/Module
- [ ] Feature
- [ ] Build related changes

#### Module/Algorithm
- [ ] Unit tests
- [x] End-to-end tests
- [ ] Code documentation


cpp/meta_module/algorithm/meta.cpp
```diff
@@ -0,0 +1,307 @@
...
+  for (const auto &[property, _] : node.Properties()) {
+    insert(property_key_cnt, property, add);
+  }
+}
+
+void Metadata::update_property_key_cnt(const mgp::Relationship &relationship, int add) {
+  for (const auto &[property, _] : relationship.Properties()) {
+    insert(property_key_cnt, property, add);
+  }
+}
```
imilinovic, line 88: Could be templated but is it necessary for such small function?

cpp/meta_module/meta_module.cpp
```diff
@@ -0,0 +1,44 @@
...
+                 {}, module, memory);
+
+    AddProcedure(Meta::StatsOnline, Meta::kProcedureStatsOnline, mgp::ProcedureType::Read,
+                 {mgp::Parameter(Meta::kStatsOnlineArg1, mgp::Type::Bool)},
+                 {mgp::Return(Meta::kReturnStats1, mgp::Type::Int), mgp::Return(Meta::kReturnStats2, mgp::Type::Int),
+                  mgp::Return(Meta::kReturnStats3, mgp::Type::Int), mgp::Return(Meta::kReturnStats4, mgp::Type::Int),
+                  mgp::Return(Meta::kReturnStats5, mgp::Type::Int), mgp::Return(Meta::kReturnStats6, mgp::Type::Map),
+                  mgp::Return(Meta::kReturnStats7, mgp::Type::Map), mgp::Return(Meta::kReturnStats8, mgp::Type::Map),
+                  mgp::Return(Meta::kReturnStats9, mgp::Type::Map)},
+                 module, memory);
```
imilinovic, line 25: This throws a warning for some reason, unsure why: 
`note: ‘<anonymous>’ declared here`

cpp/meta_module/algorithm/meta.cpp
```diff
@@ -0,0 +1,299 @@
...
+  record.Insert(std::string(kReturnStats1).c_str(), label_count);
+  stats.Insert("labelCount", mgp::Value(label_count));
+
+  int64_t relationship_type_count = data.get_relationship_type_count();
+  record.Insert(std::string(kReturnStats2).c_str(), relationship_type_count);
+  stats.Insert("relationshipTypeCount", mgp::Value(relationship_type_count));
+
+  int64_t property_key_count = data.get_property_key_count();
+  record.Insert(std::string(kReturnStats3).c_str(), property_key_count);
+  stats.Insert("propertyKeyCount", mgp::Value(property_key_count));
```
imilinovic, line 136: Unsure to add lambda function for this

cpp/meta_module/algorithm/meta.cpp
```diff
@@ -0,0 +1,284 @@
...
+
+namespace Meta {
+
+namespace {
+
+class Metadata {
+ public:
+  Metadata() { node_cnt = relationship_cnt = 0; }
+
+  int64_t node_cnt;
```
antoniofilipovic, line 15: Move these vars after functions

cpp/meta_module/algorithm/meta.cpp
```diff
@@ -0,0 +1,284 @@
...
+  void update_property_key_cnt(const mgp::Node &node, int add);
+  void update_property_key_cnt(const mgp::Relationship &relationship, int add);
+  void update_relationship_types(const mgp::Relationship &relationship, int add);
+  void update_relationship_types_cnt(const mgp::Relationship &relationship, int add);
+};
+
+// Global variable
+// This is unsafe in the multithreaded environment, the workaround would be building a thread-safe dynamic storage
+// implementation
+Metadata metadata;
```
antoniofilipovic, line 37: I think you are right, but we can do one other thing in my opinion. We can have lock per variable, and in that case when you want to access `labels` you take `lables` lock. You update labels and that should be all right. What will we do with that? I think we will have thread safe, but not "fully correct" implementation. Because if one thread updates labels, and other updates properties, it might happen that properties thread will get update of labels, but it didn't make such change. And that is all right, because it actually takes into account commits, and it works as is. I think it is more granular this way than using one lock for whole object. Also, for counters `relationship_cnt` and `node_cnt` we can use atomics.

antoniofilipovic (commented):
I think we need to take into account multi-threading

cpp/meta_module/algorithm/meta.cpp
antoniofilipovic: @Josipmrden @antepusic tagging you here too

cpp/meta_module/algorithm/meta.cpp
```diff
@@ -0,0 +1,284 @@
+#include "meta.hpp"
```
Josipmrden, line 1: new line break after line 1

cpp/meta_module/algorithm/meta.cpp
```diff
@@ -0,0 +1,284 @@
...
+#include <cstdint>
+#include <string>
+#include <string_view>
+#include "mgp.hpp"
```
Josipmrden, line 5: newline break after line 4

Josipmrden: the structure should be 1st the header file of the file followed by std libraries and others, followed by local imports like `mgp.hpp`

e2e/meta_online_test/test_online_stats_update_4/input.cyp
```diff
@@ -0,0 +1,19 @@
+setup: |-
+  CALL meta.reset();
+
+queries:
+  - |-
+    MERGE (a:Node {id: 0}) MERGE (b:Node {id: 1}) CREATE (a)-[:RELATION]->(b);
+    MERGE (a:Node {id: 1}) MERGE (b:Node {id: 2}) CREATE (a)-[:RELATION]->(b);
+    MERGE (a:Node {id: 2}) MERGE (b:Node {id: 0}) CREATE (a)-[:RELATION]->(b);
+    MERGE (a:Node {id: 3}) MERGE (b:Node {id: 3}) CREATE (a)-[:RELACIJA]->(b);
```
Josipmrden, line 9: please restrict to english language :(, it can be anything you want, but keep it english if it does not signify first name and last name

cpp/meta_module/algorithm/meta.cpp
```diff
@@ -0,0 +1,284 @@
...
+
+namespace Meta {
+
+namespace {
+
+class Metadata {
+ public:
+  Metadata() { node_cnt = relationship_cnt = 0; }
```
Josipmrden, line 13:
```suggestion
  Metadata() : node_cnt{0}, relationship_cnt{0} {}
```

cpp/meta_module/algorithm/meta.cpp
```diff
@@ -0,0 +1,284 @@
...
+      map.erase(iterator);
+    }
+  } else {
+    map[key] = add;
+  }
+}
+
+void insert(std::unordered_map<std::string, int64_t> &map, std::string_view key_view, int64_t add) {
+  auto key = std::string(key_view);
+  insert(map, key, add);
```
Josipmrden, line 53:
```suggestion
  insert(map, std::move(key), add);
```

cpp/meta_module/algorithm/meta.cpp
```diff
@@ -0,0 +1,284 @@
...
+  void update_relationship_types(const mgp::Relationship &relationship, int add);
+  void update_relationship_types_cnt(const mgp::Relationship &relationship, int add);
+};
+
+// Global variable
+// This is unsafe in the multithreaded environment, the workaround would be building a thread-safe dynamic storage
+// implementation
+Metadata metadata;
+
+void insert(std::unordered_map<std::string, int64_t> &map, std::string &key, int64_t add) {
```
Josipmrden, line 39: key can be passed by value and then moved in the calling operations

cpp/meta_module/algorithm/meta.cpp
Josipmrden: or just remove this function, seems unnecessary. If you need std::string you can even cast it in the upper function

cpp/meta_module/algorithm/meta.cpp
```diff
@@ -0,0 +1,284 @@
...
+  std::unordered_map<std::string, int64_t> property_key_cnt;
+  std::unordered_map<std::string, int64_t> relationship_types;
+  std::unordered_map<std::string, int64_t> relationship_types_cnt;
+
+  int64_t get_label_count() const { return static_cast<int64_t>(labels.size()); }
+  int64_t get_relationship_type_count() const { return static_cast<int64_t>(relationship_types_cnt.size()); }
+  int64_t get_property_key_count() const { return static_cast<int64_t>(property_key_cnt.size()); }
+
+  void reset();
+  void update_labels(const mgp::Node &node, int add);
```
Josipmrden, line 27: pascal case for class functions

cpp/meta_module/algorithm/meta.cpp
```diff
@@ -0,0 +1,284 @@
...
+void Metadata::update_relationship_types_cnt(const mgp::Relationship &relationship, int add) {
+  insert(relationship_types_cnt, relationship.Type(), add);
+}
+
+Metadata get_all_metadata(mgp_graph *memgraph_graph) {
+  Metadata data;
+
+  const mgp::Graph graph{memgraph_graph};
+
+  for (const auto node : graph.Nodes()) {
```
Josipmrden, line 106: &node

cpp/meta_module/algorithm/meta.hpp
```diff
@@ -0,0 +1,42 @@
...
+constexpr const std::string_view kStatsOnlineArg1 = "update_stats";
+constexpr const std::string_view kReturnStats1 = "labelCount";
+constexpr const std::string_view kReturnStats2 = "relationshipTypeCount";
+constexpr const std::string_view kReturnStats3 = "propertyKeyCount";
+constexpr const std::string_view kReturnStats4 = "nodeCount";
+constexpr const std::string_view kReturnStats5 = "relationshipCount";
+constexpr const std::string_view kReturnStats6 = "labels";
+constexpr const std::string_view kReturnStats7 = "relationshipTypes";
+constexpr const std::string_view kReturnStats8 = "relationshipTypesCount";
+constexpr const std::string_view kReturnStats9 = "stats";
```
Josipmrden, line 29: https://learn.microsoft.com/en-us/cpp/cpp/constexpr-cpp?view=msvc-170 all `constexpr` variables are const

cpp/meta_module/algorithm/meta.hpp
```diff
@@ -0,0 +1,42 @@
+#pragma once
+
+#include <mgp.hpp>
+#include <string_view>
+
+namespace Meta {
+
+/* update constants */
+constexpr const std::string_view kProcedureUpdate = "update";
```
Josipmrden, line 9:
```suggestion
constexpr std::string_view kProcedureUpdate = "update";
```

Josipmrden (commented):
Some comments, overall good job

cpp/meta_module/algorithm/meta.cpp
antoniofilipovic: Actually doesn't need &  here this is actually copied under the hood.

cpp/meta_module/meta_module.cpp
imilinovic: Changed but still getting the warning.

cpp/meta_module/algorithm/meta.cpp
```diff
@@ -0,0 +1,330 @@
...
+
+void StatsOffline(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result, mgp_memory *memory) {
+  mgp::memory = memory;
+  const auto arguments = mgp::List(args);
+  const auto record_factory = mgp::RecordFactory(result);
+
+  try {
+    Metadata data;
+    UpdateAllMetadata(memgraph_graph, data, false);
+    OutputFromMetadata(data, record_factory, false);
```
antepusic, line 309: I think `StatsOffline` should also reset metadata because of the case that users have already run `stats_online` before and saved metadata thus exists.

cpp/meta_module/algorithm/meta.hpp
```diff
@@ -0,0 +1,42 @@
...
+constexpr std::string_view kUpdateArg1 = "createdObjects";
+constexpr std::string_view kUpdateArg2 = "deletedObjects";
+constexpr std::string_view kUpdateArg3 = "removedVertexProperties";
+constexpr std::string_view kUpdateArg4 = "removedEdgeProperties";
+constexpr std::string_view kUpdateArg5 = "setVertexLabels";
+constexpr std::string_view kUpdateArg6 = "removedVertexLabels";
+
+/* stats constants */
+constexpr std::string_view kProcedureStatsOnline = "stats_online";
+constexpr std::string_view kProcedureStatsOffline = "stats_offline";
```
antepusic, line 19: In MAGE, the `_online` and `_offline` procedures are named `set` and `get` -> in this case `stats_set` and `stats_get`.

antepusic (changes requested):
Good job! I think there's one additional case where metadata need to be reset (see comment in meta.cpp).

cpp/meta_module/algorithm/meta.cpp
imilinovic: If you have already ran the online version there is no reason to run the offline reason because the online version works in O(1). Also offline version doesn't modify the global metadata object.

cpp/meta_module/algorithm/meta.cpp
```diff
@@ -0,0 +1,330 @@
...
+  std::unordered_map<std::string, int64_t> relationship_types;
+  std::unordered_map<std::string, int64_t> relationship_types_cnt;
+};
+
+struct MetadataMutex {
+  std::mutex labels_mutex;
+  std::mutex property_key_cnt_mutex;
+  std::mutex relationship_types_mutex;
+  std::mutex relationship_types_cnt_mutex;
+};
```
antoniofilipovic, line 35: I think it makes sense to have this mutex as part of `Metadata`

cpp/meta_module/algorithm/meta.cpp
```diff
@@ -0,0 +1,330 @@
...
+
+// Global variable
+Metadata metadata;
+MetadataMutex metadata_mutex;
+
+void LockAndExecute(std::mutex &mutex, bool online, auto Function) {
```
antoniofilipovic, line 41: I think a better name for this is `MaybeLockAndExecute`, because you are maybe locking. Also, naming of `Function` should be `function` (lowercase)

cpp/meta_module/algorithm/meta.cpp
```diff
@@ -0,0 +1,330 @@
...
+  auto ClearMap = [](std::mutex &mutex, auto &map) { LockAndExecute(mutex, true, [&]() { map.clear(); }); };
+
+  ClearMap(metadata_mutex.labels_mutex, labels);
+  ClearMap(metadata_mutex.property_key_cnt_mutex, property_key_cnt);
+  ClearMap(metadata_mutex.relationship_types_mutex, relationship_types);
+  ClearMap(metadata_mutex.relationship_types_cnt_mutex, relationship_types_cnt);
+}
+
+void Metadata::UpdateLabels(const mgp::Node &node, int add, bool online) {
+  auto IterateAndInsert = [&]() {
```
antoniofilipovic, line 83: This should be lowercase with _, so `iterate_insert`

cpp/meta_module/algorithm/meta.cpp
```diff
@@ -0,0 +1,330 @@
...
+      auto key = std::string(label);
+      Insert(labels, key, add);
+    }
+  };
+
+  LockAndExecute(metadata_mutex.labels_mutex, online, IterateAndInsert);
+}
+
+void Metadata::UpdatePropertyKeyCnt(const auto &node_or_relationship, int add, bool online) {
+  auto IterateAndInsert = [&]() {
```
antoniofilipovic, line 94: Same comment considering naming

cpp/meta_module/algorithm/meta.cpp
```diff
@@ -0,0 +1,330 @@
...
+  }
+}
+
+void StatsOffline(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result, mgp_memory *memory) {
+  mgp::memory = memory;
+  const auto arguments = mgp::List(args);
+  const auto record_factory = mgp::RecordFactory(result);
+
+  try {
+    Metadata data;
```
antoniofilipovic, line 307: You can change this name to `local_metadata`

antoniofilipovic (changes requested):
Few changes and that should be it

cpp/meta_module/algorithm/meta.cpp
```diff
@@ -0,0 +1,330 @@
...
+
+  try {
+    const auto created_objects{arguments[0].ValueList()};
+    const auto deleted_objects{arguments[1].ValueList()};
+    const auto removed_vertex_properties{arguments[2].ValueList()};
+    const auto removed_edge_properties{arguments[3].ValueList()};
+    const auto set_vertex_labels{arguments[4].ValueList()};
+    const auto removed_vertex_labels{arguments[5].ValueList()};
+
+    auto UpdateObject = [](const auto &objects, std::string_view vertex_event, std::string_view edge_event, int add) {
```
antoniofilipovic, line 216: Make this vertex_event and edge_event hardcoded as constant, there is no need to pass it as function parameter as it is not changing anywhere

imilinovic: It changes depending whether it is called on created_objects or deleted_objects

antoniofilipovic (approved):
Looks good to me

antepusic (approved):

PR #300, opened 2023-08-08
[E-node < T501] Implement node.relationship_types

imilinovic:
### Description

Implement node.relationship_types

### Pull request type

- [ ] Bugfix
- [X] Algorithm/Module
- [ ] Feature
- [ ] Build related changes

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation

### DIFF
--- a/cpp/node_module/algorithm/node.cpp
+++ b/cpp/node_module/algorithm/node.cpp
@@ -1,9 +1,6 @@
 #include "node.hpp"
-
-#include <cstdint>
-#include <sstream>
-#include <string_view>
 #include <unordered_set>
+
 #include "mgp.hpp"
 
 namespace {
--- a/cpp/node_module/algorithm/node.cpp
+++ b/cpp/node_module/algorithm/node.cpp
@@ -1,1 +1,72 @@
 #include "node.hpp"
+
+#include <cstdint>
+#include <sstream>
+#include <string_view>
+#include <unordered_set>
+#include "mgp.hpp"
+
+namespace {
+
+std::unordered_map<std::string_view, uint8_t> get_type_direction(const mgp::Value &types) {
+  std::unordered_map<std::string_view, uint8_t> result;
+  for (const auto &type_value : types.ValueList()) {
+    auto type = type_value.ValueString();
+    if (type.starts_with('<')) {
+      result[type.substr(1, type.size() - 1)] |= 1;
+    } else if (type.ends_with('>')) {
+      result[type.substr(0, type.size() - 1)] |= 2;
+    } else {
+      result[type] |= 3;
+    }
+  }
+  return result;
+}
+
+mgp::List get_relationship_types(const mgp::Value &node_value, const mgp::Value &types_value) {
+  auto type_direction = get_type_direction(types_value);
+
+  std::unordered_set<std::string_view> types;
+  const auto node = node_value.ValueNode();
+  if (type_direction.empty()) {
+    for (const auto relationship : node.InRelationships()) {
+      types.insert(relationship.Type());
+    }
+    for (const auto relationship : node.OutRelationships()) {
+      types.insert(relationship.Type());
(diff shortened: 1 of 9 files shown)


cpp/node_module/node_module.cpp
```diff
@@ -6,6 +6,12 @@ extern "C" int mgp_init_module(struct mgp_module *module, struct mgp_memory *mem
   try {
     mgp::memory = memory;
 
+    AddProcedure(
+        Node::RelationshipTypes, Node::kProcedureRelationshipTypes, mgp::ProcedureType::Read,
+        {mgp::Parameter(Node::kRelationshipTypesArg1, mgp::Type::Node),
+         mgp::Parameter(Node::kRelationshipTypesArg2, {mgp::Type::List, mgp::Type::String}, mgp::Value(mgp::List{}))},
+        {mgp::Return(Node::kResultRelationshipTypes, {mgp::Type::List, mgp::Type::String})}, module, memory);
```
antepusic, line 13: Going off APOC documentation ([link 1](https://neo4j.com/labs/apoc/4.1/overview/apoc.node/apoc.node.relationship.types/) & [link 2](https://neo4j-contrib.github.io/neo4j-apoc-procedures/3.5/nodes-relationships/node-functions/)), I see that the second argument is not a list of strings, but rather a string with relationship types joined by `"|"`, e.g. `"FRIEND|MENTORS>|<REPORTS_TO"`. Can you confirm if the difference in implementation is OK from a  compatibility perspective?

antepusic: @antoniofilipovic I'd also appreciate your opinion

cpp/node_module/algorithm/node.cpp
```diff
@@ -1 +1,69 @@
...
+  std::unordered_map<std::string_view, uint8_t> result;
+  for (const auto &type_value : types.ValueList()) {
+    auto type = type_value.ValueString();
+    if (type.starts_with('<')) {
+      result[type.substr(1, type.size() - 1)] |= 1;
+    } else if (type.ends_with('>')) {
+      result[type.substr(0, type.size() - 1)] |= 2;
+    } else {
+      result[type] |= 3;
+    }
```
antepusic, line 18: The case where `type` both starts with `'<'` and ends with `'>'` (e.g. '<KNOWS>') should also be addressed.

antepusic (changes requested):
The code looks good, but I'd like to see two things addressed before approving:
* the difference in function signature between us and APOC
* if the user gives e.g. "<KNOWS>" as a relationship type, that should be handled appropriately

cpp/node_module/node_module.cpp
imilinovic: We decided to use a list instead of splitting by | for simplicity.

cpp/node_module/algorithm/node.cpp
imilinovic: We internally discussed it and decided that <> case isn't needed because it it is just like without the <> (since it covers both in and out nodes). Will discuss tomorrow on daily to either throw error or allow it.

antepusic: I've checked that and you're right about it :smile:. However, if the user gives e.g. `"<KNOWS>"`, the current code would treat that like `"<KNOWS"` and not `"KNOWS"`.

imilinovic: My bad, we already said to throw an error. Fixed

cpp/node_module/node_module.cpp
antoniofilipovic: Yes, we decided to use list for simplicity

antepusic (approved):
Thanks for the clarification about procedure signature. I think this is good to go now :rocket:

PR #302, opened 2023-08-08
[E-nodes < T478] Implement nodes.delete

imilinovic:
### Description

Implement nodes.delete

### Pull request type

- [ ] Bugfix
- [x] Algorithm/Module
- [ ] Feature
- [ ] Build related changes

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation

### DIFF
--- a/cpp/nodes_module/algorithm/nodes.cpp
+++ b/cpp/nodes_module/algorithm/nodes.cpp
@@ -1,1 +1,43 @@
 #include "nodes.hpp"
+#include "mgp.hpp"
+
+namespace {
+void throw_exception(const mgp::Value &value) {
+  std::ostringstream oss;
+  oss << value.Type();
+  throw mgp::ValueException("Unsuppported type for this operation, received type: " + oss.str());
+};
+
+}  // namespace
+
+void Nodes::Delete(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result, mgp_memory *memory) {
+  mgp::memory = memory;
+  const auto arguments = mgp::List(args);
+  const auto record_factory = mgp::RecordFactory(result);
+  try {
+    mgp::Graph graph{memgraph_graph};
+    auto nodes{arguments[0]};
+
+    if (nodes.IsInt()) {
+      graph.DetachDeleteNode(graph.GetNodeById(mgp::Id::FromInt(nodes.ValueInt())));
+    } else if (nodes.IsNode()) {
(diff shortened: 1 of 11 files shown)


cpp/nodes_module/algorithm/nodes.cpp
```diff
@@ -1 +1,43 @@
...
+          graph.DetachDeleteNode(graph.GetNodeById(mgp::Id::FromInt(list_item.ValueInt())));
+        } else if (list_item.IsNode()) {
+          graph.DetachDeleteNode(list_item.ValueNode());
+        } else {
+          throw_exception(list_item);
+        }
+      }
+    } else {
+      throw_exception(nodes);
+    }
```
antepusic, line 37: Not a big deal, but I'd recommend wrapping the `IsInt` and `IsNode` checks in a function to avoid repeating them. Something like this:
```
void DetachDeleteNode(mgp::Value &node) {
  if (node.IsInt()) {
    graph.DetachDeleteNode(graph.GetNodeById(mgp::Id::FromInt(node.ValueInt())));
  } else if (node.IsNode()) {
    graph.DetachDeleteNode(node.ValueNode());
  } else {
    throw_exception(nodes);
  }
}
```

antepusic (commented):
Looks good! I've left a comment about avoiding code repetition, but the current state of the module is fine the way it is.

cpp/nodes_module/algorithm/nodes.cpp
```diff
@@ -2,12 +2,22 @@
 #include "mgp.hpp"
 
 namespace {
-void throw_exception(const mgp::Value &value) {
+void ThrowException(const mgp::Value &value) {
   std::ostringstream oss;
   oss << value.Type();
   throw mgp::ValueException("Unsuppported type for this operation, received type: " + oss.str());
 };
```
antepusic, line 9: Since there's only one point in code where `ThrowException` is called, perhaps we could do away with this function and move its body there.

imilinovic: I think that function will also be used in other nodes procedures. Ideally I would put it some kind of utils (but I don't think the current structure uses that) or even cpp-api mgp::Value since its used very often.

antepusic: I see, thanks!

antepusic (approved):
I think we're good to go now!

PR #301, opened 2023-08-08
[E-nodes < T502] Implement nodes.relationship_types

imilinovic:
### Description

Implement nodes.relationship_types

### Pull request type

- [ ] Bugfix
- [X] Algorithm/Module
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

######################################

#### Module/Algorithm
- [ ] Core algorithm/module implementation
- [ ] [Query module](https://memgraph.com/docs/memgraph/reference-guide/query-modules) implementation
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation
- [ ] README short description
- [ ] Documentation on [memgraph/docs](https://github.com/memgraph/docs)
- [ ] Update GQLALchemy signatures in [query builder](https://github.com/memgraph/gqlalchemy/blob/main/gqlalchemy/graph_algorithms/query_builder.py  ) using [query module signature generator](https://github.com/memgraph/gqlalchemy/blob/main/scripts/query_module_signature_generator.py)

######################################

PR #298, opened 2023-08-07
[E-node < T499] Implement node.relationshipsExist

mpintaric55334:
### Description

Please briefly explain the changes you made here.

### Pull request type

- [ ] Bugfix
- [ ] Algorithm/Module
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

### Related issues

Delete if this PR doesn't resolve any issues. Link the issue if it does.

######################################

#### Module/Algorithm
- [ ] Core algorithm/module implementation
- [ ] [Query module](https://memgraph.com/docs/memgraph/reference-guide/query-modules) implementation
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation
- [ ] README short description
- [ ] Documentation on [memgraph/docs](https://github.com/memgraph/docs)
- [ ] Update GQLALchemy signatures in [query builder](https://github.com/memgraph/gqlalchemy/blob/main/gqlalchemy/graph_algorithms/query_builder.py  ) using [query module signature generator](https://github.com/memgraph/gqlalchemy/blob/main/scripts/query_module_signature_generator.py)

######################################

PR #303, opened 2023-08-09
[E-node < T498] Implement relationship_exists

ind1xa:
### Description

Implemented node.relationship_exists procedure.

### Pull request type

- [ ] Bugfix
- [x] Algorithm/Module
- [ ] Feature
- [ ] Build related changes

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation

### DIFF
--- a/cpp/node_module/algorithm/node.cpp
+++ b/cpp/node_module/algorithm/node.cpp
@@ -45,8 +45,8 @@ void Node::RelExists(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *resu
         out_rels.insert(pattern.substr(0, pattern.size() - 1));
         continue;
       }
-      in_rels.insert(pattern);
-      out_rels.insert(pattern);
+      in_rels.insert(std::move(pattern));
+      out_rels.insert(std::move(pattern));
     }
 
     auto record = record_factory.NewRecord();
--- a/cpp/node_module/algorithm/node.cpp
+++ b/cpp/node_module/algorithm/node.cpp
@@ -19,16 +19,15 @@ void Node::RelExists(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *resu
   try {
     const auto node = arguments[0].ValueNode();
     const auto pattern = arguments[1].ValueString();
-
     bool exists = false;
 
     if (pattern[0] == '<' && pattern[pattern.size() - 1] == '>') {
       exists = (FindRelationship(pattern.substr(1, pattern.size() - 2), node.InRelationships()) &&
                 FindRelationship(pattern.substr(1, pattern.size() - 2), node.OutRelationships()));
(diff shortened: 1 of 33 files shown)


cpp/node_module/algorithm/node.cpp
```diff
@@ -1 +1,43 @@
...
+  const auto arguments = mgp::List(args);
+  const auto record_factory = mgp::RecordFactory(result);
+  try {
+    const auto node = arguments[0].ValueNode();
+    const auto pattern = arguments[1].ValueString();
+    bool exists = false;
+
+    if (pattern[0] == '<' && pattern[pattern.size() - 1] == '>') {
+      throw mgp::ValueException("Invalid relationship specification!");
+    }
```
antepusic, line 26: This should be consistent with `node.relationship_types` (#300) and `nodes.relationship_types` (#301) - I think the <> case is the same as the bidirectional relationship case.

cpp/node_module/node_module.cpp
```diff
@@ -5,6 +5,10 @@
 extern "C" int mgp_init_module(struct mgp_module *module, struct mgp_memory *memory) {
   try {
     mgp::memory = memory;
+    AddProcedure(Node::RelExists, Node::kProcedureRelExists, mgp::ProcedureType::Read,
+                 {mgp::Parameter(Node::kArgumentsNode, mgp::Type::Node),
+                  mgp::Parameter(Node::kArgumentsPattern, mgp::Type::String, "")},
```
antepusic, line 10: We should also accommodate the case where multiple patterns are given (#300 and #301 use a list for the second parameter).

antepusic (changes requested):
The code looks good! However, there are two cases where this procedure should behave consistently with #300 and #301 and I've left comments on that.

cpp/node_module/algorithm/node.cpp
ind1xa: We decided to throw error in this case

cpp/node_module/node_module.cpp
ind1xa: That is the only difference between node.relationship_exists and node.relationships_exist, if I do this, those will be the same procedures right? https://github.com/memgraph/mage/pull/298/files

cpp/node_module/algorithm/node.cpp
antepusic: I see, thanks!

cpp/node_module/node_module.cpp
antepusic: The return would still be different. `relationships_exist` returns the existence of a relationship for _each_ of the given patterns, whereas this method returns whether there is a relationship that fits `at least one` of the given pattern(s).

antepusic (approved):

PR #306, opened 2023-08-10
[E-neighbors < T486] Implement athop

ind1xa:
### Description

Implemented neighbors.athop procedure.

### Pull request type

- [ ] Bugfix
- [x] Algorithm/Module
- [ ] Feature
- [ ] Build related changes

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation


cpp/neighbors_module/neighbors_module.cpp
```diff
@@ -5,6 +5,11 @@
 extern "C" int mgp_init_module(struct mgp_module *module, struct mgp_memory *memory) {
   try {
     mgp::memory = memory;
+    AddProcedure(Neighbors::AtHop, Neighbors::kProcedureAtHop, mgp::ProcedureType::Read,
+                 {mgp::Parameter(Neighbors::kArgumentsNode, mgp::Type::Node),
+                  mgp::Parameter(Neighbors::kArgumentsRelType, mgp::Type::String),
+                  mgp::Parameter(Neighbors::kArgumentsDistance, mgp::Type::Int)},
+                 {mgp::Return(Neighbors::kReturnAtHop, mgp::Type::Node)}, module, memory);
```
antepusic, line 12: In APOC, apoc.neighbors.athop can take a relationship filter (`rel_types`) that accepts multiple types of relationships ([docs](https://neo4j.com/labs/apoc/4.3/graph-querying/neighborhood/#neighborhood-search-rel-filters)).
As there are already several PRs for functions that use relationship filters, e.g. #300 and #301, I suggest changing the second argument of this function so that it works the same way as in those PRs (i.e. as a list of strings).

antepusic (changes requested):
The code looks good! I've left one comment so that using this query procedure is consistent with others that use relationship filters.

cpp/neighbors_module/algorithm/neighbors.cpp
```diff
@@ -1 +1,92 @@
...
+      }
+      if (rel_type[rel_type.size() - 1] == '>') {
+        out_rels.insert(rel_type.substr(0, rel_type.size() - 1));
+        continue;
+      }
+      in_rels.insert(std::move(rel_type));
+      out_rels.insert(std::move(rel_type));
+    }
+
+    std::list<std::unordered_set<mgp::Node>> list;
```
antoniofilipovic, line 54: Why `std::list`, can you use `std::vector` here? It makes sense to use list if you are moving from one list to another, in that case, you avoid constructors.

cpp/neighbors_module/algorithm/neighbors.cpp
```diff
@@ -1 +1,92 @@
...
+            }
+          }
+        }
+        if (!out_rels.empty()) {
+          for (auto relationship : node.OutRelationships()) {
+            if ((out_rels.contains("") || out_rels.contains(relationship.Type())) && !Known(relationship.To(), list)) {
+              set.insert(relationship.To());
+            }
+          }
+        }
```
antoniofilipovic, line 75: These two parts are really similar, please extract it to some function. For example function which accepts container and type and node, and returns whether check passed

cpp/neighbors_module/algorithm/neighbors.cpp
```diff
@@ -1 +1,92 @@
...
+        in_rels.insert(rel_type.substr(1, rel_type.size()));
+        continue;
+      }
+      if (rel_type[rel_type.size() - 1] == '>') {
+        out_rels.insert(rel_type.substr(0, rel_type.size() - 1));
+        continue;
+      }
+      in_rels.insert(std::move(rel_type));
+      out_rels.insert(std::move(rel_type));
+    }
```
antoniofilipovic, line 52: This can be part of some kind of preprocessing function, makes sense to extract that

cpp/neighbors_module/algorithm/neighbors.cpp
```diff
@@ -1 +1,92 @@
...
+  const auto arguments = mgp::List(args);
+  const auto record_factory = mgp::RecordFactory(result);
+  try {
+    const auto node = arguments[0].ValueNode();
+    auto input_rel_type = arguments[1].ValueString();
+    const auto distance = arguments[2].ValueInt();
+
+    std::unordered_set<std::string_view> rel_types;
+    auto i{0};
+    auto j{input_rel_type.find("|")};
```
antoniofilipovic, line 27: Nice! :smile:

antoniofilipovic (changes requested):
Few changes, but good work

cpp/neighbors_module/algorithm/neighbors.cpp
ind1xa: Hm I'm not quite sure what do you mean, they are not that similar since inserting relationship.From() or relationship.To() node depends on whether I'm iterating over in or out relationships so I would still have to check that when extracting function

cpp/neighbors_module/algorithm/neighbors.cpp
```diff
@@ -1 +1,91 @@
...
+    }
+  }
+  return false;
+}
+
+void DetermineDirection(mgp::List &rel_types, std::unordered_set<std::string_view> &in_rels,
+                        std::unordered_set<std::string_view> &out_rels) {
+  if (rel_types.Empty()) {
+    rel_types.AppendExtend(mgp::Value(""));
+  }
```
antepusic, line 19: Why is this necessary? Just asking because later on it ends up in `in_rels` and `out_rels` and there are several occasions where you check if those two are empty

cpp/neighbors_module/algorithm/neighbors.cpp
antepusic: I think that extracting this into a function is non-trivial because the function would need to e.g. call the right method (`InRelationships` vs. `OutRelationships` and `To` vs. `From`) depending on a relationship direction parameter.
IMHO this is mergeable but figuring out a neat way to handle cases like this one will be useful for MAGE development in general.

antoniofilipovic: ah, okay, works then

antoniofilipovic (approved):

PR #309, opened 2023-08-10
[E-ivan < T460] Implement create.remove_rel_properties

imilinovic:
### Description

Implement create.remove_rel_properties procedure

### Pull request type

- [ ] Bugfix
- [X] Algorithm/Module
- [ ] Feature
- [ ] Build related changes

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation


cpp/create_module/algorithm/create.cpp
```diff
@@ -1,8 +1,73 @@
...
+        DeleteAndOutput(relationship_copy, keys, record_factory);
+      } else if (value.IsInt()) {
+        ids.insert(value.ValueInt());
+      } else {
+        ThrowException(value);
+      }
+    };
+
+    if (!arguments[0].IsList()) {
+      ParseValue(arguments[0]);
```
antoniofilipovic, line 47: You can return here, right?

cpp/create_module/algorithm/create.cpp
```diff
@@ -1,8 +1,73 @@
...
+      }
+    }
+
+    if (ids.empty()) {
+      return;
+    }
+
+    for (const auto relationship : graph.Relationships()) {
+      if (ids.contains(relationship.Id().AsInt())) {
+        auto relationship_copy = relationship;
```
antoniofilipovic, line 60: Why do you need here copy?

antoniofilipovic (changes requested):
Few changes

cpp/create_module/algorithm/create.cpp
imilinovic: No because ParseValue can insert ids into the unordered_set (if id is given). I could add a check
`if (arguments[0].IsInt()) return...`
but I don't think its needed because few lines later that is checked

cpp/create_module/algorithm/create.cpp
imilinovic: I don't :)

antepusic (approved):
I think we're good here :rocket: 

NB: reviewed the PR at commit da35928 because the final commit (merging with parent) messed up the diff

PR #317, opened 2023-08-11
[E-path < T488] Add path.create

imilinovic:
### Description

Add path.create

### Pull request type

- [ ] Bugfix
- [ ] Algorithm/Module
- [ ] Feature
- [ ] Build related changes

### Related issues

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation


cpp/path_module/path_module.cpp
```diff
@@ -6,6 +6,15 @@ extern "C" int mgp_init_module(struct mgp_module *module, struct mgp_memory *mem
...
     mgp::memory = memory;
 
+    auto empty_list = mgp::Value(mgp::List{});
+    auto empty_map = mgp::Map{};
+    empty_map.Insert("key", empty_list);
+
+    AddProcedure(Path::Create, Path::kProcedureCreate, mgp::ProcedureType::Read,
+                 {mgp::Parameter(Path::kCreateArg1, mgp::Type::Node),
+                  mgp::Parameter(Path::kCreateArg2, {mgp::Type::Map, mgp::Type::List}, mgp::Value(empty_map))},
+                 {mgp::Return(Path::kResultCreate, mgp::Type::Path)}, module, memory);
```
antepusic, line 16: I suggest changing the second argument/`relationships` to a list type:
* in the algorithm, you only work with `relationships["rel"]`
* simpler calls, e.g. `path.create(club, {rel: [inLeague, inCountry, inConfederation]})`→ `path.create(club, [inLeague, inCountry, inConfederation])`
* consistent with APOC

cpp/text_module/algorithm/text.cpp
```diff
@@ -0,0 +1,34 @@
...
+      return;
+    }
+
+    auto iterator = list.begin();
+    result += (*iterator).ValueString();
+
+    for (++iterator; iterator != list.end(); ++iterator) {
+      result += delimiter;
+      result += (*iterator).ValueString();
+    }
```
antepusic, line 25: Because you're already working with `std::string`, I recommend using `string::append` ([reference](https://cplusplus.com/reference/string/string/append/)).

antepusic (changes requested):

cpp/text_module/algorithm/text.cpp
imilinovic: This actually isn't a part of this PR but resulted from merging with main, my bad.

I've actually checked and in most implementations += is defined as inline append call so should be the same.

cpp/path_module/path_module.cpp
imilinovic: The problem is that null value can't be sent in a list ex. [relationship, relationship, null] but it can be sent this way. This is important because the main usage of this function is with the optional clause

antepusic: I see, thanks!

antepusic (approved):

PR #326, opened 2023-08-23
[main < T622] Add convert str2object

mpintaric55334:
### Description

### Pull request type

- [ ] Bugfix
- [x] Algorithm/Module
- [ ] Feature
- [ ] Build related changes

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation


mpintaric55334:
NOTE: E2e tests on this will fail until Ivan's PR for file loading https://github.com/memgraph/mage/pull/322 is merged into main and tests use it, because I load files in the way which his PR enables

python/convert.py
```diff
@@ -0,0 +1,7 @@
+import mgp
+from json import loads
+
+@mgp.function
+def str2object(string: str) -> mgp.Any:
+    if string:
+        return loads(string)
```
antoniofilipovic, line 7: What do you return otherwise? Do we need to throw error?

antoniofilipovic (commented):
I am not sure what should happen if value is not string

python/convert.py
mpintaric55334: No no, this is not a check if it is a string, that is automatic since in function definition we have str, this checks if the input string is empty, because if it is json loads gets crazy

e2e/convert_test/test_str2object1/test.yml
```diff
@@ -0,0 +1,8 @@
+query: >
+    LOAD CSV FROM "/_file1" WITH header AS row
+    RETURN convert.str2object(row.b) AS list;
+
+output: 
+  - list: [ "1","2", 3]
+  - list: [["1",{"key": 5}]]
+  - list: {"prop": {"prop2": {"key": 10}}}
```
antepusic, line 8: Format stuff:
```suggestion
  - list: ["1", "2", 3]
  - list: [["1", {"key": 5}]]
  - list: {"prop": {"prop2": {"key": 10}}}
```

antepusic (approved):

antepusic (changes requested):
I’d approved this PR before, but #322 has been merged and checks are still failing – just a heads-up for us not to merge this

python/convert.py
```diff
@@ -0,0 +1,10 @@
+import mgp
+from json import loads
+
+
+@mgp.function
+def str2object(string: str) -> mgp.Any:
+    if string:
+        return loads(string)
+    else:
+        return None
```
antoniofilipovic, line 10: here you can just make it like this:
```suggestion
      return None
```

antoniofilipovic: Since you are returning on early line, you can return here too same way

antoniofilipovic (changes requested):
Few last changes

antepusic (approved):

antoniofilipovic (approved):

PR #327, opened 2023-08-24
[main < T350-MAGE] Implement graphml import and export

ind1xa:
### Description

Implemented import_util.graphml and export_util.graphml

### Pull request type

- [ ] Bugfix
- [x] Algorithm/Module
- [ ] Feature
- [ ] Build related changes

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation


python/export_util.py
```diff
@@ -76,16 +77,82 @@ def convert_to_isoformat(
...
+        if graphML:
+            properties = {
+                key: convert_to_isoformat_graphML(vertex.properties.get(key))
+                for key in vertex.properties.keys()
+            }
+        else:
+            properties = {
+                key: convert_to_isoformat(vertex.properties.get(key))
+                for key in vertex.properties.keys()
+            }
```
antoniofilipovic, line 155:
```suggestion
        if graphML:
            properties = {
                key: convert_to_isoformat_graphML(vertex.properties.get(key))
                for key in vertex.properties.keys()
            }
            continue
       
          properties = {
                key: convert_to_isoformat(vertex.properties.get(key))
                for key in vertex.properties.keys()
            }
```

python/export_util.py
```diff
@@ -138,7 +211,7 @@ def json_stream(ctx: mgp.ProcCtx) -> mgp.Record(stream=str):
     """
     Procedure to export the whole database to a stream.
     """
-    return mgp.Record(stream=js.dumps(get_graph(ctx)))
+    return mgp.Record(stream=js.dumps(get_graph(ctx, 0)))
```
antoniofilipovic, line 214: Pass false again instead of 0

python/export_util.py
```diff
@@ -235,3 +311,287 @@ def csv_query(
         data = csv_to_stream(data_list)
 
     return mgp.Record(file_path=file_path, data=data)
+
+
+def write_header(output: io.StringIO):
```
antoniofilipovic, line 316: This is only for graphml so I would call function: write_graphml_header

python/export_util.py
```diff
@@ -235,3 +311,287 @@ def csv_query(
...
+        return "float"
+    if isinstance(variable, int):
+        return "int"
+    raise Exception(
+        "Property values can only be primitive types or arrays of primitive types."  # noqa: E501
+    )
+
+
+def check_if_elements_same_type(variable: Any):
+    for element in variable:
```
antoniofilipovic, line 338: You need to check here that variable is list and is not empty, otherwise exception will be raised on line 339 when acessing elements

python/export_util.py
```diff
@@ -235,3 +311,287 @@ def csv_query(
...
+        if config.get("useTypes"):
+            if isinstance(value, str):
+                output.write(f' attr.type="{value}"')
+            else:
+                output.write(f' attr.type="{value[0]}" attr.list="{value[1]}"')
+        output.write("/>\n")
+
+
+def write_graph(output: io.StringIO):
+    output.write('<graph id="G" edgedefault="directed">\n')
```
antoniofilipovic, line 411: Do you have idea for more descriptive function, like write_graph_ml?

python/export_util.py
```diff
@@ -235,3 +311,287 @@ def csv_query(
...
+
+    Parameters
+    ----------
+    path : str
+        Path to the graphML file containing the exported graph database.
+    config : Map
+
+    """
+
+    graph = get_graph(ctx, 1)
```
antoniofilipovic, line 558: Again, be more expressive in code: get_graph(ctx, True)

python/export_util.py
```diff
@@ -235,3 +311,287 @@ def csv_query(
...
+    set_default_config(config)
+
+    if config.get("leaveOutLabels") or config.get("leaveOutProperties"):
+        for element in graph:
+            if config.get("leaveOutLabels") and element.get("type") == "node":
+                element.update({"labels": []})
+            if config.get("leaveOutProperties"):
+                element.update({"properties": {}})
+
```
antoniofilipovic, line 567: For me it makes sense to pass these config settings to get graph function as part of some config. So for example, get_graph can accept ctx and config dict {graphML:false, leaveOutLabels:false, leaveOutProps:false} can be default and inside that function we do additional processing. I think it is bit faster because geting of properties is expensive (O(n) for each node) but in case of setting leaveOutProperties we are doing that in any case

python/export_util.py
```diff
@@ -235,3 +311,287 @@ def csv_query(
...
+                    output.write(f":{label}")
+            output.write('">')
+
+            write_labels_as_data(element, output, config)
+
+            for key, value in element.get("properties").items():
+                output.write(
+                    f'<data key="{key}">{get_value_string(value)}</data>'
+                )
+            output.write("</node>\n")
```
antoniofilipovic, line 484: I think this can be part of process_graph_ml_node_object

python/export_util.py
```diff
@@ -235,3 +311,287 @@ def csv_query(
...
+                    f'<data key="label">{element.get("label")}</data>'
+                )
+            if config.get("format").upper() == "GEPHI":
+                output.write(f'<data key="TYPE">{element.get("label")}</data>')
+
+            for key, value in element.get("properties").items():
+                output.write(
+                    f'<data key="{key}">{get_value_string(value)}</data>'
+                )
+            output.write("</edge>\n")
```
antoniofilipovic, line 506: And this can be part of process_graph_ml_relationship_object

python/import_util.py
```diff
@@ -122,13 +128,244 @@ def json(ctx: mgp.ProcCtx, path: str) -> mgp.Record():
...
+            raise KeyError(
+                "The provided file does not match the correct JSON format."
+            )
 
     return mgp.Record()
+
+
+def find_node(
+    ctx: mgp.ProcCtx, label: str, prop_key: str, prop_value: str
+) -> int:  # TODO prop value cast??
```
antoniofilipovic, line 153: this returns union[int, None], right?

python/export_util.py
```diff
@@ -235,3 +311,287 @@ def csv_query(
...
+    rel_keys = dict()
+    for element in graph:
+        if element.get("type") == "node":
+            if element.get("labels"):
+                if config.get("format").upper() == "TINKERPOP":
+                    node_keys.update({"labelV": get_type_string("labelV")})
+                else:
+                    node_keys.update(
+                        {"labels": get_type_string("labels")}
+                    )  # what if two nodes have same property name but different value type? or node's property name is "labels"?  # noqa: E501
```
antoniofilipovic, line 374: What does this mean? What will happen here?

python/export_util.py
```diff
@@ -235,3 +311,287 @@ def csv_query(
...
+
+
+def write_labels_as_data(
+    element: Union[Node, Relationship], output: io.StringIO, config: mgp.Map
+):
+    if not element.get("labels"):
+        return
+    if config.get("format").upper() == "TINKERPOP":
+        output.write('<data key="labelV">')
+        for i in range(0, len(element.get("labels"))):
```
antoniofilipovic, line 434: Check here how to write this with enumerate: https://realpython.com/python-enumerate/

python/export_util.py
```diff
@@ -235,3 +311,287 @@ def csv_query(
...
+        output.write("</data>")
+    if config.get("format").upper() == "GEPHI":
+        output.write('<data key="TYPE">')
+        for label in element.get("labels"):
+            output.write(f":{label}")
+        output.write("</data>")
+        output.write(
+            f'<data key="label">{get_gephi_label_value(element, config)}</data>'  # noqa: E501
+        )
+    else:
```
antoniofilipovic, line 448: You don't need else here, you can do an early return

antoniofilipovic (changes requested):
Good work, left few comments to do

python/import_util.py
```diff
@@ -122,13 +128,244 @@ def json(ctx: mgp.ProcCtx, path: str) -> mgp.Record():
...
+            "Config parameter must be a map with specific \
+             keys and values described in documentation."
+        )
+
+
+@mgp.write_proc
+def graphml(
+    ctx: mgp.ProcCtx,
+    path: str = "",
+    config: mgp.Map = {},
```
antepusic, line 233:
```suggestion
    config: mgp.Map = dict(),
```
Using an empty dictionary as a default parameter may cause the module to act unpredictably because that variable will persist until the next call ([source](https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python)).

python/import_util.py
```diff
@@ -122,13 +128,244 @@ def json(ctx: mgp.ProcCtx, path: str) -> mgp.Record():
...
+
+    node_keys = dict()
+    rel_keys = dict()
+
+    for key in root.findall(".//graphml:key", namespace):
+        # key value legend: value[0] = attr.name
+        #                   value[1] = attr.type
+        #                   value[2] = attr.list
+        #                   value[3] = default
+        # -> should I make a class out of it?
```
antepusic, line 265: I think a `StrEnum` would be handy: https://docs.python.org/3/library/enum.html#enum.StrEnum

python/export_util.py
```diff
@@ -76,16 +77,82 @@ def convert_to_isoformat(
         return property
 
 
-def get_graph(ctx: mgp.ProcCtx) -> List[Union[Node, Relationship]]:
+def to_duration_iso_format(value: timedelta) -> str:
```
antepusic, line 80: To be consistent with Python’s `datetime.isoformat` method:
```suggestion
def to_duration_isoformat(value: timedelta) -> str:
```

antepusic, line 80: I think the part with calculating hours/minutes/etc. can be made neater with `divmod` ([example](https://github.com/jayaddison/timedelta-isoformat/blob/main/src/timedelta_isoformat/__init__.py#L163-L179)).

python/export_util.py
```diff
@@ -235,3 +311,287 @@ def csv_query(
...
+    if isinstance(variable, int):
+        return "int"
+    raise Exception(
+        "Property values can only be primitive types or arrays of primitive types."  # noqa: E501
+    )
+
+
+def check_if_elements_same_type(variable: Any):
+    for element in variable:
+        if not isinstance(element, type(variable[0])):
```
antepusic, line 339: Extract `type(variable[0])` ahead of the for loop to avoid calling it more than once.

python/export_util.py
```diff
@@ -235,3 +311,287 @@ def csv_query(
...
+            "Config parameter must be a map with specific \
+             keys and values described in documentation."
+        )
+
+
+@mgp.read_proc
+def graphml(
+    ctx: mgp.ProcCtx,
+    path: str = "",
+    config: mgp.Map = {},
```
antepusic, line 545: See above comment about default dict parameters
```suggestion
    config: mgp.Map = dict(),
```

antepusic (commented):
Left a few comments, I'll take a more detailed look later too :)

python/export_util.py
ind1xa: but wont continue go to the next vertex in ctx.graph.vertices and lines 157-173 will be skipped?

python/export_util.py
ind1xa: before calling this function I check if I call it for a tuple, i'm gonna change the typings

python/import_util.py
ind1xa: why strEnum? how would i use that? i can only think of using int enum for clearer access to array indexs

python/export_util.py
ind1xa: it will break things 🙃 because key id should be unique for every combination of key name and key type but since we are using key name as key id, it is not possible to have properties that have the same name and different value type because one will overwrite the other. I left it like that since neo4j did it like that also, I think they decided that it is a good tradeoff because it simplifies the process (by not using key ids) and rarely we have same property name for different type in same database?

python/export_util.py
antoniofilipovic: Hmm, but I think that is part of this function, because someone else can forget that check and still call this function, so I would push that check here

python/import_util.py
```diff
@@ -122,13 +131,253 @@ def json(ctx: mgp.ProcCtx, path: str) -> mgp.Record():
...
```
antoniofilipovic, line 273: I think you can make class of this

python/export_util.py
antoniofilipovic: yes but isn't that okay?

python/import_util.py
antepusic: Should’ve been clearer, my idea was to make a StrEnum for `"attr.type"` etc. and an IntEnum for the 0...3 indices.

antoniofilipovic (approved):
Looks good to me

antepusic (approved):

PR #335, opened 2023-09-01
[E-idora < T368, T369] Implement refactor.clone_subgraph and subgraph_from_path

ind1xa:
### Description

Implemented refactor.clone_subgraph and refactor.clone_subgraph_from_path procedures.

### Pull request type

- [ ] Bugfix
- [x] Algorithm/Module
- [ ] Feature
- [ ] Build related changes

### Related issues

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation


cpp/refactor_module/algorithm/refactor.cpp
```diff
@@ -0,0 +1,165 @@
...
+    const auto rels = arguments[1].ValueList();
+    const auto config_map = arguments[2].ValueMap();
+
+    std::unordered_set<mgp::Node> distinct_nodes;
+    std::unordered_set<mgp::Relationship> distinct_rels;
+
+    for (auto node : nodes) {
+      if (node.IsNode()) {
+        distinct_nodes.insert(node.ValueNode());
+      }
```
antoniofilipovic, line 138: These are 100% nodes so no need for if here. In `CallCustomProcedure` function in memgraph you can check there we are checking if types are correct

cpp/refactor_module/algorithm/refactor.cpp
```diff
@@ -0,0 +1,165 @@
...
+    }
+    for (auto rel : rels) {
+      if (rel.IsRelationship()) {
+        distinct_rels.insert(rel.ValueRelationship());
+      }
+    }
+
+    if (distinct_rels.size() == 0 && distinct_nodes.size() > 1) {
```
antoniofilipovic, line 146: why bigger than 1 and not `>=1` ?

cpp/refactor_module/algorithm/refactor.cpp
```diff
@@ -0,0 +1,165 @@
...
+      return true;
+    }
+  }
+  return false;
+}
+
+void Refactor::CloneNodesAndRels(mgp_graph *memgraph_graph, mgp_result *result, mgp_memory *memory,
+                                 const std::vector<mgp::Node> &nodes, const std::vector<mgp::Relationship> &rels,
+                                 const mgp::Map &config_map) {
+  mgp::List standin_nodes;
```
antoniofilipovic, line 48: What does `standin_nodes` mean?

cpp/refactor_module/algorithm/refactor.cpp
```diff
@@ -0,0 +1,165 @@
...
+  mgp::List skip_props;
+  if ((!config_map.At("standinNodes").IsList() && !config_map.At("standinNodes").IsNull()) ||
+      (!config_map.At("skipProperties").IsList() && !config_map.At("skipProperties").IsNull())) {
+    throw mgp::ValueException("Configuration map must consist of specific keys and values described in documentation.");
+  }
```
antoniofilipovic, line 53: I think you need to check that before this function. So correct steps would be:

```
void Refactor::CloneSubgraph{
  isolate nodes and rels
  
  do check on config maps
  isolate nodes and other stuff from config maps
  
  Refactor::CloneNodesAndRels(standinNodes, ....)
}
```

cpp/refactor_module/algorithm/refactor.cpp
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    for (auto prop : node.Properties()) {
+      if (skip_props.Empty() || !skip_props_searchable.contains(mgp::Value(prop.first))) {
+        new_node.SetProperty(prop.first, prop.second);
+      }
+    }
+    old_new_node_mirror.insert({node, new_node});
+    InsertCloneNodesRecord(memgraph_graph, result, memory, node.Id().AsInt(), new_node.Id().AsInt());
+  }
```
antoniofilipovic, line 82: I think you can also put this in different function, which is just dealing with cloning nodes.

cpp/refactor_module/algorithm/refactor.cpp
```diff
@@ -0,0 +1,165 @@
...
+  for (auto rel : rels) {
+    mgp::Relationship new_relationship =
+        graph.CreateRelationship(GetStandinOrCopy(standin_nodes, rel.From(), old_new_node_mirror),
+                                 GetStandinOrCopy(standin_nodes, rel.To(), old_new_node_mirror), rel.Type());
+    for (auto prop : rel.Properties()) {
+      if (skip_props.Empty() || !skip_props_searchable.contains(mgp::Value(prop.first))) {
+        new_relationship.SetProperty(prop.first, prop.second);
+      }
+    }
+  }
```
antoniofilipovic, line 93: Same for this part

cpp/refactor_module/algorithm/refactor.cpp
```diff
@@ -0,0 +1,165 @@
...
+    if (!pair.IsList() || !pair.ValueList()[0].IsNode() || !pair.ValueList()[1].IsNode()) {
+      throw mgp::ValueException(
+          "Configuration map must consist of specific keys and values described in documentation.");
+    }
+    if (node == pair.ValueList()[0].ValueNode()) {
+      return pair.ValueList()[1].ValueNode();
+    }
+  }
+  try {
+    return old_new_node_mirror.at(node);  // what if they send me wrong path
```
antoniofilipovic, line 26: What does this mean, if they send the wrong path?

cpp/refactor_module/algorithm/refactor.cpp
```diff
@@ -0,0 +1,165 @@
...
+  } catch (const std::out_of_range &e) {
+    throw mgp::ValueException("Can't clone relationship without cloning relationship's source and/or target nodes.");
+  }
+}
+
+bool CheckIfStandin(mgp::Node node, mgp::List standin_nodes) {
```
antoniofilipovic, line 32: You can pass these params here as const ref. Here we don't need to copy them, although it is a cheap copy. So correct params passing would be:
`bool CheckIfStandin(const mgp::Node &node, const mgp::List &standin_nodes)`

cpp/refactor_module/algorithm/refactor.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+void Refactor::InsertCloneNodesRecord(mgp_graph *graph, mgp_result *result, mgp_memory *memory, const int cycle_id,
+                                      const int node_id) {
+  auto *record = mgp::result_new_record(result);
+
+  mg_utility::InsertIntValueResult(record, std::string(kResultClonedNodeId).c_str(), cycle_id, memory);
+  mg_utility::InsertNodeValueResult(graph, record, std::string(kResultNewNode).c_str(), node_id, memory);
+}
+
+mgp::Node GetStandinOrCopy(mgp::List &standin_nodes, mgp::Node node,
```
antoniofilipovic, line 14: I think you can pass this node also as const &

e2e_correctness/refactor_test/test_clone_subgraph_1/input.cyp
```diff
@@ -0,0 +1 @@
+MERGE (ana:Ana{name:'Ana', id:0}) MERGE (marija:Marija{name:'Marija', id: 1}) MERGE (p2:Person{name:'person2', id:2}) MERGE (p3:Person{name:'person3', id:3}) MERGE (p4:Person{name:'person4', id:4}) MERGE (p5:Person{name:'person5', id:5}) MERGE (p6:Person{name:'person6', id:6}) CREATE (ana)-[:KNOWS]->(p2)-[:KNOWS]->(p3)-[:KNOWS]->(p4) CREATE (p4)<-[:KNOWS]-(p5) CREATE (p5)-[:LOVES]->(p6) CREATE (marija)-[:KNOWS]->(p6);
```
antoniofilipovic, line 1: Add empty line here

antoniofilipovic: Also add them in rest of the file where lines are missing

antoniofilipovic (changes requested):
Good work, I see few improvements we can do

cpp/refactor_module/algorithm/refactor.cpp
ind1xa: just a note for myself to surround it with a try-catch block, which i did but forgot to delete the comment, sorry

cpp/refactor_module/algorithm/refactor.cpp
ind1xa: I did it this way because I need to do it for both CloneSubgraph and CloneSubgraphFromPaths so this way I only have that piece of code once.. My idea of CloneNodesAndRels was to unite everything that both those procedures need. Does that make sense?

cpp/refactor_module/algorithm/refactor.cpp
ind1xa: sorry, working with config maps got me used to always checking what gets passed😂

cpp/refactor_module/algorithm/refactor.cpp
ind1xa: I did it but I usually don't like doing that because I am just passing 1000 arguments around and it is rarely reusable since all the arguments are specific to that one procedure.. Is there maybe some other way to make it more readable (I'm guessing that is why you ask to do it), like surrounding it with a comment block or?

antoniofilipovic (approved):
Looks good to me

PR #355, opened 2023-09-14
[main < T647]  Add correctness testing for path output

mpintaric55334:
### Description

Added an option in correctness testing, to see if output of path results is equal to one another. The order of the paths in the result is not important, because the paths will be sorted.

 
### Pull request type

- [x] Feature

### Related issues

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation


mpintaric55334:
Would like to discuss it on daily, since I took some liberties in implementing it we have not exactly agreed on.

e2e_correctness/query_neo_mem.py
```diff
@@ -282,3 +298,62 @@ def neo4j_get_graph(neo4j_driver: neo4j.BoltDriver) -> Graph:
...
+    paths.sort()
+    return paths
+
+
+def path_to_string_mem(path):
+    path_string = "NODES:"
+
+    for node in path._nodes:
+        path_string += "(" + str(node._properties.get("id")) + ")-"
+    path_string = path_string[: len(path_string) - 1]
```
imilinovic, line 343: I would shorten it to:
`path_string += `-`.join([f"({node._properties.get("id")})" for node in nodes])

e2e_correctness/query_neo_mem.py
```diff
@@ -282,3 +298,62 @@ def neo4j_get_graph(neo4j_driver: neo4j.BoltDriver) -> Graph:
...
+
+    path_string += " RELATIONSHIPS: "
+    for relationship in path._relationships:
+        path_string += "[" + str(relationship._properties.get("id")) + "]-"
+    path_string = path_string[: len(path_string) - 1]
```
imilinovic, line 348: Same here

e2e_correctness/query_neo_mem.py
imilinovic: Also just for knowledge you can use:
`list[:-1]` instead of
`list[: len(path_string) - 1]`

e2e_correctness/query_neo_mem.py
```diff
@@ -282,3 +298,62 @@ def neo4j_get_graph(neo4j_driver: neo4j.BoltDriver) -> Graph:
...
+
+    return path_string
+
+
+def parse_mem(results):
+    paths = []
+    for result in results:
+        paths.append(path_to_string_mem(result["result"]))
```
imilinovic, line 356: `paths = [path_to_string_mem(result["result"]) for result in results]`

e2e_correctness/test_modules.py
```diff
@@ -174,11 +187,49 @@ def _run_test(test_dir: str, memgraph_db: Memgraph, neo4j_driver: neo4j.BoltDriv
...
+    """
+    Run input queries on Memgraph and Neo4j and compare path results after running test query
+    """
+    input_cyphers = (
+        test_dir.joinpath(TestConstants.INPUT_FILE).open("r").readlines()
+    )
+    mg_execute_cyphers(input_cyphers, memgraph_db)
+    logger.info(f"Imported data into Memgraph from {input_cyphers}")
+    neo4j_execute_cyphers(input_cyphers, neo4j_driver)
+    logger.info(f"Imported data into Neo4j from {input_cyphers}")
```
imilinovic, line 202: You can ignore this comment if it doesn't make sense but I would print something along the line of:
logger: Importing from cypher: input_cyphers
logger: Imported data into memgraph
logger: Imported data into neo4j
Since you are printing the same possibly huge data twice

imilinovic (changes requested):
Few stylistic changes otherwise looks good!

e2e_correctness/query_neo_mem.py
imilinovic: Actually you can probably shorten that whole function to one fstring

mpintaric55334: Changed the way of string formation, so cant do it like this

e2e_correctness/query_neo_mem.py
```diff
@@ -282,3 +286,121 @@ def neo4j_get_graph(neo4j_driver: neo4j.BoltDriver) -> Graph:
...
+            )
+            + "]-"
+        )
+
+    return path_string
+
+
+def parse_neo4j(results):
+    paths = []
+    paths = [path_to_string_neo4j(res) for res in results]
```
imilinovic, line 354: duplicate

e2e_correctness/query_neo_mem.py
```diff
@@ -282,3 +286,121 @@ def neo4j_get_graph(neo4j_driver: neo4j.BoltDriver) -> Graph:
...
+                + str(relationship.get("id"))
+                + " type: "
+                + relationship.type
+                + " "
+                + rel_props
+            )
+            + "]-"
+        )
+
+    return path_string
```
imilinovic, line 349: Python += -> very bad
I think it would be much better to create path_string_list -> append to it and return ''.join(path_string_list) at the end
Furthermore:
string = string + string + string + string generally should be replaced with an fstring or .format unless its only 2 strings, so (str + str).

e2e_correctness/query_neo_mem.py
```diff
@@ -282,3 +286,121 @@ def neo4j_get_graph(neo4j_driver: neo4j.BoltDriver) -> Graph:
...
+                + str(relationship._properties.get("id"))
+                + " type: "
+                + relationship._type
+                + " "
+                + rel_props
+            )
+            + "]-"
+        )
+
+    return path_string
```
imilinovic, line 400: same comment as previous function

imilinovic (changes requested):
Small changes!

e2e_correctness/query_neo_mem.py
```diff
@@ -282,3 +286,121 @@ def neo4j_get_graph(neo4j_driver: neo4j.BoltDriver) -> Graph:
...
+                + str(node_labels)
+                + " "
+                + str(node_props)
+            )
+            + ")-"
+        )
+
+        if i == n - 1:
+            path_string = path_string[:-1]
+            continue
```
imilinovic, line 332: You can just return path_string[:-1] at the end

e2e_correctness/test_modules.py
```diff
@@ -174,11 +190,56 @@ def _run_test(test_dir: str, memgraph_db: Memgraph, neo4j_driver: neo4j.BoltDriv
...
+    )
+    neo4j_results = execute_query_neo4j(
+        neo4j_driver, test_dict[TestConstants.NEO4J_QUERY]
+    )
+    neo4j_paths = parse_neo4j(neo4j_results)
+    logger.info("Done")
+
+    assert memgraph_paths == neo4j_paths
+
+def check_path_option(test_dir):
```
mpintaric55334, line 231: @antoniofilipovic @imilinovic is this what you envisioned, this check if exists config and option

e2e_correctness/query_neo_mem.py
```diff
@@ -297,3 +301,88 @@ def neo4j_get_graph(neo4j_driver: neo4j.BoltDriver) -> Graph:
...
+
+    for i in range(0, n):
+        node = path.nodes[i]
+        node_labels = list(node.labels)
+        node_labels.sort()
+        node_props = str(sort_dict(node._properties))
+        path_string_list.append("(id:")
+        path_string_list.append(str(node.get("id")) + " labels: ")
+        path_string_list.append(str(node_labels) + " ")
+        path_string_list.append(str(node_props) + ")-")
```
imilinovic, line 336: `path_string_list.append(f"(id:{str(node.get("id))} "labels: {std(node_labels)} {str(node_props)})-")`
I maybe you don't even need {str(not_string)}, you can just do {not_string} this way but you need to check

e2e_correctness/query_neo_mem.py
```diff
@@ -297,3 +301,88 @@ def neo4j_get_graph(neo4j_driver: neo4j.BoltDriver) -> Graph:
...
+            path_string = "".join(path_string_list)
+            return path_string[:-1]
+
+        relationship = path.relationships[i]
+        rel_props = str(sort_dict(relationship._properties))
+        path_string_list.append("[")
+        path_string_list.append("id:" + str(relationship.get("id")))
+        path_string_list.append(" type: " + relationship.type)
+        path_string_list.append(" " + rel_props)
+        path_string_list.append("]-")
```
imilinovic, line 348: same as previous

e2e_correctness/query_neo_mem.py
```diff
@@ -297,3 +301,88 @@ def neo4j_get_graph(neo4j_driver: neo4j.BoltDriver) -> Graph:
...
+
+    for i in range(0, n):
+        node = path._nodes[i]
+        node_labels = list(node._labels)
+        node_labels.sort()
+        node_props = str(sort_dict(node._properties))
+        path_string_list.append("(id:")
+        path_string_list.append(str(node._properties.get("id")) + " labels: ")
+        path_string_list.append(str(node_labels) + " ")
+        path_string_list.append(str(node_props) + ")-")
```
imilinovic, line 370: same as above

e2e_correctness/query_neo_mem.py
```diff
@@ -297,3 +301,88 @@ def neo4j_get_graph(neo4j_driver: neo4j.BoltDriver) -> Graph:
...
+            path_string = "".join(path_string_list)
+            return path_string[:-1]
+
+        relationship = path._relationships[i]
+        rel_props = str(sort_dict(relationship._properties))
+        path_string_list.append("[")
+        path_string_list.append("id:" + str(relationship._properties.get("id")))
+        path_string_list.append(" type: " + relationship._type)
+        path_string_list.append(" " + rel_props)
+        path_string_list.append("]-")
```
imilinovic, line 382: same as above

imilinovic (changes requested):
One small change and good to go

e2e_correctness/test_modules.py
```diff
@@ -174,11 +190,57 @@ def _run_test(test_dir: str, memgraph_db: Memgraph, neo4j_driver: neo4j.BoltDriv
...
+    config_path = test_dir.joinpath(TestConstants.CONFIG_FILE)
+    if(os.path.exists(config_path)):
+        config_dict = _load_yaml(config_path)
+        if "path_option" in config_dict:
+            option = config_dict["path_option"].strip()
+            return ( option == "True")
```
imilinovic, line 237: Why is "True" used instead of a boolean?

mpintaric55334: To force whoever is writing tests to write True, because if we use python conversion from string to boolean, you can write anything and it will be true. For example, "Bla bla bla bla bla " = True

e2e_correctness/query_neo_mem.py
```diff
@@ -297,3 +301,88 @@ def neo4j_get_graph(neo4j_driver: neo4j.BoltDriver) -> Graph:
...
+
+
+def parse_neo4j(results):
+    paths = [path_to_string_neo4j(res) for res in results]
+    paths.sort()
+    return paths
+
+
+def path_to_string_mem(path):
```
antoniofilipovic, line 357: Add typings here

e2e_correctness/query_neo_mem.py
```diff
@@ -297,3 +301,88 @@ def neo4j_get_graph(neo4j_driver: neo4j.BoltDriver) -> Graph:
...
+        relationship = path.relationships[i]
+        rel_props = str(sort_dict(relationship._properties))
+        path_string_list.append("[")
+        path_string_list.append("id:" + str(relationship.get("id")))
+        path_string_list.append(" type: " + relationship.type)
+        path_string_list.append(" " + rel_props)
+        path_string_list.append("]-")
+
+
+def parse_neo4j(results):
```
antoniofilipovic, line 351: And add typings here for input and output

e2e_correctness/query_neo_mem.py
```diff
@@ -297,3 +301,88 @@ def neo4j_get_graph(neo4j_driver: neo4j.BoltDriver) -> Graph:
...
     return create_graph_neo4j_json(json_data)
+
+
+# additions for path testing
+def sort_dict(dict):
+    keys = list(dict.keys())
+    keys.sort()
+    sorted_dict = {i: dict[i] for i in keys}
+    if "id" in sorted_dict:
+        sorted_dict.pop("id")
```
antoniofilipovic, line 312: This part is not part of sorted_dict, I think you should move that to other function

antoniofilipovic: So, this function should just sort dict, and you return the sorted dict from this function and then remove id

antoniofilipovic (changes requested):
Few changes left and that is it

imilinovic (approved):
Format with black, otherwise looks good to go

e2e_correctness/test_modules.py
```diff
@@ -143,11 +153,15 @@ def _graphs_equal(memgraph_graph: Graph, neo4j_graph: Graph) -> bool:
     return True
 
 
-def _run_test(test_dir: str, memgraph_db: Memgraph, neo4j_driver: neo4j.BoltDriver):
+def _run_test(
+    test_dir: str, memgraph_db: Memgraph, neo4j_driver: neo4j.BoltDriver
+):
```
antoniofilipovic, line 158: If you return nothing, you can put
`-> None`
```suggestion
def _run_test(
    test_dir: str, memgraph_db: Memgraph, neo4j_driver: neo4j.BoltDriver
) -> None:
```

e2e_correctness/test_modules.py
```diff
@@ -174,11 +190,57 @@ def _run_test(test_dir: str, memgraph_db: Memgraph, neo4j_driver: neo4j.BoltDriv
     ), "The graphs are not equal, check the logs for more details"
 
 
+def _run_path_test(
+    test_dir: str, memgraph_db: Memgraph, neo4j_driver: neo4j.BoltDriver
+):
```
antoniofilipovic, line 195:
```suggestion
def _run_path_test(
    test_dir: str, memgraph_db: Memgraph, neo4j_driver: neo4j.BoltDriver
) -> None:
```

same comment here

antoniofilipovic (commented):
Just two more comments, that is it

antoniofilipovic (approved):

PR #361, opened 2023-09-18
[E-graphQL < T-collections] Change collections to abide by graphQL

mpintaric55334:
### Description

### Pull request type

- [ ] Bugfix
- [ ] Algorithm/Module
- [ ] Feature
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes

### Related issues

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation

### DIFF
--- a/cpp/collections_module/algorithm/collections.cpp
+++ b/cpp/collections_module/algorithm/collections.cpp
@@ -5,8 +5,7 @@
 #include <unordered_set>
 #include <vector>
 
-void SetResult(mgp::Result &result, const mgp::Value &value){
-
+void Collections::SetResult(mgp::Result &result, const mgp::Value &value) {
   switch (value.Type()) {
     case mgp::Type::Bool:
       return result.SetValue(value.ValueBool());
@@ -36,7 +35,7 @@ void SetResult(mgp::Result &result, const mgp::Value &value){
       return result.SetValue(value.ValueDuration());
 
     default:
-      throw mgp::ValueException("No Record.Insert for this datatype");
+      throw mgp::ValueException("No Result.SetValue for this datatype");
   }
 }
 
@@ -101,13 +100,13 @@ void Collections::ContainsAll(mgp_list *args, mgp_func_context *ctx, mgp_func_re
     const auto list1{arguments[0].ValueList()};
     std::unordered_set<mgp::Value> set;
 
-    for (const auto &elem: list1){
+    for (const auto &elem : list1) {
       set.insert(elem);
     }
     const auto list2{arguments[1].ValueList()};
     std::unordered_set<mgp::Value> values;
 
-    for (const auto &elem: list2){
+    for (const auto &elem : list2) {
       values.insert(elem);
     }
     result.SetValue(std::all_of(values.begin(), values.end(), [&](const auto &x) { return set.contains(x); }));
@@ -126,14 +125,14 @@ void Collections::Intersection(mgp_list *args, mgp_func_context *ctx, mgp_func_r
   try {
     const auto list1{arguments[0].ValueList()};
     std::unordered_set<mgp::Value> set1;
-    for (const auto &elem: list1){
+    for (const auto &elem : list1) {
       set1.insert(elem);
     }
 
     const auto list2{arguments[1].ValueList()};
     std::unordered_set<mgp::Value> set2;
 
-    for (const auto &elem: list2){
+    for (const auto &elem : list2) {
       set2.insert(elem);
     }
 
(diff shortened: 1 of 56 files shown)


cpp/collections_module/algorithm/collections.cpp
```diff
@@ -5,10 +5,44 @@
...
+      return result.SetValue(value.ValueDate());
+    case mgp::Type::LocalTime:
+      return result.SetValue(value.ValueLocalTime());
+    case mgp::Type::LocalDateTime:
+      return result.SetValue(value.ValueLocalDateTime());
+    case mgp::Type::Duration:
+      return result.SetValue(value.ValueDuration());
+
+    default:
+      throw mgp::ValueException("No Result.SetValue for this datatype");
```
antoniofilipovic, line 38:
```suggestion
      throw mgp::ValueException(fmt::format("No Result.SetValue for {} datatype", value.Type()));
```

Check how to link `fmt` module, you should just link it and it should work fine afterwards

cpp/collections_module/algorithm/collections.cpp
```diff
@@ -50,59 +83,56 @@ void Collections::Avg(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *res
...
   const auto arguments = mgp::List(args);
-  const auto record_factory = mgp::RecordFactory(result);
+  auto result = mgp::Result(res);
 
   try {
     const auto list1{arguments[0].ValueList()};
     std::unordered_set<mgp::Value> set;
 
-    for (const auto &elem: list1){
+    for (const auto &elem : list1) {
```
antoniofilipovic, line 103: I think with new Memgraph this can actually be changed to work as was intended:

`std::unordered_set<mgp::Value> set(list1.begin(), list2.begin());`

cpp/collections_module/algorithm/collections.cpp
```diff
@@ -50,59 +83,56 @@ void Collections::Avg(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *res
...
       set.insert(elem);
     }
     const auto list2{arguments[1].ValueList()};
     std::unordered_set<mgp::Value> values;
 
-    for (const auto &elem: list2){
+    for (const auto &elem : list2) {
```
antoniofilipovic, line 109: Same for this part of iterations

cpp/collections_module/algorithm/collections.cpp
antoniofilipovic: This should be fixed with Idora's PR on Memgraph with C++ API traits added under Iterator class in List

cpp/collections_module/algorithm/collections.cpp
```diff
@@ -50,59 +83,56 @@ void Collections::Avg(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *res
...
       values.insert(elem);
     }
-    auto record = record_factory.NewRecord();
-    record.Insert(std::string(kResultContainsAll).c_str(),
-                  std::all_of(values.begin(), values.end(), [&](const auto &x) { return set.contains(x); }));
+    result.SetValue(std::all_of(values.begin(), values.end(), [&](const auto &x) { return set.contains(x); }));
```
antoniofilipovic, line 112:
```suggestion
    result.SetValue(std::all_of(values.begin(), values.end(), [&set](const auto &x) { return set.contains(x); }));
```
It is suggested that you don't use default capture (reference capture), or at least that we use reference capture by variables we need.

So I would recommend in this case changing it like this, in other cases you can check in memgraph that we are using init capture: https://stackoverflow.com/questions/25408190/lambda-captures-c14

cpp/collections_module/algorithm/collections.cpp
```diff
@@ -50,59 +83,56 @@ void Collections::Avg(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *res
...
   mgp::memory = memory;
   const auto arguments = mgp::List(args);
-  const auto record_factory = mgp::RecordFactory(result);
+  auto result = mgp::Result(res);
 
   try {
     const auto list1{arguments[0].ValueList()};
     std::unordered_set<mgp::Value> set1;
-    for (const auto &elem: list1){
+    for (const auto &elem : list1) {
```
antoniofilipovic, line 128: Same comment about unordered set

antoniofilipovic (changes requested):
Few changes to do, but looks good

cpp/collections_module/algorithm/collections.cpp
mpintaric55334: Have mercy with the fix provided, the fmt wont link

imilinovic (approved):
Good job! Looks good!

cpp/collections_module/algorithm/collections.cpp
antoniofilipovic: Fix is okay, I am not sure why fmt won't link, will check

antoniofilipovic (approved):
Seems good to go

PR #364, opened 2023-09-19
[E-refactor < T371] Add refactor.extract_node

imilinovic:
### Description

Add refactor.extract_node

### Pull request type

- [ ] Bugfix
- [X] Algorithm/Module
- [ ] Feature
- [ ] Build related changes

### Related issues

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation


antepusic (changes requested):
If this method is analogous to [`apoc.refactor.extractNode`](https://neo4j.com/labs/apoc/4.1/overview/apoc.refactor/apoc.refactor.extractNode/), it should also take the relationship properties and add them to the newly created node.

imilinovic:
E2E tests were used since Neo4j procedure has bugs.

antepusic (approved):

cpp/refactor_module/algorithm/refactor.cpp
```diff
@@ -460,9 +469,73 @@ void Refactor::CollapseNode(mgp_list *args, mgp_graph *memgraph_graph, mgp_resul
...
+    }
+
+  } catch (const std::exception &e) {
+    record_factory.SetErrorMessage(e.what());
+    return;
+  }
+}
+
+void Refactor::ExtractNode(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result, mgp_memory *memory) {
+  mgp::memory = memory;
```
antoniofilipovic, line 481: Let's start using DispatchGuard here

cpp/refactor_module/algorithm/refactor.cpp
```diff
@@ -460,9 +469,73 @@ void Refactor::CollapseNode(mgp_list *args, mgp_graph *memgraph_graph, mgp_resul
...
+  const auto arguments = mgp::List(args);
+  const auto record_factory = mgp::RecordFactory(result);
+  try {
+    mgp::Graph graph{memgraph_graph};
+    mgp::Value rel_or_id{arguments[0]};
+    auto labels{arguments[1].ValueList()};
+    auto out_type{arguments[2].ValueString()};
+    auto in_type{arguments[3].ValueString()};
+
+    auto extract = [&](mgp::Relationship relationship) {
```
antoniofilipovic, line 491: I think you only need here &graph, and &labels. It is better to catch by reference only what you need since this function may catch other stuff here too

cpp/refactor_module/algorithm/refactor.cpp
```diff
@@ -460,9 +469,73 @@ void Refactor::CollapseNode(mgp_list *args, mgp_graph *memgraph_graph, mgp_resul
...
+      auto new_node = graph.CreateNode();
+      for (const auto &label : labels) {
```
antoniofilipovic, line 493: I think you don't need here reference, since labels is returning string_view and it is not copied actually

antoniofilipovic (changes requested):
Few changes, otherwise looks good

cpp/refactor_module/algorithm/refactor.cpp
imilinovic: labels is a list which contains mgp::Value, not string_view

antoniofilipovic: Ah, sorry, I got something wrong from the code, thought that we got labels from `node.Labels()`. All good!

cpp/refactor_module/algorithm/refactor.cpp
```diff
@@ -460,9 +469,73 @@ void Refactor::CollapseNode(mgp_list *args, mgp_graph *memgraph_graph, mgp_resul
...
+      }
+    }
+
+    if (ids.empty()) {
+      return;
+    }
+
+    for (auto relationship : graph.Relationships()) {
+      if (ids.contains(relationship.Id().AsInt())) {
+        extract(relationship);
```
antoniofilipovic, line 536: One tiny detail, I think this auto relationship can be `const auto` and you extract can accept `const auto &`. Here we will nonetheless in iterations over relationships create Relationship objects so it is okay that it is not & in my opinion as it is temporary, but extract can accept `const &` as from it's perspective that object exists

cpp/refactor_module/algorithm/refactor.cpp
```diff
@@ -460,9 +469,73 @@ void Refactor::CollapseNode(mgp_list *args, mgp_graph *memgraph_graph, mgp_resul
...
+  const auto arguments = mgp::List(args);
+  const auto record_factory = mgp::RecordFactory(result);
+  try {
+    mgp::Graph graph{memgraph_graph};
+    mgp::Value rel_or_id{arguments[0]};
+    auto labels{arguments[1].ValueList()};
+    auto out_type{arguments[2].ValueString()};
+    auto in_type{arguments[3].ValueString()};
+
+    auto extract = [&graph, &labels, &record_factory, out_type, in_type](mgp::Relationship relationship) {
```
antoniofilipovic, line 491: This can accept const & in my opinion

antoniofilipovic (changes requested):
Sorry for taking too long, but one small change and that is it

antoniofilipovic (approved):

PR #367, opened 2023-09-19
[E-refactor < T374] Add refactor.normalize_as_boolean

imilinovic:
### Description

Add refactor.normalize_as_boolean

### Pull request type

- [ ] Bugfix
- [X] Algorithm/Module
- [ ] Feature
- [ ] Build related changes

### Related issues

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation


antepusic (changes requested):
Everything looks fine, but please test the case of overlap between `true_values` and `false_values`.

imilinovic:
@antepusic Wanted to discuss how to handle the overlap on the daily tomorrow so I left the PR in draft.
Sorry, should've tagged after converting to real PR.

imilinovic:
It was decided to extend the procedure to lists as well as it should be a bit better complexity wise (vs calling it for each list element separately). Also if a property is contained in both true_values and false_values an exception will be thrown.

cpp/refactor_module/algorithm/refactor.cpp
```diff
@@ -496,6 +496,8 @@ void NormalizeToBoolean(node_or_rel object, std::string &property_key, std::unor
     object.SetProperty(property_key, mgp::Value(false));
   } else if (!in_true && !in_false) {
     object.RemoveProperty(property_key);
+  } else {
+    throw mgp::ValueException("Property is contained in both true_values and false_values.");
```
antepusic, line 500: I think this is a more informative error message:
```suggestion
    throw mgp::ValueException(fmt::format("The value {} is contained in both true_values and false_values.", value.ToString()));
```

antepusic (commented):
One comment left from my side

cpp/refactor_module/algorithm/refactor.cpp
imilinovic: Done without fmt as it felt unnecessary to add fmt to CMake for one use.

cpp/refactor_module/algorithm/refactor.cpp
```diff
@@ -1,9 +1,18 @@
...
+
+#include <mg_utils.hpp>
 #include "mgp.hpp"
 
+namespace {
+void ThrowInvalidTypeException(const mgp::Value &value) {
+  std::ostringstream oss;
+  oss << value.Type();
+  throw mgp::ValueException("Unsupported type for this operation, received type: " + oss.str());
+}
```
antoniofilipovic, line 13: Can we isolate this in some function, this is a repeating pattern. And did you try to link it with `fmt`?

cpp/refactor_module/algorithm/refactor.cpp
```diff
@@ -460,9 +469,80 @@ void Refactor::CollapseNode(mgp_list *args, mgp_graph *memgraph_graph, mgp_resul
...
+  } catch (const std::exception &e) {
+    record_factory.SetErrorMessage(e.what());
+    return;
+  }
+}
+
+namespace {
+
+template <typename node_or_rel>
+void NormalizeToBoolean(node_or_rel object, std::string &property_key, std::unordered_set<mgp::Value> &true_values,
```
antoniofilipovic, line 483: Also, I think this can be passed as &, because here we don't need to make a copy as this object exists somewhere else.

cpp/refactor_module/algorithm/refactor.cpp
```diff
@@ -460,9 +469,80 @@ void Refactor::CollapseNode(mgp_list *args, mgp_graph *memgraph_graph, mgp_resul
...
+                        std::unordered_set<mgp::Value> &false_values) {
+  auto old_value = object.GetProperty(property_key);
+  if (old_value.IsNull()) {
+    return;
+  }
+
+  bool in_true = true_values.contains(old_value);
```
antoniofilipovic, line 490: Name in_true, as `property_in_true_vals`, so we know what is that reference to

cpp/refactor_module/algorithm/refactor.cpp
```diff
@@ -460,9 +469,80 @@ void Refactor::CollapseNode(mgp_list *args, mgp_graph *memgraph_graph, mgp_resul
...
+  bool in_false = false_values.contains(old_value);
```
antoniofilipovic, line 491: Similarly here, it is okay to be more descriptive

cpp/refactor_module/algorithm/refactor.cpp
```diff
@@ -460,9 +469,80 @@ void Refactor::CollapseNode(mgp_list *args, mgp_graph *memgraph_graph, mgp_resul
...
+
+  if (in_true && !in_false) {
+    object.SetProperty(property_key, mgp::Value(true));
+  } else if (!in_true && in_false) {
+    object.SetProperty(property_key, mgp::Value(false));
+  } else if (!in_true && !in_false) {
+    object.RemoveProperty(property_key);
+  } else {
+    throw mgp::ValueException("The value {" + old_value.ToString() +
+                              "} is contained in both true_values and false_values.");
```
antoniofilipovic, line 501: Try formatting this with `fmt::format`

cpp/refactor_module/algorithm/refactor.cpp
```diff
@@ -460,9 +469,80 @@ void Refactor::CollapseNode(mgp_list *args, mgp_graph *memgraph_graph, mgp_resul
...
+    auto convert_to_set = [](const mgp::List &list) {
+      std::unordered_set<mgp::Value> set;
+      for (const auto &list_item : list) {
+        set.insert(list_item);
+      }
+      return set;
+    };
+
+    std::unordered_set<mgp::Value> true_values{convert_to_set(true_values_list)};
+    std::unordered_set<mgp::Value> false_values{convert_to_set(false_values_list)};
```
antoniofilipovic, line 526: can't you do this in the following way:
```
std::unordered_set<mgp::Value> true_values{true_values_list.begin(), true_values_list.end() };
```

antoniofilipovic (changes requested):
Good work, there are few changes I suggest to do

cpp/refactor_module/algorithm/refactor.cpp
imilinovic: Isn't the C++ API the only place where we could add this so its usable from all modules?

cpp/refactor_module/algorithm/refactor.cpp
imilinovic: In function parse object.ValueNode() is sent which can't be sent with reference since it creates a new object :(

imilinovic: And this function is always called with either object.ValueNode() or object.ValueRelationship()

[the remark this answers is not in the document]
imilinovic: String also can be const, so added that as well

cpp/refactor_module/algorithm/refactor.cpp
imilinovic: Actually this can be a one liner since we can use value.ToString() so we don't need this function

antoniofilipovic (approved):
Looks good to me

cpp/refactor_module/algorithm/refactor.cpp
antoniofilipovic: yes, true

PR #370, opened 2023-09-26
[E-path < T667-T668] Path subgraph improvement and path refactor

imilinovic:
### Description

Refactor the path module and change path subgraph to bfs instead of dfs which should improve the speed and lower memory usage by a lot

### Pull request type

- [ ] Bugfix
- [x] Algorithm/Module
- [ ] Feature
- [x] Code style update (formatting, renaming)
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes

#### Module/Algorithm
- [x] Unit tests
- [x] End-to-end tests
- [ ] Code documentation

### DIFF
--- a/cpp/path_module/algorithm/path.cpp
+++ b/cpp/path_module/algorithm/path.cpp
@@ -201,7 +201,7 @@ void Path::PathHelper::ParseRelationships(const mgp::List &list_of_relationships
 }
 
 void Path::Create(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result, mgp_memory *memory) {
-  mgp::memory = memory;
+  mgp::MemoryDispatcherGuard guard{memory};
   const auto arguments = mgp::List(args);
   const auto record_factory = mgp::RecordFactory(result);
   try {
@@ -254,68 +254,75 @@ void Path::Create(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result,
   }
 }
 
-/*function used for traversal and filtering*/
-void Path::PathDFS(mgp::Path &path, const mgp::RecordFactory &record_factory, int64_t path_size,
-                   PathHelper &path_helper, std::unordered_set<int64_t> &visited) {
-  const mgp::Node node{path.GetNodeAt(path.Length())};
+void Path::PathExpand::ExpandPath(mgp::Path &path, const mgp::Relationship &relationship, int64_t path_size) {
+  path.Expand(relationship);
+  visited_.insert(relationship.Id().AsInt());
+  DFS(path, path_size + 1);
+  visited_.erase(relationship.Id().AsInt());
+  path.Pop();
+}
 
-  LabelBools label_bools = path_helper.GetLabelBools(node);
-  if (path_helper.PathSizeOk(path_size) && path_helper.AreLabelsValid(label_bools)) {
-    auto record = record_factory.NewRecord();
+void Path::PathExpand::Expand(mgp::Path &path, mgp::Relationships relationships, bool outgoing, int64_t path_size,
+                              std::set<std::pair<std::string_view, int64_t>> &seen) {
+  for (const auto relationship : relationships) {
+    auto type = std::string(relationship.Type());
+    auto wanted_direction = helper_.GetDirection(type);
+
+    if ((wanted_direction == RelDirection::kNone && !helper_.AnyDirected(outgoing)) ||
+        visited_.contains(relationship.Id().AsInt())) {
+      continue;
+    }
+
+    RelDirection curr_direction = outgoing ? RelDirection::kOutgoing : RelDirection::kIncoming;
+
+    if (wanted_direction == RelDirection::kAny || curr_direction == wanted_direction || helper_.AnyDirected(outgoing)) {
+      ExpandPath(path, relationship, path_size);
+    } else if (wanted_direction == RelDirection::kBoth) {
+      if (outgoing && seen.contains({type, relationship.To().Id().AsInt()})) {
+        ExpandPath(path, relationship, path_size);
+      } else {
+        seen.insert({type, relationship.From().Id().AsInt()});
+      }
+    }
+  }
+}
+
+/*function used for traversal and filtering*/
+void Path::PathExpand::DFS(mgp::Path &path, int64_t path_size) {
+  const mgp::Node node{path.GetNodeAt(path_size)};
+
+  LabelBools label_bools = helper_.GetLabelBools(node);
+  if (helper_.PathSizeOk(path_size) && helper_.AreLabelsValid(label_bools)) {
+    auto record = record_factory_.NewRecord();
     record.Insert(std::string(kResultExpand).c_str(), path);
   }
 
-  if (!path_helper.ContinueExpanding(label_bools, path_size + 1)) {
+  if (!helper_.ContinueExpanding(label_bools, path_size + 1)) {
     return;
   }
 
   std::set<std::pair<std::string_view, int64_t>> seen;
-
-  auto expand = [&](mgp::Relationships relationships, bool outgoing) {
-    for (const auto relationship : relationships) {
-      auto type = std::string(relationship.Type());
-      auto wanted_direction = path_helper.GetDirection(type);
-
-      if ((wanted_direction == RelDirection::kNone && !path_helper.AnyDirected(outgoing)) ||
-          visited.contains(relationship.Id().AsInt())) {
-        continue;
-      }
-
-      RelDirection curr_direction = outgoing ? RelDirection::kOutgoing : RelDirection::kIncoming;
-
-      auto expand_path = [&]() {
-        path.Expand(relationship);
-        visited.insert(relationship.Id().AsInt());
-        PathDFS(path, record_factory, path_size + 1, path_helper, visited);
-        visited.erase(relationship.Id().AsInt());
-        path.Pop();
-      };
-
-      if (wanted_direction == RelDirection::kAny || curr_direction == wanted_direction ||
-          path_helper.AnyDirected(outgoing)) {
-        expand_path();
-      } else if (wanted_direction == RelDirection::kBoth) {
-        if (outgoing && seen.contains({type, relationship.To().Id().AsInt()})) {
-          expand_path();
-        } else {
-          seen.insert({type, relationship.From().Id().AsInt()});
-        }
-      }
-    }
-  };
-
-  expand(node.InRelationships(), false);
-  expand(node.OutRelationships(), true);
+  this->Expand(path, node.InRelationships(), false, path_size, seen);
+  this->Expand(path, node.OutRelationships(), true, path_size, seen);
 }
 
-void Path::StartFunction(const mgp::Node &node, const mgp::RecordFactory &record_factory, PathHelper &path_helper) {
+void Path::PathExpand::StartAlgorithm(const mgp::Node node) {
   mgp::Path path = mgp::Path(node);
-  std::unordered_set<int64_t> visited;
-  PathDFS(path, record_factory, 0, path_helper, visited);
+  DFS(path, 0);
+}
+
+void Path::PathExpand::Parse(const mgp::Value &value) {
+  if (value.IsNode()) {
+    StartAlgorithm(value.ValueNode());
+  } else if (value.IsInt()) {
+    StartAlgorithm(graph_.GetNodeById(mgp::Id::FromInt(value.ValueInt())));
+  } else {
+    throw mgp::ValueException("Invalid start type. Expected Node, Int, List[Node, Int]");
+  }
 }
 
 void Path::Expand(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result, mgp_memory *memory) {
-  mgp::memory = memory;
+  mgp::MemoryDispatcherGuard guard{memory};
   const auto arguments = mgp::List(args);
   const auto record_factory = mgp::RecordFactory(result);
   try {
@@ -326,25 +333,16 @@ void Path::Expand(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result,
     int64_t min_hops{arguments[3].ValueInt()};
     int64_t max_hops{arguments[4].ValueInt()};
 
-    PathHelper path_helper = PathHelper(labels, relationships, min_hops, max_hops);
-
-    auto parse = [&](const mgp::Value &value) {
-      if (value.IsNode()) {
-        StartFunction(value.ValueNode(), record_factory, path_helper);
-      } else if (value.IsInt()) {
-        StartFunction(graph.GetNodeById(mgp::Id::FromInt(value.ValueInt())), record_factory, path_helper);
-      } else {
-        throw mgp::ValueException("Invalid start type. Expected Node, Int, List[Node, Int]");
-      }
-    };
+    PathHelper path_helper{labels, relationships, min_hops, max_hops};
+    PathExpand path_expand{path_helper, record_factory, graph};
 
     if (!start_value.IsList()) {
-      parse(start_value);
+      path_expand.Parse(start_value);
       return;
     }
 
     for (const auto &list_item : start_value.ValueList()) {
-      parse(list_item);
+      path_expand.Parse(list_item);
     }
 
   } catch (const std::exception &e) {
@@ -353,43 +351,66 @@ void Path::Expand(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result,
   }
 }
 
-void GetStartNodes(const mgp::Value element, const mgp::Graph &graph, std::unordered_set<mgp::Node> &start_nodes) {
-  if (!(element.IsNode() || element.IsInt())) {
+void Path::PathSubgraph::Parse(const mgp::Value &value) {
+  if (!(value.IsNode() || value.IsInt())) {
     throw mgp::ValueException("The first argument needs to be a node, an integer ID, or a list thereof.");
   }
-  if (element.IsNode()) {
-    start_nodes.insert(element.ValueNode());
+  if (value.IsNode()) {
+    start_nodes_.insert(value.ValueNode());
     return;
   }
-  start_nodes.insert(graph.GetNodeById(mgp::Id::FromInt(element.ValueInt())));
+  start_nodes_.insert(graph_.GetNodeById(mgp::Id::FromInt(value.ValueInt())));
 }
 
-void Path::RunBFS(std::unordered_set<mgp::Node> &start_nodes, mgp::List &to_be_returned_nodes,
-                  Path::PathHelper &path_helper) {
-  struct Pair {
-    mgp::Node node;
-    int64_t hop_count;
-  };
+void Path::PathSubgraph::Expand(Pair &pair, mgp::Relationships relationships, bool outgoing, std::queue<Pair> &queue,
+                                std::set<std::pair<std::string_view, int64_t>> &seen) {
+  for (const auto relationship : relationships) {
+    auto next_node = outgoing ? relationship.To() : relationship.From();
+    auto type = std::string(relationship.Type());
+    auto wanted_direction = helper_.GetDirection(type);
 
+    if (helper_.FilterRelationships(pair.hop_count == 0)) {
+      if ((wanted_direction == RelDirection::kNone && !helper_.AnyDirected(outgoing)) ||
+          visited_.contains(next_node.Id().AsInt())) {
+        continue;
+      }
+    }
+
+    RelDirection curr_direction = outgoing ? RelDirection::kOutgoing : RelDirection::kIncoming;
+
+    if (wanted_direction == RelDirection::kAny || curr_direction == wanted_direction || helper_.AnyDirected(outgoing)) {
+      visited_.insert(next_node.Id().AsInt());
+      queue.push({next_node, pair.hop_count + 1});
+    } else if (wanted_direction == RelDirection::kBoth) {
+      if (outgoing && seen.contains({type, relationship.To().Id().AsInt()})) {
+        visited_.insert(next_node.Id().AsInt());
+        queue.push({next_node, pair.hop_count + 1});
+        to_be_returned_nodes_.AppendExtend(mgp::Value{next_node});
+      } else {
+        seen.insert({type, relationship.From().Id().AsInt()});
+      }
+    }
+  }
+}
+
+void Path::PathSubgraph::InsertNode(const mgp::Node &node, int64_t hop_count, LabelBools &label_bools) {
+  if (helper_.FilterNodes(hop_count == 0)) {
+    if (helper_.AreLabelsValid(label_bools)) {
+      to_be_returned_nodes_.AppendExtend(mgp::Value(node));
+    }
+    return;
+  }
+
+  if (!visited_.contains(node.Id().AsInt())) {
+    to_be_returned_nodes_.AppendExtend(mgp::Value(node));
+  }
+}
+
+mgp::List Path::PathSubgraph::BFS() {
   std::queue<Pair> queue;
   std::unordered_set<int64_t> visited;
 
-  auto insert_node = [&path_helper, &to_be_returned_nodes, &visited](const mgp::Node &node, int64_t hop_count,
-                                                                     LabelBools &label_bools) {
-    if (path_helper.FilterNodes(hop_count == 0)) {
-      if (path_helper.AreLabelsValid(label_bools)) {
-        std::cout << "ubacujem " << node.GetProperty("name").ValueString() << " on this\n";
-        to_be_returned_nodes.AppendExtend(mgp::Value(node));
-      }
-      return;
-    }
-
-    if (!visited.contains(node.Id().AsInt())) {
-      to_be_returned_nodes.AppendExtend(mgp::Value(node));
-    }
-  };
-
-  for (const auto &node : start_nodes) {
+  for (const auto &node : start_nodes_) {
     queue.push({node, 0});
     visited.insert(node.Id().AsInt());
   }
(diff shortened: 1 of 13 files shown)


imilinovic:
Needs this [Memgraph PR1249](https://github.com/memgraph/memgraph/pull/1249)

cpp/path_module/algorithm/path.hpp
```diff
@@ -56,52 +57,102 @@ struct LabelBoolsStatus {
   bool termination_activated = false;
 };
 
-struct RelationshipSets {
-  std::unordered_set<std::string> outgoing_rel;
-  std::unordered_set<std::string> incoming_rel;
-  std::unordered_set<std::string> any_rel;
+enum RelDirection { kNone = -1, kAny = 0, kIncoming = 1, kOutgoing = 2, kBoth = 3 };
```
antoniofilipovic, line 60: Prefer scoped enums to unscoped enums, it is general practice in C++. I would suggest here to change `enum` to `enum class` and use that.

cpp/path_module/algorithm/path.cpp
```diff
@@ -1,301 +1,343 @@
...
+
+void Path::PathExpand::ExpandPath(mgp::Path &path, const mgp::Relationship &relationship, int64_t path_size) {
+  path.Expand(relationship);
+  visited_.insert(relationship.Id().AsInt());
+  DFS(path, path_size + 1);
+  visited_.erase(relationship.Id().AsInt());
+  path.Pop();
+}
+
+void Path::PathExpand::Expand(mgp::Path &path, mgp::Relationships relationships, bool outgoing, int64_t path_size,
```
antoniofilipovic, line 260: Can this part have different naming than mgp::Path::Exand, it is difficult to track. Maybe  `StartExpanding`

cpp/path_module/algorithm/path.cpp
```diff
@@ -1,301 +1,343 @@
...
+
+void Path::PathExpand::Parse(const mgp::Value &value) {
+  if (value.IsNode()) {
+    StartAlgorithm(value.ValueNode());
+  } else if (value.IsInt()) {
+    StartAlgorithm(graph_.GetNodeById(mgp::Id::FromInt(value.ValueInt())));
+  } else {
+    throw mgp::ValueException("Invalid start type. Expected Node, Int, List[Node, Int]");
+  }
 }
```
antoniofilipovic, line 317: I think this Parse is not just parsing but also doing expansions. It make sense that this just returns node, and then we startAlgorithm in Path::Exapnd

cpp/path_module/algorithm/path.cpp
antoniofilipovic: And this also has same name as Path::Exapnd, so we have three different expands, it is quite a lot of similar names.  mgp::Path::Expand stays as is, Path::Expand can be called Path::ExpandProc, and Path::PathExpand::Expand can be call Path::PathExpand::StartExpanding

cpp/path_module/algorithm/path.cpp
```diff
@@ -1,301 +1,343 @@
...
+    RelDirection curr_direction = outgoing ? RelDirection::kOutgoing : RelDirection::kIncoming;
+
+    if (wanted_direction == RelDirection::kAny || curr_direction == wanted_direction || helper_.AnyDirected(outgoing)) {
+      ExpandPath(path, relationship, path_size);
+    } else if (wanted_direction == RelDirection::kBoth) {
+      if (outgoing && seen.contains({type, relationship.To().Id().AsInt()})) {
+        ExpandPath(path, relationship, path_size);
+      } else {
+        seen.insert({type, relationship.From().Id().AsInt()});
+      }
```
antoniofilipovic, line 280: Can you explain logic a bit here too

cpp/path_module/algorithm/path.cpp
```diff
@@ -1,301 +1,343 @@
...
 
-  if (labelSets.whitelist.find(label) != labelSets.whitelist.end()) {  // if label is whitelisted
-    labelBools.whitelisted = true;
+  if (config_.label_sets.whitelist.find(label) != config_.label_sets.whitelist.end()) {  // if label is whitelisted
+    label_bools.whitelisted = true;
   }
 }
 
 /*function that takes input list of labels, and sorts them into appropriate category
 sets were used so when filtering is done, its done in O(1)*/
```
antoniofilipovic, line 131: I think function comments are done as follows:
```
// Function that takes input list of labels, and sorts them into appropriate category
// sets were used so when filtering is done, its done in O(1)
```

cpp/path_module/algorithm/path.cpp
```diff
@@ -1,301 +1,343 @@
...
+      } else if (ends_with) {
+        config_.any_outgoing = true;
+      } else {
+        config_.relationship_sets[rel_type] = RelDirection::kAny;
+      }
+      continue;
+    }
+
+    if (starts_with && ends_with) {
+      config_.relationship_sets[rel_type.substr(1, rel_type.size() - 2)] = RelDirection::kBoth;
```
antoniofilipovic, line 187: Why is here -2?

antoniofilipovic: Maybe add comment how this should look

cpp/path_module/algorithm/path.cpp
```diff
@@ -1,301 +1,343 @@
...
+          (label_bools.end_node || Whitelisted(label_bools.whitelisted)));
 }
-bool Path::ShouldExpand(const LabelBools &labelBools, const LabelBoolsStatus &labelStatus) {
-  return !labelBools.blacklisted && ((labelBools.end_node && labelStatus.end_node_activated) || labelBools.terminated ||
-                                     (!labelStatus.termination_activated && !labelStatus.end_node_activated &&
-                                      Whitelisted(labelBools.whitelisted, labelStatus.whitelist_empty)));
+
+bool Path::PathHelper::PathSizeOk(const int64_t path_size) const {
+  return (path_size <= config_.max_hops) && (path_size >= config_.min_hops);
 }
```
antoniofilipovic, line 90: Why this part checks max and min hops and in ContinueExpandin we just check max hops?

cpp/path_module/algorithm/path.cpp
```diff
@@ -1,301 +1,343 @@
...
+  if (!(config.At("minHops").IsNull() || config.At("minHops").IsInt()) ||
+      !(config.At("maxHops").IsNull() || config.At("maxHops").IsInt()) ||
+      !(config.At("relationshipFilter").IsNull() || config.At("relationshipFilter").IsList()) ||
+      !(config.At("labelFilter").IsNull() || config.At("labelFilter").IsList()) ||
+      !(config.At("filterStartNode").IsNull() || config.At("filterStartNode").IsBool()) ||
+      !(config.At("beginSequenceAtStart").IsNull() || config.At("beginSequenceAtStart").IsBool()) ||
+      !(config.At("bfs").IsNull() || config.At("bfs").IsBool())) {
+    throw mgp::ValueException(
+        "The config parameter needs to be a map with keys and values in line with the documentation.");
+  }
```
antoniofilipovic, line 24: Is it possible to make function which can accept expected types and check for each one that is one of the expected types and throws if it is not

```
if (!config.At("minHops").IsExpectedType({Type::Null, Type::Int})){
    throw mgp::ValueException...
}
```

antoniofilipovic (changes requested):
Good work overall! I plan to run this algorithm to see how fast it is. Do you have some numbers here for how fast is expansion and SubgraphNodes?

antoniofilipovic:
Don't forget to update README on MAGE too with this algorithm, we forgot to do that, update for this one and I will create PR to update for rest.

Also ping @vpavicic to be in loop once we have release note

antoniofilipovic:
@imilinovic which version of Memgraph this needs?

imilinovic:
It needs this [PR1249](https://github.com/memgraph/memgraph/pull/1249) which is already in main but supposed to be released in Memgraph 2.12

cpp/path_module/algorithm/path.cpp
```diff
@@ -454,34 +465,28 @@ void Path::SubgraphNodes(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *
...
-    for (const auto node : start_nodes) {
-      VisitNode(node, visited_nodes, true, config, 0, labelFilterSets, to_be_returned_nodes);
-    }
+    auto to_be_returned_nodes = path_subgraph.BFS();
 
     std::unordered_set<mgp::Node> to_be_returned_nodes_searchable;
-    for (auto node : to_be_returned_nodes) {
+    for (const auto &node : to_be_returned_nodes) {
       to_be_returned_nodes_searchable.insert(node.ValueNode());
     }
```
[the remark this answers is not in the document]
imilinovic: Not possible because node is of type mgp::Value, and we need to turn it into mgp::Node with ValueNode(). I can do this which makes me feel smart but I don't know if it's better:

`std::transform(
        to_be_returned_nodes.begin(), to_be_returned_nodes.end(),
        std::inserter(to_be_returned_nodes_searchable, to_be_returned_nodes_searchable.begin()),
        [](const mgp::Value &node) { return node.ValueNode(); }
    );`

cpp/path_module/algorithm/path.cpp
```diff
@@ -1,301 +1,343 @@
...
+  // there is a termination node, so only paths ending with it are allowed
+  config_.label_bools_status.termination_activated = !config_.label_sets.termination_list.empty();
 }
+
 /*function to set appropriate parameters for filtering*/
-void Path::FilterLabel(const std::string_view label, const LabelSets &labelSets, LabelBools &labelBools) {
-  if (labelSets.blacklist.find(label) != labelSets.blacklist.end()) {  // if label is blacklisted
-    labelBools.blacklisted = true;
+void Path::PathHelper::FilterLabel(std::string_view label, LabelBools &label_bools) {
+  if (config_.label_sets.blacklist.find(label) != config_.label_sets.blacklist.end()) {  // if label is blacklisted
```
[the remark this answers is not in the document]
imilinovic: I've changed to contains in most parts of the code but somehow left this unchanged

cpp/path_module/algorithm/path.cpp
```diff
@@ -304,115 +346,90 @@ void Path::Expand(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result,
...
 
-bool IsLabelListed(const mgp::Node node, std::unordered_set<std::string_view> &set) {
-  for (const auto label : node.Labels()) {
-    if (set.contains(label)) {
-      return true;
+    RelDirection curr_direction = outgoing ? RelDirection::kOutgoing : RelDirection::kIncoming;
+
+    if (wanted_direction == RelDirection::kAny || curr_direction == wanted_direction || helper_.AnyDirected(outgoing)) {
+      visited_.insert(next_node.Id().AsInt());
+      queue.push({next_node, pair.hop_count + 1});
```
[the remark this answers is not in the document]
imilinovic: This is an optimisation that can sometimes help a lot. If a node is unvisited you know that the current path you're on is the shortest path since you are running BFS so it's always the correct choice. If you don't mark it now you'll just add a lot of nodes in the queue that will be skipped over later because when you get to them the first path that expanded to them will already visit them and render them useless.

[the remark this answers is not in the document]
imilinovic: I can elaborate further on daily if needed.

cpp/path_module/algorithm/path.cpp
```diff
@@ -304,115 +346,90 @@ void Path::Expand(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result,
...
+    } else if (wanted_direction == RelDirection::kBoth) {
+      if (outgoing && seen.contains({type, relationship.To().Id().AsInt()})) {
+        visited_.insert(next_node.Id().AsInt());
+        queue.push({next_node, pair.hop_count + 1});
+        to_be_returned_nodes_.AppendExtend(mgp::Value{next_node});
+      } else {
+        seen.insert({type, relationship.From().Id().AsInt()});
+      }
     }
```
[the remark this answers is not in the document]
imilinovic: If the required relationship filter is <rel> that means you need to have A-[rel]->B and B-[rel]->A.
First we iterate through all incoming relationships and mark them {type, id_of_node_they_come_from}
Then when we iterate through the outgoing relationships we check if we have seen a relationship that goes from the outgoing (to) node with the same type to the current one. 

In simpler terms we check if we have seen this relationship type with the opposite direction between the same two nodes.

cpp/path_module/algorithm/path.cpp
imilinovic: Same logic as the previous practically same piece of code.

cpp/path_module/algorithm/path.hpp
```diff
@@ -56,52 +57,102 @@ struct LabelBoolsStatus {
   bool termination_activated = false;
 };
```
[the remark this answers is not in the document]
imilinovic: Added a comment for each member variable of LabelBoolsStatus and LabelBool explaining what it is.

cpp/path_module/algorithm/path.hpp
```diff
@@ -56,52 +57,102 @@ struct LabelBoolsStatus {
...
+  explicit PathHelper(const mgp::List &labels, const mgp::List &relationships, int64_t min_hops, int64_t max_hops);
+  explicit PathHelper(const mgp::Map &config);
 
-void FilterLabelBoolStatus(const LabelSets &labelSets, LabelBoolsStatus &labelStatus);
+  RelDirection GetDirection(std::string &rel_type);
+  LabelBools GetLabelBools(const mgp::Node &node);
 
-bool ShouldExpand(const LabelBools &labelBools, const LabelBoolsStatus &labelStatus);
+  bool AnyDirected(bool outgoing) const { return outgoing ? config_.any_outgoing : config_.any_incoming; }
+  bool FilterNodes(bool is_start) const { return (config_.filter_start_node || !is_start); }
```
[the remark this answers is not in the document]
imilinovic: FilterNode -> `IsNotStartOrSupportsStartNode` 
FilterRelationship -> `IsNotStartOrSupportsStartRel`

cpp/path_module/algorithm/path.hpp
```diff
@@ -56,52 +57,102 @@ struct LabelBoolsStatus {
...
+
+ protected:
+  PathHelper helper_;
+  const mgp::RecordFactory &record_factory_;
+  const mgp::Graph &graph_;
+  std::unordered_set<int64_t> visited_;
+};
 
-bool Whitelisted(const bool &whitelisted, const bool &whitelist_empty);
+class PathExpand : PathData {
```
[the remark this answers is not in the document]
imilinovic: I did it since PathSubgraph and PathExpand share a lot of variables so they be just called with variable instead of class.variable, but I guess that's not a good practice

cpp/path_module/algorithm/path.hpp
```diff
@@ -56,52 +57,102 @@ struct LabelBoolsStatus {
...
+  void StartAlgorithm(mgp::Node node);
+  void Parse(const mgp::Value &value);
+  void DFS(mgp::Path &path, int64_t path_size);
+};
 
-void PathDFS(mgp::Path path, std::unordered_set<mgp::Relationship> &relationships_set,
-             const mgp::RecordFactory &record_factory, int64_t path_size, const int64_t min_hops,
-             const int64_t max_hops, const LabelSets &labelFilters, const LabelBoolsStatus &labelStatus,
-             const RelationshipSets &relationshipSets, bool &any_outgoing, bool &any_incoming);
+class PathSubgraph : PathData {
```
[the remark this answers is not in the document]
imilinovic: I did it since PathSubgraph and PathExpand share a lot of variables so they be just called with variable instead of class.variable

cpp/path_module/algorithm/path.cpp
```diff
@@ -304,115 +346,90 @@ void Path::Expand(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result,
...
-  if (!(config.At("minLevel").IsInt() && config.At("maxLevel").IsInt() && config.At("relationshipFilter").IsList() &&
-        config.At("labelFilter").IsList() && config.At("filterStartNode").IsBool()) ||
-      (config.At("minLevel").ValueInt() != 0 && config.At("minLevel").ValueInt() != 1)) {
-    throw mgp::ValueException(
-        "The config parameter needs to be a map with keys and values in line with the documentation.");
+
+  while (!queue.empty()) {
+    auto pair = queue.front();
+    queue.pop();
+
```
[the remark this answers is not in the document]
imilinovic: Copy paste previous comment:
This is an optimisation that can sometimes help a lot. If a node is unvisited you know that the current path you're on is the shortest path since you are running BFS so it's always the correct choice. If you don't mark it now you'll just add a lot of nodes in the queue that will be skipped over later because when you get to them the first path that expanded to them will already visit them and render them useless.

So thats why we want to add it before expanding to the node.

cpp/path_module/algorithm/path.hpp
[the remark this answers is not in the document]
imilinovic: I think its the simplest to just add PathData variables to both classes since otherwise it would be something like path_data_.getHelper().something(). And it's only 4 shared variables.

cpp/path_module/algorithm/path.hpp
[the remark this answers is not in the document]
imilinovic: Same comment as previous

cpp/path_module/algorithm/path.cpp
imilinovic: Path::PathExpand::Expand -> ExpandFromRelationships
I think other can stay the same since Path::Expand is called that because the procedure is called path.expand

cpp/path_module/algorithm/path.cpp
imilinovic: Added comments
<rel_type> -> size is original size - 2, remove the starting < and > at the end

cpp/path_module/algorithm/path.cpp
imilinovic: This will actually be useful for ExpandConfig!

cpp/path_module/algorithm/path.cpp
imilinovic: For a path to be appended in results it has to be of size  minhops <= size <= maxhops but if it doesn't satisfy minhops yet we want to still continue expanding

cpp/path_module/algorithm/path.cpp
imilinovic: Would this be ok?

`auto same_type_or_null = [] (const mgp::Type type, const mgp::Type wanted_type) {
    return type == wanted_type || type == mgp::Type::Null;
  };

  if (!same_type_or_null(config.At("minHops").Type(), mgp::Type::Int) ||
     !same_type_or_null(config.At("maxHops").Type(), mgp::Type::Int) ||
     !same_type_or_null(config.At("relationshipFilter").Type(), mgp::Type::List) ||
     !same_type_or_null(config.At("labelFilter").Type(), mgp::Type::List) ||
     !same_type_or_null(config.At("filterStartNode").Type(), mgp::Type::Bool) ||
     !same_type_or_null(config.At("beginSequenceAtStart").Type(), mgp::Type::Bool) ||
     !same_type_or_null(config.At("bfs").Type(), mgp::Type::Bool)){
    throw mgp::ValueException(
        "The config parameter needs to be a map with keys and values in line with the documentation.");
     }`

cpp/path_module/algorithm/path.cpp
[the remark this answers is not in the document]
antoniofilipovic: This is good, do that yes :smile:

cpp/path_module/algorithm/path.cpp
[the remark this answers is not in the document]
antoniofilipovic: Ah, okay, I think I get it, good

cpp/path_module/algorithm/path.cpp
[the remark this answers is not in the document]
antoniofilipovic: Aha, okay, works!

cpp/path_module/algorithm/path.cpp
antoniofilipovic: Does it make sense to isolate this part of code @imilinovic then? In function or something

cpp/path_module/algorithm/path.cpp
antoniofilipovic: Makes sense

cpp/path_module/algorithm/path.cpp
antoniofilipovic: Yes, works!

cpp/path_module/algorithm/path.hpp
```diff
@@ -44,64 +45,113 @@ struct LabelSets {
...
+};
 
-void PathDFS(mgp::Path path, std::unordered_set<mgp::Relationship> &relationships_set,
-             const mgp::RecordFactory &record_factory, int64_t path_size, const int64_t min_hops,
-             const int64_t max_hops, const LabelSets &labelFilters, const LabelBoolsStatus &labelStatus,
-             const RelationshipSets &relationshipSets, bool &any_outgoing, bool &any_incoming);
+class PathSubgraph {
+ public:
+  explicit PathSubgraph(PathHelper &&helper, const mgp::RecordFactory &record_factory, const mgp::Graph &graph)
+      : helper_(helper), record_factory_(record_factory), graph_(graph) {}
```
antoniofilipovic, line 131: You need to call on helper std::move, otherwise, it is copy constructed: https://godbolt.org/z/rxxKxnzjs

helper is still lvalue, you need to cast it with std::move to rvalue:https://godbolt.org/z/o4se6orET

cpp/path_module/algorithm/path.hpp
```diff
@@ -44,64 +45,113 @@ struct LabelSets {
...
-                         bool &any_incoming, bool outgoing);
+ private:
+  Config config_;
+};
 
-bool Whitelisted(const bool &whitelisted, const bool &whitelist_empty);
+class PathExpand {
+ public:
+  explicit PathExpand(PathHelper &&helper, const mgp::RecordFactory &record_factory, const mgp::Graph &graph)
+      : helper_(helper), record_factory_(record_factory), graph_(graph) {}
```
antoniofilipovic, line 110: Same comment about helper as below

cpp/path_module/algorithm/path.hpp
```diff
@@ -44,64 +45,113 @@ struct LabelSets {
...
-  bool blacklisted = false;
-  bool terminated = false;
-  bool end_node = false;
-  bool whitelisted = false;
+  bool blacklisted = false;  // no node in the path will be blacklisted
+  bool terminated = false;   // returned paths end with a termination node but don't continue to be expanded further,
+                             // takes precedence over end nodes
+  bool end_node = false;     // returned paths end with an end node but continue to be expanded further
+  bool whitelisted = false;  // all nodes in the path will be whitelisted (except end and termination nodes)
+  // end and termination nodes don't have to respect whitelists and blacklists
```
antoniofilipovic, line 53: Comments are usually written above variable
```suggestion
   // no node in the path will be blacklisted
  bool blacklisted = false;
  // returned paths end with a termination node but don't continue to be expanded further,
  // takes precedence over end nodes
  bool terminated = false;   
   // returned paths end with an end node but continue to be expanded further                         
  bool end_node = false; 
  // all nodes in the path will be whitelisted (except end and termination nodes)
   // end and termination nodes don't have to respect whitelists and blacklists   
  bool whitelisted = false;  
 
```

cpp/path_module/algorithm/path.hpp
```diff
@@ -44,64 +45,113 @@ struct LabelSets {
...
 };
 
 struct LabelBoolsStatus {
-  bool end_node_activated = false;
-  bool whitelist_empty = false;
-  bool termination_activated = false;
+  bool end_node_activated = false;  // true if there is an end node -> only paths ending with it can be saved as result,
+                                    // but they can be expanded further
+  bool whitelist_empty = false;     // true if no whitelist is given -> all nodes are whitelisted
+  bool termination_activated = false;  // true if there is a termination node -> only paths ending with it are allowed
```
antoniofilipovic, line 60: Same comment here about comments

antoniofilipovic (commented):
Few comments left and that is it

cpp/path_module/algorithm/path.hpp
```diff
@@ -44,64 +45,113 @@ struct LabelSets {
...
+  void Parse(const mgp::Value &value);
+  void DFS(mgp::Path &path, int64_t path_size);
+  void RunAlgorithm();
+
+ private:
+  PathHelper helper_;
+  const mgp::RecordFactory &record_factory_;
+  const mgp::Graph &graph_;
+  std::unordered_set<int64_t> visited_;
+  std::unordered_set<mgp::Node> start_nodes_;
```
antoniofilipovic, line 125: I remembered, you can make this a class, and make PathSubgraph and PathExpand friends of this class and then you can reference private variables

cpp/path_module/algorithm/path.cpp
imilinovic: I think not really because this part can also have some other stuff in it, potentionally bfs and dfs could be isolated later for when ExpandConfig will be implemented.

cpp/path_module/algorithm/path.hpp
imilinovic: Yes, great suggestion!

imilinovic:
I don't have the numbers how fast this is, we could also benchmark other path functions, I'll write this down to discuss on daily

antoniofilipovic (approved):
Looks good to me

PR #377, opened 2023-10-06
[main < T] Bugfix for map.merge

mpintaric55334:
### Description

User noticed a bug, that when the second map contains null elements, merge will throw errors. This happened because we checked the existence of keys using At. If we check it like that, if the key exists, but contains null value, it will be as if the key doesnt exist. Instead I made a set of keys, and checked the existence of keys taht way. @antoniofilipovic  maybe it would make sense to add some Keys() function to map Cpp API to automatize the checking for key existence. 
### Pull request type

- [x] Bugfix
- [ ] Algorithm/Module
- [ ] Feature
- [ ] Build related changes

### Related issues

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation


antoniofilipovic:
@mpintaric55334 it makes sense to add key exists function to C++ API, I agree. Currently seems it is not possible to check if key exists but property value is null

mpintaric55334:
This PR https://github.com/memgraph/memgraph/pull/1336 in memgraph needs to be merged for this to work, we've added new functionality (KeyExists) in Cpp API

antoniofilipovic (approved):
Looks good to me

PR #414, opened 2023-11-13
Add dockerfile, requirements and test for no ML build

antoniofilipovic:
### Description

The idea of this PR is to add a Dockerfile for build of MAGE without ML, which is often used when shipping customer images

### Pull request type

- [ ] Bugfix
- [ ] Algorithm/Module
- [ ] Feature
- [ ] Build related changes

### Related issues

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation


antepusic (approved):
Having compared this with the existing Dockerfiles, I think this is fit for approving :+1:

Dockerfile.no_ML
```diff
@@ -0,0 +1,86 @@
...
+    python3-dev     `mage-memgraph` \
+    clang           `mage-memgraph` \
+    git             `mage-memgraph` \
+    unixodbc-dev        `mage-memgraph` \
+    --no-install-recommends \
+    && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+COPY memgraph-${TARGETARCH}.deb .
+
+RUN dpkg -i memgraph-${TARGETARCH}.deb && rm memgraph-${TARGETARCH}.deb
```
MarkoBarisic, line 33: Dockerfile.release has a fix that was introduced in 1.12.1 release

`RUN dpkg -i memgraph-${TARGETARCH}.deb && rm memgraph-${TARGETARCH}.deb && rm /usr/lib/memgraph/query_modules/schema.so  \
&& rm /usr/lib/memgraph/query_modules/example_cpp.so`

Should this fix be applied here as well?

MarkoBarisic:
Do you think tests for no-ML build should be added to `./github/workflows/test.yml` or the tests aren't needed because it can be implied from regular tests?

Dockerfile.no_ML
antoniofilipovic: So, we don't need that here, because these modules cause an issue only for pytorch modules, and here we don't use them.

antoniofilipovic:
> Do you think tests for no-ML build should be added to `./github/workflows/test.yml` or the tests aren't needed because it can be implied from regular tests?

Oh, nice idea. I think we should add tests most likely because that way we are sure no-ML image also works. Good catch

Dockerfile.no_ML
MarkoBarisic: Thanks

MarkoBarisic (approved):
LGTM, let's merge EOD so I can make on demand platform builds available ASAP :muscle:

PR #415, opened 2023-11-14
Modify CreateCsvFile behavior to not include newline

cizl:
### Description

Modifies the behavior of the `CreateCsvFile` procedure to not include a newline at the end of the line.
This behaviour is required by the Lab.

### Pull request type

- [x] Bugfix
- [ ] Algorithm/Module
- [ ] Feature
- [ ] Build related changes

### Related issues

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation

### DIFF
--- a/cpp/csv_utils_module/algorithm/csv_utils.cpp
+++ b/cpp/csv_utils_module/algorithm/csv_utils.cpp
@@ -14,7 +14,7 @@ void CreateCsvFile(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result
 
     std::ofstream fout;
     fout.open(std::string(filepath), is_append ? std::ofstream::app : std::ofstream::out);
-    fout << content << std::endl;
+    fout << content;
     fout.close();
 
     auto record = record_factory.NewRecord();
--- a/cpp/csv_utils_module/algorithm/csv_utils.cpp
+++ b/cpp/csv_utils_module/algorithm/csv_utils.cpp
@@ -14,7 +14,7 @@ void CreateCsvFile(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result
 
     std::ofstream fout;
     fout.open(std::string(filepath), is_append ? std::ofstream::app : std::ofstream::out);
-    fout << content;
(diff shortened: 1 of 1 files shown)


cpp/csv_utils_module/algorithm/csv_utils.cpp
```diff
@@ -14,7 +14,7 @@ void CreateCsvFile(mgp_list *args, mgp_graph *memgraph_graph, mgp_result *result
 
     std::ofstream fout;
     fout.open(std::string(filepath), is_append ? std::ofstream::app : std::ofstream::out);
-    fout << content << std::endl;
+    fout << content;
```
tonilastre, line 17: Do you need to use `std::flush` here? (https://en.cppreference.com/w/cpp/io/manip/flush) or `fout.close()` will flush the buffer?

cizl: Not sure actually :man_shrugging: 
I guess I do since `std::endl` actually calls `std::flush`.

antoniofilipovic: From cpp ref on `ofstream` [link](https://en.cppreference.com/w/cpp/io/basic_ofstream) we can see following:
- `fout`'s destructor will actually call close: https://en.cppreference.com/w/cpp/io/basic_ofstream/close
- close will call implicit flush via the following function: https://en.cppreference.com/w/cpp/io/basic_filebuf/close

So is unnecessary, and close too.

Double-checked here:
https://stackoverflow.com/questions/16268433/what-does-stdofstreamclose-actually-do

I trust cpp ref, but since this is a really sensitive matter, if I failed to understand some case here, this might be really difficult to debug issue, so would leave it as it is.

antoniofilipovic (approved):

cpp/csv_utils_module/algorithm/csv_utils.cpp
cizl: I included `std::flush` just in case then. (as a replacement for endl because that also flushes)

PR #431, opened 2023-12-10
Community detection memory tracking

### LINKED ISSUE #219: Running `community_detection_online.set()` causes Memgraph to crash
I ran `community_detection_online.set()` on a large dataset (around 200k nodes, 1M relationship) on a running Memgraph instance with 24 GB RAM. Memory spiked and caused Memgraph to crash. It would be great if, in that case, Memgraph does not crash, but the transaction is stopped instead.

### LINKED ISSUE #379: [BUG]  Community_detection does not release memory while it's processing, uses > 2x db size
**Memgraph version**
2.10.2

**Environment**
AWS EC2 Ubuntu 22 running memgraph-platform in a docker container

**Describe the bug**
I run the following query on BlueJ to do community detection:

CALL community_detection.get()
YIELD node, community_id
WITH node, community_id
WHERE 'Person' IN labels(node)
SET node.community_id = community_id;

And the DB size on a fresh start uses 13 out of 28 gigs of memory

Once I run the query, it'll execute and allocate memory until all of the 28 Gb avalable is taken, and keeps going -- up until there's no memory left for allocating file / network handles, so ssh and other services become unavailable.

The only solution is to reboot the server as it'll stay in this state for multiple days

**To Reproduce**

Download the latest bluej db snapshop from https://bluej.memgraph.com/dump/bluej-latest.cypherl.gz and run the query on a linux host (& other platforms presumably) with 32 gb of ram. The machine will crash

**Expected behavior**
No crash. Crash bad

Screenshots of start of query & moment the server became unavailable:

imilinovic:
### Description

Community detection memory tracking was not working properly when multiple threads were used so it was fixed.

Add thread memory tracking to community detection.

Requires [PR436](https://github.com/memgraph/mage/pull/436) and [PR1631](https://github.com/memgraph/memgraph/pull/1631) before merging

Note to reviewer (!!!):
Review with suffix ```?w=1``` to ignore whitespaces changes which are automatically made by IDE (and there are a lot of them).
[Link for reviewing
](https://github.com/memgraph/mage/pull/431/files?w=1)

### Pull request type

- [X] Bugfix
- [ ] Algorithm/Module
- [ ] Feature
- [ ] Build related changes

# Release note:
Community detection now respects memory limits and doesn't crash database.

### Related issues
closes #379 
closes #219 
closes https://github.com/memgraph/memgraph/issues/955

#### Module/Algorithm
- [ ] Unit tests
- [ ] End-to-end tests
- [ ] Code documentation


gitbuda:
@imilinovic, quick question, what's the issue related? Can you link it? 👀

cpp/community_detection_module/CMakeLists.txt
```diff
@@ -7,25 +7,11 @@ set(community_detection_module_src
...
-        community_detection_test.cpp
-        algorithm/louvain.cpp)
-
-    add_executable(community_detection_test "${community_detection_test_src}")
-    target_include_directories(community_detection_test PRIVATE ${GRAPPOLO_HEADERS} ${CMAKE_CURRENT_SOURCE_DIR})
-    target_link_libraries(community_detection_test PRIVATE mg_utility basic_cd full_syn_opt coloring util OpenMP::OpenMP_CXX mage_gtest)
-    gtest_add_tests(TARGET community_detection_test)
-endif()
-
-
```
imilinovic, line 28: These tests were ported to e2e tests.
The tests were not usable because now community_detection library uses mgp_graph and the graph testing library doesn't use that.

cpp/community_detection_module/algorithm/louvain.cpp
```diff
@@ -104,7 +105,11 @@ void LoadUndirectedEdges(const mg_graph::GraphView<> &memgraph_graph, GrappoloGr
   for (std::size_t i = 0; i < number_of_vertices; i++) added[i] = 0;
 
     // Build the edgeList from edgeListTmp:
-#pragma omp parallel for
+#pragma omp parallel
+{
+  [[maybe_unused]] const enum mgp_error tracking_error = mgp_track_current_thread_allocations(graph);
```
DavIvek, line 110: Do we need to track memory that isn't dynamically allocated? Maybe @antoniofilipovic could check what should we track in this algorithms?

DavIvek (commented):
Looks good to me. Only a small note. 
As I understand the tests you moved to e2e tests are tests that existed before and were passing before, even though community detection algorithms weren't usable?
It would be good to add some test that will check if memory limit is being respected with introduced changes.

antepusic (commented):
Good job (& thanks for the reviewers’ note)!

I agree with @DavIvek that it would be good to have a test to verify that this query module now respects the memory limits. Also double-check if there remain any (de)allocations that aren’t tracked.

NB: Is it possible to disable SonarCloud warnings for this PR alone? While it’s a useful tool, its recommendations aren’t exactly relevant to this PR:
* repeated code: this PR mainly consists of adding memory tracking calls
* cognitive complexity of functions: not applicable here because all code save for the memory tracking calls is third-party

imilinovic:
I think the test to check that this works can't be done because of the following:

Memory limit is always being respected but the problem was that memory counting was wrong. The only way would be to compare it with the system memory (or see if the system crashes when setting the limit close to available RAM on the machine) which I don't think is possible at least not with the current MAGE testing infrastructure.

antepusic (approved):
Approved, thanks for the clarification about testing!

PR #445, opened 2024-02-23
Add set property module

Josipmrden:
### Description

Add set property module to MAGE

### Pull request type

- [ ] Bugfix
- [X] Algorithm/Module
- [X] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
- [ ] Other (please describe):

######################################

#### Module/Algorithm
- [X] Core algorithm/module implementation
- [X] [Query module](https://memgraph.com/docs/memgraph/reference-guide/query-modules) implementation
- [ ] Unit tests
- [X] End-to-end tests
- [X] Code documentation
- [X] README short description
- [X] Documentation on [memgraph/docs](https://github.com/memgraph/docs)
- [ ] Update GQLALchemy signatures in [query builder](https://github.com/memgraph/gqlalchemy/blob/main/gqlalchemy/graph_algorithms/query_builder.py  ) using [query module signature generator](https://github.com/memgraph/gqlalchemy/blob/main/scripts/query_module_signature_generator.py)

######################################

@kgolubic "Added set_property module inside Memgraph MAGE"
https://github.com/memgraph/documentation/pull/500

PR #446, opened 2024-02-23
Periodic delete

Josipmrden:
### Description

Periodic delete as a workaround for large number of deltas during deletion

### Pull request type

- [ ] Bugfix
- [X] Algorithm/Module
- [ ] Feature
- [ ] Build related changes

#### Module/Algorithm
- [X] Core algorithm/module implementation
- [X] [Query module](https://memgraph.com/docs/memgraph/reference-guide/query-modules) implementation
- [ ] Unit tests
- [X] End-to-end tests
- [X] Code documentation
- [X] README short description
- [X] Documentation on [memgraph/docs](https://github.com/memgraph/docs)

@kgolubic "Added query module for periodic delete of graph entities"
https://github.com/memgraph/documentation/pull/501


antepusic (changes requested):
Good job, the query module logic is sound and everything else is fine as well. :+1: 

I’ve only had a few comments about avoiding duplication, and please add a test where you delete nodes with a specific label (or repurpose `test_periodic_delete_on_just_nodes` for that

cpp/periodic_module/periodic.cpp
```diff
@@ -27,6 +35,50 @@ struct ParamNames {
...
+  const auto *maybe_password = std::getenv(kMgPassword);
+  if (maybe_password) {
+    password = std::move(maybe_password);
+  }
+
+  return mg::Client::Params{.host = std::move(host),
+                            .port = std::move(port),
+                            .username = std::move(username),
+                            .password = std::move(password)};
+}
```
antoniofilipovic, line 80: Same comment as for #445 PR

cpp/periodic_module/periodic.cpp
```diff
@@ -193,36 +246,189 @@ void ValidateBatchSize(const mgp::Value &batch_size_value) {
...
+  ValidateDeletionConfigEntities(config, labels_key);
+  ValidateDeletionConfigEntities(config, edge_types_key);
+}
+
+void EmplaceFromConfig(const mgp::Map &config, std::vector<std::string> &vec, std::string &config_key) {
+  auto key = std::string_view(config_key);
+  if (config.KeyExists(key)) {
+    auto value = config.At(key);
+    if (value.IsString()) {
+      vec.emplace_back(std::string(value.ValueString()));
```
antoniofilipovic, line 301: You don't need std::string, as a `value.ValueString()` returns std::string_view and it will be constructed in place:
https://godbolt.org/z/94Kz5sM1n

cpp/periodic_module/periodic.cpp
```diff
@@ -193,36 +246,189 @@ void ValidateBatchSize(const mgp::Value &batch_size_value) {
...
+
+  EmplaceFromConfig(config, labels, labels_key);
+  EmplaceFromConfig(config, edge_types, edge_types_key);
+
+  return {.batch_size = static_cast<uint64_t>(batch_size),
+          .labels = std::move(labels),
+          .edge_types = std::move(edge_types)};
+}
+
+void ExecutePeriodicDelete(DeletionInfo deletion_info, DeletionResult &deletion_result) {
```
antoniofilipovic, line 330: You can accept here by const &, and then you don't need to move, as this way we need move ctor, and other with const &, it is just variable creation

antoniofilipovic (changes requested):
Few suggestions, otherwise good work :muscle:

cpp/periodic_module/periodic.cpp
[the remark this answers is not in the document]
Josipmrden: I'm not sure what you mean on the first comment

Josipmrden:
@antepusic  @antoniofilipovic  Pls review again!

cpp/periodic_module/periodic.cpp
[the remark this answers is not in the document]
antepusic: About the first comment: `ValidateBatchSize` and `ValidateDeletionConfig` have different checks & error messages for the `batch_size` parameter. I advise to make them consistent, and then to extract that into its own function to avoid repeated code.

antepusic (changes requested):
Made the comment in https://github.com/memgraph/mage/pull/446/files#r1521938923 clearer, just make sure to address it

antepusic (approved):
I think all’s good now :heavy_check_mark:

antoniofilipovic (approved):

PR #486, opened 2024-06-27
Improve rc artifacts workflow

MarkoBarisic:
### Description

This PR improves the initial workflow for creating releasable artifacts for all mage builds, which are the following:
- dev (dev + ML + RelWithDebInfo)
- prod (prod + ML + Release)
- relwithdebinfo (prod + ML + RelWithDebInfo)
- no-ml (prod + no ML + Release)
- no-ml-relwithdebinfo (prod + no ML + RelWithDebInfo)	

### Pull request type

- [ ] Bugfix
- [ ] Algorithm/Module
- [ ] Feature
- [x] Build related changes

### Related issues

#### Module/Algorithm
- [ ] Tests provided (unit / e2e)
- [ ] Code documentation

### Documentation checklist
- [x] Add the documentation label tag
- [x] Add the bug / feature label tag
- [x] Add the milestone for which this feature is intended
    - If not known, set for a later milestone
- [ ] Write a release note, including added/changed clauses
    - **[Release note text]**
- [ ] Link the documentation PR here
    - **[Documentation PR link]**
- [ ] Tag someone from docs team in the comments

### DIFF
--- a/.github/workflows/reusable_package_mage.yaml
+++ b/.github/workflows/reusable_package_mage.yaml
@@ -211,4 +211,4 @@ jobs:
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           AWS_REGION: "eu-west-1"
           SOURCE_DIR: "output"
-          DEST_DIR: "${{ inputs.s3_dest_dir }}/${{ env.ARTIFACT_NAME }}/"
+          DEST_DIR: "${{ inputs.s3_dest_dir }}/"
--- a/.github/workflows/reusable_package_mage.yaml
+++ b/.github/workflows/reusable_package_mage.yaml
@@ -54,6 +54,10 @@ on:
         type: string
         description: "Destination directory in S3 bucket"
         default: "mage-unofficial"
+      large_runner:
+        type: string
+        description: "Use large runner?"
+        default: false
 
 env:
   OS: ${{ inputs.arch == 'arm' && 'debian-11-aarch64' || 'debian-11' }}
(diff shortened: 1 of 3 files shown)


.github/workflows/reusable_package_mage.yaml
```diff
@@ -54,14 +54,18 @@ on:
         type: string
         description: "Destination directory in S3 bucket"
         default: "mage-unofficial"
+      large_runner:
+        type: string
+        description: "Use large runner?"
```
hal-eisen-MG, line 59: Why would the user ever say "no" this this option? Explain the cost/benefit.

MarkoBarisic: We use github hosted runners for this workflow. There are two types of github runners, basic (let's call them) and large. Both are billed on a pay as you go basis. A fixed amount of basic runner minutes are included in memgraph's monthly subscription plan, large runner minutes aren't at all. So the cost here is :heavy_dollar_sign: . The reason for using large runners here is because the dev build takes up too much disk and the basic runners can't handle it.

MarkoBarisic: https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners/about-larger-runners
https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#per-minute-rates

hal-eisen-MG: So the default is false, which means unless the user overrides - it will always pick basic and fail the build. This doesn't seem logical to me.

MarkoBarisic: I understand your concern, let me clear things up.
The user will never be able to manually run the `reusable_package_mage.yaml` workflow.
It is only called by the `release_artifact.yaml` and `package_mage.yaml` workflows. 
I have put in safeguards for this with the following lines:
  - `large_runner: ${{ github.event.inputs.mage_build_target == 'dev' && 'true' || 'false' }}`
  - `large_runner: ${{ matrix.build_target == 'dev' && 'true' || 'false' }}`
Additionally, all builds except `dev` will pass on a basic runner.

hal-eisen-MG (approved):
LGTM

PR #557, opened 2025-03-07
Add migration from Neo4j

Josipmrden:
### Description

Add migration module from Neo4j.
Usage:
```
CALL migrate.neo4j("(:Label)", {host: "localhost", port: 7687}) YIELD row RETURN row.labels, row.properties;

CALL migrate.neo4j("[:TYPE]", {host: "localhost", port: 7687}) YIELD row RETURN row.from_labels, row.to_labels, row.from_properties, row.to_properties, row.edge_properties;

WITH "MATCH (n) RETURN count(n) as cnt" as query
CALL migrate.neo4j(query, {host: "localhost", port: 7687}) YIELD row RETURN row.cnt;
```

### DIFF
--- a/python/migrate.py
+++ b/python/migrate.py
@@ -2,7 +2,7 @@ from decimal import Decimal
 import boto3
 import csv
 import io
-from gqlalchemy import Memgraph, Neo4j
+from gqlalchemy import Neo4j
 import json
 import mgp
 import mysql.connector as mysql_connector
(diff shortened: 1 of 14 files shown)


python/migrate.py
```diff
@@ -433,12 +462,107 @@ def cleanup_migrate_s3():
...
+mgp.add_batch_read_proc(neo4j, init_migrate_neo4j, cleanup_migrate_neo4j)
+
+
+def _formulate_cypher_query(label_or_rel_or_query: str) -> str:
+    words = label_or_rel_or_query.split()
+    if len(words) > 1:
+        return (
+            label_or_rel_or_query  # Treat it as a Cypher query if multiple words exist
+        )
+    node_match = re.match(r"^\(\s*:(\w+)\s*\)$", label_or_rel_or_query)
```
imilinovic, line 544: add comment what does this match
[ :Relationship ]
( :Label )

python/migrate.py
```diff
@@ -433,12 +462,107 @@ def cleanup_migrate_s3():
...
+    rel_match = re.match(r"^\[\s*:(\w+)\s*\]$", label_or_rel_or_query)
+
+    if node_match:
+        label = node_match.group(1)
+        return (
+            f"MATCH (n:{label}) RETURN labels(n) as labels, properties(n) as properties"
+        )
+    elif rel_match:
```
imilinovic, line 552: if

imilinovic (approved):
Two small comments

PR #564, opened 2025-03-14
Add onelogin

mattkjames7:
### Description

Added missing `onelogin` `pip` package to the dockerfiles. Also added missing `--no-cache-dir` arguments to `pip` calls.

### Pull request type

- [ ] Bugfix
- [ ] Algorithm/Module
- [ ] Feature
- [ ] Build related changes

#### Module/Algorithm
- [ ] Tests provided (unit / e2e)
- [ ] Code documentation

### Documentation checklist
- [ ] Add the documentation label tag
- [ ] Add the bug / feature label tag
- [ ] Add the milestone for which this feature is intended
    - If not known, set for a later milestone
- [ ] Write a release note, including added/changed clauses
    - Added missing `onelogin` package and reducing image size.
- [ ] Link the documentation PR here
    - **[Documentation PR link]**
- [ ] Tag someone from docs team in the comments

### DIFF
--- a/python/requirements.txt
+++ b/python/requirements.txt
@@ -7,6 +7,7 @@ gqlalchemy==1.6.0
 igraph==0.11.8
 mysql-connector-python==9.1.0
 networkx==2.8.8
+onelogin==3.1.6
 oracledb==2.5.1
 psycopg2-binary==2.9.10
 pyodbc==5.2.0
--- a/.github/workflows/reusable_package_mage.yaml
+++ b/.github/workflows/reusable_package_mage.yaml
@@ -211,6 +211,7 @@ jobs:
         run: |
           mkdir output
           docker save ${DOCKER_REPOSITORY_NAME}:${{ env.IMAGE_TAG }} > output/${{ env.ARTIFACT_NAME }}.tar.gz
+          du -h output/${{ env.ARTIFACT_NAME }}.tar.gz
 
       - name: Push to GitHub
         if: ${{ inputs.push_to_github == 'true' }}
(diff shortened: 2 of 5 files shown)


python/requirements.txt
```diff
@@ -7,6 +7,7 @@ gqlalchemy==1.6.0
 igraph==0.11.8
 mysql-connector-python==9.1.0
 networkx==2.8.8
+onelogin
```
DavIvek, line 10: Maybe we should specify the version of the package when including it to prevent future updates with breaking changes from causing issues, saving us from having to backtrace later. 
Version 3.1.6 is installed somewhere in [Memgraph](https://github.com/memgraph/memgraph/blob/master/tests/setup.sh) so maybe we can use this one or the latest one.

manual_package.sh
```diff
@@ -130,7 +130,7 @@ docker buildx build \
 
 # Save Docker image
 echo "Saving Docker image..."
-mkdir -p output
+mkdir -p ../output
 docker save "$DOCKER_REPOSITORY_NAME:$IMAGE_TAG" > "../${ARTIFACT_NAME}.tar.gz"
```
DavIvek, line 134: Should this be saved in `output` dir?

python/requirements.txt
mattkjames7: That would be a good idea - I need to figure out which version actually gets installed. I initially set this to be the latest version, but it cause problems with `torch`.

manual_package.sh
mattkjames7: done!

DavIvek (approved):

imilinovic (approved):

PR #618, opened 2025-06-04
Add nested indices test

matea16:
Add test for nested indices

### DIFF
--- a/smoke-release-testing/mgconsole/nested_indices.bash
+++ b/smoke-release-testing/mgconsole/nested_indices.bash
@@ -1,0 +1,22 @@
+#!/bin/bash
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+source "$SCRIPT_DIR/../utils.bash"
+
+# TODO(gitbuda): Introduce default __host and __port values.
+test_nested_indices() {
+  __host="$1"
+  __port="$2"
+  echo "FEATURE: Nested Indices"
+
+  echo "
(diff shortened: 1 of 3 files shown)


smoke-release-testing/mgconsole/nested_indices.bash
```diff
@@ -0,0 +1,22 @@
...
+# TODO(gitbuda): Introduce default __host and __port values.
+test_nested_indices() {
+  __host="$1"
+  __port="$2"
+  echo "FEATURE: Nested Indices"
+
+  echo "
+    CREATE (:Product {specifications: {dimensions: {width: 10, height: 20}}});
+    CREATE INDEX ON :Product(specifications.dimensions.width);
+    RETURN 1;
```
antejavor, line 14: We should check the output of a creation. You want to run a simple query to see that the index is being hit; this test checks the creation, not that the index is working.

smoke-release-testing/mgconsole/nested_indices.bash
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
antejavor, line 1: If you take a look [here](https://github.com/memgraph/mage/blob/main/smoke-release-testing/mgconsole/indices.bash) there is an indices file that can hold different type of indices, Buda started adding it under it's own PRa: https://github.com/memgraph/mage/pull/607/files

antejavor (changes requested):
I would maybe close this PR and move a content to PR from Buda, wdyt? @matea16

PR #662, opened 2025-09-03
Add support for pytorch GPU under Docker

gitbuda:


python/embeddings.py
```diff
@@ -0,0 +1,136 @@
...
+    )
+    try:
+        # Parent imports are okay; workers import only embed_worker
+        import embed_worker  # <-- our pure worker module
+    except Exception as e:
+        logger.error(f"Failed to import worker module: {e}")
+        return mgp.Record(success=False)
+
+    try:
+        vertices = list(ctx.graph.vertices)
```
gitbuda, line 66: Batching is required here because otherwiese all nodes are going to be materialized in the memory of the query module.

python/embeddings.py
```diff
@@ -0,0 +1,136 @@
...
+    except Exception:
+        try:
+            import torch
+            return list(range(torch.cuda.device_count())) if torch.cuda.is_available() else []
+        except Exception:
+            return []
+
+
+@mgp.write_proc
+def compute_embeddings(ctx: mgp.ProcCtx) -> mgp.Record(success=bool):
```
gitbuda, line 54: Add vertices as an argument here, if passed only that subset of vertices should be calculated

python/embeddings.py
```diff
@@ -0,0 +1,136 @@
...
+from concurrent.futures import ProcessPoolExecutor, as_completed
+import mgp
+
+sys.path.append(os.path.join(os.path.dirname(__file__), "embed_worker"))
+
+logger: mgp.Logger = mgp.Logger()
+
+EXCLUDE_PROPERTIES = {"embedding"}
+BATCH_SIZE = 2000
+MODEL_ID = "all-MiniLM-L6-v2"
```
gitbuda, line 13: parametrize from the outside

python/embeddings.py
```diff
@@ -0,0 +1,136 @@
...
+                model = SentenceTransformer(MODEL_ID, device="cpu")
+                embs = model.encode(
+                    texts,
+                    batch_size=min(BATCH_SIZE, n),
+                    convert_to_numpy=True,
+                    normalize_embeddings=True,
+                    show_progress_bar=False,
+                )
+                for v, e in zip(vertices, embs.tolist()):
+                    v.properties["embedding"] = e
```
gitbuda, line 89: Also parametrize; NOTE: hardcoded on multiple places under the script

python/embeddings.py
```diff
@@ -0,0 +1,139 @@
...
+        import embed_worker  # <-- our pure worker module
+    except Exception as e:
+        logger.error(f"Failed to import worker module: {e}")
+        return mgp.Record(success=False)
+
+    try:
+        if input_vertices:
+            vertices = input_vertices
+        else:
+            vertices = list(ctx.graph.vertices)
```
gitbuda, line 69: Batching is also required here because this will pull all the vertex accessor into the QM memory

